PR #1, opened 2023-01-23
Include DB migration strategies

ccamacho:
This commit include the steps for
stopping the controller nodes services before
dumping the MariaDB data.

### DIFF
--- a/mariadb_copy.md
+++ b/mariadb_copy.md
@@ -58,6 +58,48 @@ DB_ROOT_PASSWORD=SomePassword
       mysql -h "$PODIFIED_MARIADB_IP" -uroot "-p$DB_ROOT_PASSWORD" -e 'SHOW databases;'
   ```
 
+## Procedure - stopping control plane services
+
+From each controller node it is necessary to stop the
+control-plane services to avoid inconsistencies in the
+data migrated for the data-plane adoption procedure.
+
+1- Connect to all the controller nodes and stop the control
+plane services.
+
+2- Stop the services.
+
+```bash
+
+# Configure SSH variables to stop the services
+# in each controller node. For example:
+
+CONTROLLER1_SSH="ssh -F $ENV_DIR/director_standalone/vagrant_ssh_config vagrant@standalone"
+CONTROLLER2_SSH=":"
+CONTROLLER3_SSH=":"
+
+# Update the services list to be stoped
+
+ServicesToStop=("tripleo_horizon.service"
+                "tripleo_keystone.service"
+                "tripleo_cinder_api.service"
+                "tripleo_glance_api.service"
+                "tripleo_neutron_api.service"
+                "tripleo_nova_api.service"
+                "tripleo_placement_api.service")
+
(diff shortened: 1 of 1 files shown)


jistr (changes requested):
I wonder whether this text should rather be input for discussion/meeting + selection of one approach, rather than edit of the user-facing docs? If we document 3 approaches, all of them will have to be implemented (with actual commands, ideally copy-pastable, on how to make each approach happen) and supported by us. I think if at all possible we should select one approach and add the commands for that approach into this guide.

Personally i'd lean towards implementing the "stopping the (OpenStack) services" strategy, that's what's been done for FFUs previously and i think there are least surprises to be had there.

On "stopping the databases" strategy i'd be curious how we do the SQL file dump when the DB is stopped.

On "marking DBs as read-only", i think that should be tested on what that means for the service operation. Can we really claim that the services keep running, or will they start failing in weird ways?  I wonder if there are periodic intra-service or inter-service checks/actions which expect the DB to be writable even though there is no user-triggered API request.

mariadb_copy.md
```diff
@@ -58,6 +58,37 @@ DB_ROOT_PASSWORD=SomePassword
...
+                "tripleo_nova_api.service"
+                "tripleo_placement_api.service")
+
+echo "Stopping the OpenStack services"
+
+for service in ${ServicesToStop[*]}; do
+    echo "Stopping the service: $service"
+    systemctl stop $service
+done
+```
```
jistr, line 86: I think it would be a bit nicer if we could assume the snippet is executed from outside the controller, and the snippet would include ssh command. E.g. something like

CONTROLLER1_SSH="ssh -F $ENV_DIR/director_standalone/vagrant_ssh_config vagrant@standalone"
CONTROLLER2_SSH=":"
CONTROLLER3_SSH=":"

And in the for cycle we could have

```
$CONTROLLER1_SSH sudo systemctl stop $service
$CONTROLLER2_SSH sudo systemctl stop $service
$CONTROLLER3_SSH sudo systemctl stop $service
```

This could cover both 1 controller and 3 controller scenarios. The ":" in the SSH commands is a bash no-op. Not a blocker but i think it would make it a bit nicer, allow people to make their own copy-pastable snippets tailored for their environment. WDYT? Maybe a follow-up PR?

jistr (approved):

mariadb_copy.md
ccamacho: cool, done

jistr: :+1: thanks!

jistr (approved):

bogdando:
> On "marking DBs as read-only", i think that should be tested on what that means for the service operation. Can we really claim that the services keep running, or will they start failing in weird ways? I wonder if there are periodic intra-service or inter-service checks/actions which expect the DB to be writable even though there is no user-triggered API request.

another strategy could be freezing tables and making lvm snapshot of the mysql data logical volume (but implies specific nodes provisioning requirements, like thin lvm, unallocated space reserved for snapshots etc

PR #2, opened 2023-01-24
Remove password synchronization between source/destination

jistr:
Doing the adoption as a "big bang" rather than gradual, we no longer need to ensure that the passwords in source match the passwords in destination, so let's not do that. The reasons are:

* Making the procedure shorter by removing unnecessary steps.

* Working better with the current podified control plane PoC created by "install_yamls", which forces use a single password for all service DBs, and is thus incompatible with the source environment in that regard.

* Allowing some services to perhaps gain stronger passwords during the migration procedure.

### DIFF
--- a/mariadb_copy.md
+++ b/mariadb_copy.md
@@ -29,10 +29,13 @@ Define the shell variables used in the steps below. The values are
 just illustrative, use values which are correct for your environment:
 
 ```
-EXTERNAL_MARIADB_IP=192.168.24.3
 PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
 MARIADB_IMAGE=quay.io/tripleozedcentos9/openstack-mariadb:current-tripleo
-DB_ROOT_PASSWORD=SomePassword
+
+# Use your environment's values for these:
+EXTERNAL_MARIADB_IP=192.168.24.3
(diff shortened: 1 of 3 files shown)


openstack_control_plane_deployment.md
```diff
@@ -50,18 +39,13 @@ KEYSTONE_DATABASE_PASSWORD=SomePassword
...
-  > addressed further.
-
   ```
   cat > clouds-adopted.yaml <<EOF
   clouds:
     adopted:
       auth:
         auth_url: http://keystone-public-openstack.apps-crc.testing
-        password: $ADMIN_PASSWORD
+        password: 12345678
```
jistr, line 48: Actually maybe rather than hard-coding the install_yamls PoC password, i should probably reintroduce this as a bash variable, but no longer assume that the admin passwords are the same between podified and original.

mariadb_copy.md
```diff
@@ -30,9 +30,10 @@ just illustrative, use values which are correct for your environment:
 
 ```
 EXTERNAL_MARIADB_IP=192.168.24.3
+EXTERNAL_DB_ROOT_PASSWORD=SomePassword
```
ccamacho, line 33: Maybe something like

`EXTERNAL_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep 'MysqlRootPassword' | tr -d ' ' | cut -d ':' -f 2)`

Is a bit more illustrative than SomePassword or 12345678 at least that will show a place where the passwords might be located.

jistr: Thanks that's a good point, updating it.

PR #5, opened 2023-02-01
Add Glance Adoption POC

fmount:
This patch adds the `Glance Adoption POC` document, and fixes the Ceph statement to move the procedure at the end of the entire adoption process for RBD.

Signed-off-by: Francesco Pantano <fpantano@redhat.com>

### DIFF
--- a/glance_adoption.md
+++ b/glance_adoption.md
@@ -1,0 +1,240 @@
+# Glance Adoption
+
+Adopting Glance means that an existing `OpenStackControlPlane` CR, where Glance
+is supposed to be disabled, should be patched to start the service with the
+configuration parameters provided by the source environment.
+
+When the procedure is over, the expectation is to see the `GlanceAPI` service
+up and running: the `Keystone endpoints` should be updated and the same backend
+of the source Cloud will be available. If the conditions above are met, the
+adoption is considered concluded.
+
+This guide also assumes that:
+
+1. A `TripleO` environment (the source Cloud) is running on one side;
+2. A `SNO` / `CodeReadyContainers` is running on the other side;
+3. (optional) an internal/external `Ceph` cluster is reacheable by both `crc` and
+`TripleO`
+
+In order to keep things simple, this scenario assumes that both `MariaDB` and
+`Keystone` are already adopted; the procedure is described by the following
+documentation:
+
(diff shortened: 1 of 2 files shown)


jistr:
Thanks, i was wondering whether to put all services into the "OpenStack Control Plane deployment" doc, but there may be enough configuration detail to address for each service that a doc per service would probably be better.

I'll merge and perhaps i'll also follow up with a PR to restructure things a bit to make the docs more consistent re post-check steps etc.

jistr (approved):

glance_adoption.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Test the glance service from the OpenStack cli
+
+Inspect the resulting glance pods:
+
+```
+sh-5.1# cat /etc/glance/glance.conf.d/01-custom.conf
+
+[DEFAULT]
+enabled_backends=default_backend:rbd,ceph1:rbd
```
konan-abhi, line 150: This should be defined only once

glance_adoption.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Image upload
+
+We can test that an image can be created on from the adopted service.
+
+```
+(openstack)$ export OS_CLIENT_CONFIG_FILE=clouds-adopted.yaml
+(openstack)$ export OS_CLOUD=adopted
+(openstack)$ curl -L -o /tmp/cirros-0.5.2-x86_64-disk.img http://download.cirros-cloud.net/0.5.2/cirros-0.5.2-x86_64-disk.img
+    qemu-img convert -O raw /tmp/cirros-0.5.2-x86_64-disk.img /tmp/cirros-0.5.2-x86_64-disk.img.raw
```
konan-abhi, line 184: If you use interoperable image import's image_conversion plugin then this step is not required, we can enhance it later to include how to enable image conversion plugin.

konan-abhi (commented):
Thank you for the doc, added minor comments inside.

glance_adoption.md
fmount: +1, bad paste, I'll fix it in a follow up!

glance_adoption.md
fmount: yeah, that's nice and good to follow up!

PR #6, opened 2023-02-01
Add notes about MariaDB PVC

fao89:


mariadb_copy.md
```diff
@@ -10,6 +10,8 @@ cluster.
 
   * The OpenStackControlPlane resource must be already created at this point.
 
+  * The MariaDB PVC is created
```
jistr, line 13: This is a bit superfluous as below we have "MariaDB is running" prerequisite. If it can't create PVC, it won't be running (the issue you've hit).

Rather than focusing on PVC of MariaDB, let's focus on having free PVs. So i think we don't have to mention anything additional in "MariaDB copy" or "OpenStack Control Plane Deployment" documents, but we should add a prerequisite into the "Backend services deployment" that we need free PVs so that MariaDB and Rabbit can create their PVCs. AFAIK the `make crc_storage` step creates PVs, not PVCs.

jistr (changes requested):

mariadb_copy.md
fao89: yeah, I reprovisioned CRC, and I think I understand now:
the PVC is already there, but it is pending
```
[dpadev@osp-dev-05 install_yamls]$ oc get pvc mariadb-openstack -o jsonpath='{.status.phase}{"\n"}'                                                                                           
Pending         
```
when we run `make crc_storage`  it creates the PVs, and then the `openstack-db-init ` pod bounds the PVC:
```
[dpadev@osp-dev-05 install_yamls]$ oc get pvc mariadb-openstack -o jsonpath='{.status.phase}{"\n"}'                                                                                           
Bound         
```

backend_services_deployment.md
```diff
@@ -12,6 +12,8 @@ podified OpenStack control plane services.
 * The `openstack-operator` deployed, but `OpenStackControlPlane`
   **not** deployed.
 
+* The MariaDB PVC is bound.
```
jistr, line 15: I don't think that's correct. We need an available PV to start with, but the PVC shouldn't be bound yet. The PVC will be created by the MariaDB deployment, not by `make crc_storage`. I would formulate the prerequisite as:

* Available free PVs that can be claimed by MariaDB and RabbitMQ. (In developer environments driven by install_yamls, run `make crc_storage`.)

backend_services_deployment.md
```diff
@@ -75,3 +77,16 @@ podified OpenStack control plane services.
...
+  ```
+  oc get pvc mariadb-openstack -o jsonpath='{.status.phase}{"\n"}'
+  ```
+
+  if not, run:
+
+  ```
+  # in install_yamls
+  make crc_storage
+  ```
```
jistr, line 92: The `make crc_storage` bit is just for dev envs, but the post-checks would be better to be formulated for customer envs too. If we want to add a post-check as part of this PR, i'd add one that checks that the `mariadb-openstack` pod is running. That will cover both cases both for the PVC and other failure causes. And we don't need to add specific remedy info as root causes could vary.

jistr:
I think this should be amended, but i will be doing a docs refactoring right now, so rather than iterate here on this PR and cause merge conflicts, i'll merge it as-is and refactor it as part of the PR that i'm about to send.

PR #7, opened 2023-02-02
Restructuring

jistr:
The main thing in this PR is splitting out Keystone adoption in a separate document. As is apparent from the Glance adoption doc, there may be enough configuration info for each service that we should get back to the document-per-service model. Even if eventually we may be able to patch the OpenStackControlPlane CR for all services at once, the documentation structure will be much clearer if we address each service in its own doc, rather than having one mega-document.

Additionally this PR tweaks MariaDB prerequisite info, puts setting of all service passwords into a common place right after we create the `osp-secret` at the very beginning of the procedure, and tweaks the Glance doc mainly in headings to be more consistent with the rest of the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,17 +4,6 @@ Data Plane Adoption procedure
 Work-in-progress documentation.
 
 
-## Ceph adoption
-
-If the environment includes Ceph and some of its services are
-collocated on the Controller hosts ("internal Ceph"), then Ceph
-services need to be moved out of Controller hosts as the last
-step of the OpenStack adoption.
-Follow this documentation:
-
-* [Ceph cluster migration (RBD)](ceph.md)
-
-
 ## OpenStack adoption
 
 This is a procedure for adopting an OpenStack cloud.
@@ -25,10 +14,22 @@ Perform the actions from the sub-documents in the following order:
 
 * [Copy MariaDB data](mariadb_copy.md)
 
-* [Deploy OpenStack control plane services](openstack_control_plane_deployment.md)
+* [Keystone adoption](keystone_adoption.md)
 
-* [Glance Adoption](glance_adoption.md)
+* [Glance adoption](glance_adoption.md)
+
+* [Adoption of other services](other_services_adoption.md)
 
 If you face issues during adoption, check the
 [Troubleshooting](troubleshooting.md) document for common problems and
 solutions.
+
+
+## Post-OpenStack Ceph adoption
+
+If the environment includes Ceph and some of its services are
+collocated on the Controller hosts ("internal Ceph"), then Ceph
+services need to be moved out of Controller hosts as the last step of
+the OpenStack adoption. Follow this documentation:
+
+* [Ceph cluster migration (RBD)](ceph.md)
--- a/backend_services_deployment.md
+++ b/backend_services_deployment.md
@@ -12,8 +12,6 @@ podified OpenStack control plane services.
 * The `openstack-operator` deployed, but `OpenStackControlPlane`
   **not** deployed.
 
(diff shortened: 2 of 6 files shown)


jistr:
This is probably easier to review by looking at the branch content rather than the git diff: https://github.com/jistr/data-plane-adoption/tree/f/restructure

README.md
```diff
@@ -25,10 +14,22 @@ Perform the actions from the sub-documents in the following order:
 
 * [Copy MariaDB data](mariadb_copy.md)
```
fmount, line 15: Before `MariaDB` copy, we can have a prereq section where we assume TripleO is up && running on one side, and the basic ctlplane is running on the other side: the first thing to do is run crc with all the `disabled` services.
iirc this last topic is covered by [Deploy OpenStack control plane services](https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/backend_services_deployment.md), and it should go first, right?

fao89 (approved):

fao89:
Do we want to publish this doc?
if so, I can set mkdocs like this:
https://openstack-k8s-operators.github.io/openstack-ansibleee-operator/

fmount:
> Do we want to publish this doc? if so, I can set mkdocs like this: https://openstack-k8s-operators.github.io/openstack-ansibleee-operator/

It's not a bad idea, but I'll let @jistr comment on this! Long term, I like the idea of mkdocs!

README.md
jistr: That's pretty much all in "Deploy podified backend services" which is already the step right before MariaDB copy. I could expand the prereq list to mention that we need to have the cloud we want to Adopt up and running.

jistr:
> Do we want to publish this doc? if so, I can set mkdocs like this: https://openstack-k8s-operators.github.io/openstack-ansibleee-operator/

I'd like to explore & chat about what it would mean re structuring of the content, ease of contribution, and how to generate the site locally, before we jump onto that. If we use a static site generator, there should be an easy way to generate the site locally, perhaps utilizing a Makefile + virtualenv. I'd like to explore if we get enough benefits over the built-in Github markdown renderer without causing much hassle for contributors/maintainers. Also i'm curious what other projects in the org (beyond Ansible EE Operator) plan to do regarding their docs. It would be good to get consistency. I like the sidebar menu there. So i'm leaning towards +1 but let's discuss.

fmount:
> > Do we want to publish this doc? if so, I can set mkdocs like this: https://openstack-k8s-operators.github.io/openstack-ansibleee-operator/
> 
> I'd like to explore & chat about what it would mean re structuring of the content, ease of contribution, and how to generate the site locally, before we jump onto that. If we use a static site generator, there should be an easy way to generate the site locally, perhaps utilizing a Makefile + virtualenv. I'd like to explore if we get enough benefits over the built-in Github markdown renderer without causing much hassle for contributors/maintainers. Also i'm curious what other projects in the org (beyond Ansible EE Operator) plan to do regarding their docs. It would be good to get consistency. I like the sidebar menu there. So i'm leaning towards +1 but let's discuss.

Yeah let's discuss and follow up, but definitely the restructuring patch looks good!

README.md
fmount: Thanks for clarifying! LGTM.

fmount (approved):

README.md
jistr: I'll add one more commit mentioning that the cloud which we want to adopt should be up and running, and it has to be on Wallaby. And i'll merge.

jistr:
Thanks for the reviews! :)

PR #8, opened 2023-02-03
MkDocs

fao89:
PoC of using MkDocs to publish the docs
https://fao89.github.io/data-plane-adoption/


fao89:
@jistr @fmount 
I started this MkDocs PoC, feel free to take or close it, as I'll be out next week
- https://github.com/fao89/data-plane-adoption/tree/mkdocs
- https://fao89.github.io/data-plane-adoption/

jistr:
Chatted with Fabricio, we should amend the left sidebar navigation to be multi-level, something like:

- OpenStack
  - Backend
  - MariaDB
  - Keystone
  - Glance
  - etc.
- Ceph
  - RBD
  
  Other than that the rendering looks good to me. I'll test the local site generation too and then i'm +1.

mkdocs.yml
```diff
@@ -0,0 +1,51 @@
...
+  - README.md
+  - OpenStack:
+    - backend_services_deployment.md
+    - mariadb_copy.md
+    - keystone_adoption.md
+    - glance_adoption.md
+    - other_services_adoption.md
+  - Ceph:
+    - ceph.md
+  - troubleshooting.md
```
fao89, line 37: @jistr FYI this is how we define the sidebar

jistr:
Looks good, and local docs building WFM. :+1: Let's discuss with others on the call today.

mkdocs.yml
fmount: This is actually nice and really useful, I'm going to have the RGW adoption section under `Ceph`, so it's nice I can define I list of documents to provide a better navigation!

fultonj:
IMO we should merge this. If more people want to write docs in a different format they can do so, but we can't wait for the perfect time. Some docs are better than no docs.

fmount:
> IMO we should merge this. If more people want to write docs in a different format they can do so, but we can't wait for the perfect time. Some docs are better than no docs.

Agree w/ John here, and given great part of the value of this repo is the doc/procedures that contains, +1 on merging it!

fao89:
I'm gonna merge it then

jistr:
:+1: Thanks Fabricio.

PR #10, opened 2023-02-14
Bootstrap the Adoption test suite

jistr:
So far the test suite only covers the steps upto "backend services deployment".

The interface between the execution infrastructure and the test suite
is an Ansible inventory file. Inventory samples are provided. To
execute the tests, follow this procedure.

* Create `tests/inventory.yaml` file by copying and editing one of the
  included samples (e.g. `tests/inventory.sample-crc-vagrant.yaml`) to
  provide values valid in your environment.

* Run `make test-minimal`.

### DIFF
--- a/tests/inventory.sample-crc-vagrant.yaml
+++ b/tests/inventory.sample-crc-vagrant.yaml
@@ -1,0 +1,11 @@
+adopter:
+  hosts:
+    localhost:
+      ansible_connection: local
+
+controllers:
+  hosts:
+    standalone:
+      ansible_ssh_host: standalone_hostname_or_ip
+      ansible_ssh_user: vagrant
+      ansible_ssh_private_key_file: ~/director_standalone/vagrant_private_key
--- a/tests/inventory.sample-crc-vagrant.yaml
+++ b/tests/inventory.sample-crc-vagrant.yaml
@@ -1,21 +1,4 @@
 all:
-  vars:
-      install_yamls_path: /CUSTOMIZE_THIS
-      reset_crc_storage: true
-
-      oc_header: |
-        source CUSTOMIZE_THIS.sh
-      oc_login_command: |
-        oc login -u kubeadmin -p crcadmin
-
-      cinder_password: CUSTOMIZE_THIS
-      glance_password: CUSTOMIZE_THIS
-      ironic_password: CUSTOMIZE_THIS
-      neutron_password: CUSTOMIZE_THIS
-      nova_password: CUSTOMIZE_THIS
-      octavia_password: CUSTOMIZE_THIS
-      placement_password: CUSTOMIZE_THIS
-
   hosts:
     localhost:
       ansible_connection: local
--- a/tests/inventory.sample-crc-vagrant.yaml
+++ b/tests/inventory.sample-crc-vagrant.yaml
@@ -1,11 +1,33 @@
-adopter:
+all:
+  vars:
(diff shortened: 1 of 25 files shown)


tests/inventory.sample-crc-vagrant.yaml
```diff
@@ -0,0 +1,33 @@
...
+      oc_login_command: |
+        oc login -u kubeadmin -p crcadmin
+
+      cinder_password: CUSTOMIZE_THIS
+      glance_password: CUSTOMIZE_THIS
+      ironic_password: CUSTOMIZE_THIS
+      neutron_password: CUSTOMIZE_THIS
+      nova_password: CUSTOMIZE_THIS
+      octavia_password: CUSTOMIZE_THIS
+      placement_password: CUSTOMIZE_THIS
```
jistr, line 17: This bit could be improved upon, to make the tests themselves fetch the passwords from a Heat environment file for example. The procedure for that could be slightly different per environment though, e.g. it would differ slightly in Standalone envs vs. full Director envs, so we'd have to make the fetching mechanism either generic enough, or having a "switch" logic for env types.

So it depends where do we want the "password fetching" bit going forward, whether as part of the test suite, or test infra setup.

jistr: So far i've only run against CRC+Vagrant env, once we see e.g. how would these passwords be extracted in CI context, it may become better apparent what could be the right way to abstract this in the test suite (assuming we'd want the extraction logic as part of the test suite).

jistr: Based on the call discussion i'm moving the secrets into their own file, and i'll do the same for variables. This loses the advantage of having a single file with all environment info, but may be more suitable for CI, e.g. writing out the vars/secrets files via automation without having to watch out for YAML indentation in the inventory.

tests/secrets.sample.yaml
```diff
@@ -0,0 +1,10 @@
+oc_login_command: |
+  oc login -u kubeadmin -p crcadmin
```
marios, line 2: also CUSTOMIZE_THIS for the crc login

jpodivin (changes requested):
Great work. I have just one request. Could we include some README.md in roles themselves? Right now it's small and obvious, but in the future it's bound to grow and some reference to the purpose and structure could come in handy.

The default role readme from ansible should work well as a template.

jistr:
Rebased and addressed the reviews. Given that this is a testing repo that's still going to see much churn in the coming weeks, i kept the role docs relatively brief. To document variables i just linked to defaults.yaml and provided comments there, so that the readme doesn't get out of sync with the actual variables.

tests/roles/prelude_controller/tasks/main.yaml
```diff
@@ -0,0 +1,2 @@
+- name: check connection
+  ansible.builtin.ping:
```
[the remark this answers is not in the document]
jistr: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs/mariadb_copy.md#procedure---stopping-control-plane-services

See ^^. So creating the prelude_controller role was a bit premature, it may not be needed at all.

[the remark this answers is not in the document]
jistr: (This area could be a candidate for actual end user automation later, but for now i'd like to stick to the docs procedure as closely as possible.)

fao89 (approved):

PR #11, opened 2023-02-21
Add Ceph RGW migration document

fmount:
This patch introduces the RGW migration document which shows how radosGW can be migrated to the CephStorage nodes and HA can be maintained using the Ceph IngressDaemon.


mkdocs.yml
```diff
@@ -33,7 +33,8 @@ nav:
     - glance_adoption.md
     - other_services_adoption.md
   - Ceph:
-    - ceph.md
+    - ceph_rbd.md
```
fmount, line 36: @fao89 @jistr I run this locally (`mkdocs serve`) and I see the two sections properly rendered with their title.

fao89: FYI: you can also rename it here:
```suggestion
    - Ceph new name: ceph_rbd.md
```

fao89 (approved):

jistr (approved):

PR #12, opened 2023-02-22
Move contribution docs to rendered site, add docs writing tips

jistr:

PR #13, opened 2023-02-22
Fix both RBD and RGW links

fmount:
This patch fixes the broken links at the bottom of the `README`. 
Ceph has now multiple sections and procedures according to the daemons that should be migrated, and the doc is updated accordingly.


jistr:
I have a PR which does a bit of reshuffling and also fixes the links -- let's do that one instead? https://github.com/openstack-k8s-operators/data-plane-adoption/pull/12

fmount:
> I have a PR which does a bit of reshuffling and also fixes the links -- let's do that one instead? #12

oh nice, definitely better! Thanks!

PR #19, opened 2023-03-01
Add logging to Makefile test-minimal target and configurable config

marios:
This is needed for the wireup of the test-minimal test target at [1]

[1] https://review.rdoproject.org/r/c/rdo-jobs/+/47392

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 TEST_INVENTORY ?= tests/inventory.yaml
 TEST_VARS ?= tests/vars.yaml
 TEST_SECRETS ?= tests/secrets.yaml
+TEST_OUTFILE := tests/logs/test_minimal_out_$(shell date +%FT%T%Z).log
+TEST_CONFIG ?= tests/ansible.cfg
 
 test-minimal:
-	ANSIBLE_CONFIG=tests/ansible.cfg ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_minimal.yaml
+	mkdir -p tests/logs
+	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 /tests/secrets.yaml
 /tests/playbooks/tmp
 /tests/vars.yaml
+/tests/logs
 
 # rendered docs
 /site


marios:
posted https://review.rdoproject.org/r/c/config/+/47323 to fix the logic @ https://github.com/rdo-infra/review.rdoproject.org-config/blob/8e5c2d01b955dab9cc7920985896984a90606ede/playbooks/crc/openstack.yaml#L90-L95 

we can recheck in a bit

marios:
recheck

marios:
check-rdo

Makefile
```diff
@@ -1,6 +1,7 @@
-TEST_INVENTORY ?= tests/inventory.yaml
-TEST_VARS ?= tests/vars.yaml
-TEST_SECRETS ?= tests/secrets.yaml
+TEST_INVENTORY ?= tests/inventory.sample-crc-vagrant.yaml
+TEST_VARS ?= tests/vars.sample.yaml
+TEST_SECRETS ?= tests/secrets.sample.yaml
```
jistr, line 3: I think rather than point to the sample files, we should expect the developer (or CI) to create those files? If there are things that should be reasonably defaulted, then maybe we can move them from the sample files to the defaults.

jistr: +1 on the addition of TEST_OUTFILE

tests/secrets.sample.yaml
```diff
@@ -1,10 +1,10 @@
...
-  oc login -u kubeadmin -p CUSTOMIZE_THIS
-
 # Adopted OpenStack admin password. Matching the install_yamls default
 # to reduce developer confusion.
-admin_password: 12345678
+admin_password: 123456789
+
+# Command to login to OpenShift
+oc_login_command: |
+  oc login -u kubeadmin -p {{ admin_password }}
```
jistr, line 7: The admin_password is for OpenStack while this password here is for CRC, i don't think they should be assumed to be the same. In general i didn't think CI would be using this file directly. Rather write a `test/secrets.yaml` file from scratch?

tests/vars.sample.yaml
```diff
@@ -1,11 +1,11 @@
 # Path to install_yamls repo
-install_yamls_path: /CUSTOMIZE_THIS
+install_yamls_path:  /home/zuul-worker/src/github.com/openstack-k8s-operators/install_yamls/
```
jistr, line 2: Again this is an environment-specific value that won't work outside CI. I'd rather keep the sample shouting "hey CUSTOMIZE_THIS" rather than provide CI-specific values. CI should probably create `vars.yaml` rather than try to use `vars.sample.yaml` directly?

jistr: Cool. That CI-specific vars.yaml can perhaps even live on the CI side rather than here in the repo? These files - inventory, vars.yaml and secrets.yaml - should be the interface between the test suite and the execution environment. So rather than having vars for all available execution envs in this repo, each execution env should carry its own definition of those 3 files i think.

marios: then are you OK to carry a ci specific make target here? problem is the existing target hard codes these vars (and whats more they are -e so we cant override them they have highest precedence)

jistr: I'm thinking can the CI execution just create `tests/vars.yaml` and `tests/secrets.yaml`? There maybe something i'm missing though.

Makefile
```diff
@@ -1,6 +1,8 @@
 TEST_INVENTORY ?= tests/inventory.yaml
 TEST_VARS ?= tests/vars.yaml
 TEST_SECRETS ?= tests/secrets.yaml
+TEST_OUTFILE := tests/test_minimal_out_$(shell date +%FT%T%Z).log
```
jistr, line 4: I just realized this should go into gitignore, and i was also wondering whether it would be better in a logs dir so that logs (disposable files) can be more clearly separated from vars and inventory (long-living files, at least in dev envs). But neither of that is a blocker and i'm ok just merging as-is. :+1:

jistr (approved):

Makefile
marios: o/

No problem, we can create a logs directory. I can push an update here to do that and will update the rdo side to pickup the new file location. I will add the logs dir to the gitignore.

At least this seems to be working OK now... together with the wireup on the rdo side @ https://review.rdoproject.org/r/c/rdo-jobs/+/47392 (where I've now moved ci specific vars/secrets/ansible.cfg) we have the running test with the logs at [1].

[1] https://logserver.rdoproject.org/23/46923/9/check/data-plane-adoption-github-rdo-centos-8-crc-centos-9-standalone/9dc782b/logs/subnode-1/home/zuul-worker/test_minimal_out_2023-03-16T12:25:37UTC.log

jistr (approved):
:+1: Looks good, thanks! We can merge whenever you'd like.

marios:
@jistr i think this is good to go - latest run result [1] and you can see the test logs at [2] 

 I will need to add some more of the vars as it fails on mariadb (didn't set the secret or the ip for mariadb eg in https://review.rdoproject.org/r/c/rdo-jobs/+/47392/26/playbooks/data_plane_adoption/vars.yaml )

Lets merge this side (and probably the https://review.rdoproject.org/r/c/rdo-jobs/+/47392 too ) and then i can tweak the vars with followup so we at least start to see these tests run on pull requests wdyt.

[1] https://review.rdoproject.org/zuul/build/9ed681ee9d0441d6a6c70076fdbe1526
[2] https://logserver.rdoproject.org/23/46923/9/check/data-plane-adoption-github-rdo-centos-8-crc-centos-9-standalone/9ed681e/logs/subnode-1/home/zuul-worker/test_minimal_out_2023-03-16T18:07:39UTC.log

jistr:
Sounds great, thanks!

PR #21, opened 2023-03-07
Add symlink for top-level roles directory needed by zuul

marios:
This adds a symlink for the tests/roles directory so that zuul can discover and consume the test roles.

Needed by [1] where we are trying to wireup the test_minimal

[1] https://review.rdoproject.org/r/c/rdo-jobs/+/47392

### DIFF
--- a/roles
+++ b/roles
@@ -1,0 +1,1 @@
+tests/roles


marios:
@jistr i think we should catch up at some point about this so i can give you some context ... worst case during tomorrow's meeting or we can meet tomorrow morning if you are available will ping you on slack

jistr:
Hey thanks, what you linked there looks nice but i'd have one comment -- let's not write another test playbook in the RDO jobs repo. We should be using the playbook from this repo in `tests/playbooks` dir. Ideally we should just call it as documented, using `make test-minimal`. Perhaps it would be best if we schedule a call for me & you & @fao89 to sync up tomorrow or Friday?

jistr:
Here's how we do it in OS Migrate to ensure devs & CI run the same thing:

https://github.com/os-migrate/os-migrate/blob/552bafbbb4ff2c898baf697c8dba5d7d5dfa3401/.github/workflows/consistency-functional.yml#L394

PR #22, opened 2023-03-08
Moves the reset_crc_storage_var to the pcp_cleanup role

marios:
This var is used by pcp_cleanup not prelude_local so lets move to the right defaults file. This was noticed as undefined var in [1]

[1] https://logserver.rdoproject.org/23/46923/7/check/data-plane-adoption-github-rdo-centos-8-crc-centos-9-standalone/7ba1034/job-output.txt


jistr:
:+1: Good call. The environment reset is probably something that people will address differently quite often, so this specific "install yamls CRC-focused solution" can just remain hidden and switched off by default i think.

PR #23, opened 2023-03-08
Tests for Placement + small fixes (depends on #20)

jistr:
depends on #20


ccamacho (approved):

fao89 (commented):
LGTM, is it in draft because it is waiting on https://github.com/openstack-k8s-operators/data-plane-adoption/pull/20 ?

jistr:
Yes it was because of #20. So if you're both ok with it, we can merge.

PR #25, opened 2023-03-17
Adding a tab in the OpenStackControlPlane so the code is visible

pnavarro:
Adding a tab in the OpenStackControlPlane so the code is visible


jistr (approved):
I have a fix for the defaulting in the works that does this + the defaults workaround, but let's merge this in the meantime too.

PR #28, opened 2023-03-21
Ovn adoption fix patch

booxter:
I'm not sure why this patch was modified pre-merging the original PR, but the command is failing because the format of the patch doesn't reflect the type definition. This PR fixes the patch command.


booxter:
Nevermind, there's something wrong in my env...

PR #32, opened 2023-04-06
Update keystone objects cleanup

jistr:
This is the workflow that was used in recording the demo.

### DIFF
--- a/docs/openstack/keystone_adoption.md
+++ b/docs/openstack/keystone_adoption.md
@@ -54,19 +54,14 @@
   export OS_CLIENT_CONFIG_FILE=clouds-adopted.yaml
   export OS_CLOUD=adopted
 
+  openstack endpoint list | grep keystone | awk '/admin/{ print $2; }' | xargs openstack endpoint delete || true
+
   openstack service list | grep ' cinderv3 ' | awk '{ print $2; }' | xargs openstack service delete || true
   openstack service list | grep ' glance ' | awk '{ print $2; }' | xargs openstack service delete || true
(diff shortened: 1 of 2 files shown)


docs/openstack/keystone_adoption.md
```diff
@@ -54,19 +54,14 @@
   export OS_CLIENT_CONFIG_FILE=clouds-adopted.yaml
   export OS_CLOUD=adopted
 
+  openstack endpoint list | grep ' keystone ' | grep ' admin ' | awk '{ print $2; }' | xargs openstack endpoint delete || true
```
fmount, line 57: I'm ok with this, but at first glance it can be optmized w/ something like:

```
openstack endpoint list | grep keystone | awk '/admin/{ print $2; }' | xargs openstack endpoint delete || true
```

tests/roles/keystone_adoption/tasks/main.yaml
```diff
@@ -63,16 +63,11 @@
     export OS_CLIENT_CONFIG_FILE={{ adopted_clouds_yaml_path }}
     export OS_CLOUD=adopted
 
+    openstack endpoint list | grep ' keystone ' | grep ' admin ' | awk '{ print $2; }' | xargs openstack endpoint delete || true
```
fmount, line 66: same comment here on the `2*grep`

docs/openstack/keystone_adoption.md
jistr: Thanks, updating :)

jistr:
@fmount I also simplified the `service delete` lines in a similar way.

fmount (approved):
looks good!

PR #36, opened 2023-05-03
Allow giving extra args to the test suite

jistr:
... without editing the Makefile, via an environment variable TEST_EXTRA_ARGS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,8 @@ TEST_VARS ?= tests/vars.yaml
 TEST_SECRETS ?= tests/secrets.yaml
 TEST_OUTFILE := tests/logs/test_minimal_out_$(shell date +%FT%T%Z).log
 TEST_CONFIG ?= tests/ansible.cfg
+TEST_EXTRA_ARGS ?=
 
 test-minimal:
 	mkdir -p tests/logs
-	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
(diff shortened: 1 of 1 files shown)


jistr:
Draft - maybe i can make this a bit more useful by adding tags.

Makefile
```diff
@@ -3,7 +3,8 @@ TEST_VARS ?= tests/vars.yaml
 TEST_SECRETS ?= tests/secrets.yaml
 TEST_OUTFILE := tests/logs/test_minimal_out_$(shell date +%FT%T%Z).log
 TEST_CONFIG ?= tests/ansible.cfg
+TEST_EXTRA_ARGS ?=
 
 test-minimal:
 	mkdir -p tests/logs
-	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
+	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) $(TEST_EXTRA_ARGS) tests/playbooks/test_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
```
marios, line 10: I think you'll need another -e for the new args file and use the @ like the others

-e @$(TEST_EXTRA

marios (changes requested):

Makefile
jistr: I'd like the extra args to be able to consume also e.g. `--tags` and not limit it to providing extra variables. So if we want to provide an additional variable file, the `-e @` part should go into the TEXT_EXTRA_ARGS variable itself.

PR #40, opened 2023-05-09
Replace TripleO images by TCIB images

kajinamit:
We imported container build tools from TripleO as TCIB[1]. Now we are retiring TripleO and should complete the migration.

Note that the openstack-tripleoclient container was replaced by the openstack-openstackclient container during migration to TCIB.

[1] https://github.com/openstack-k8s-operators/tcib

### DIFF
--- a/docs/openstack/backend_services_deployment.md
+++ b/docs/openstack/backend_services_deployment.md
@@ -110,7 +110,7 @@ podified OpenStack control plane services.
     mariadb:
       templates:
         openstack:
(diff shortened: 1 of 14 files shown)


fao89 (approved):

jistr (approved):
Thanks, LGTM, we should get the CI green before merging, the failure is likely unrelated so we'll work on the CI outside this PR.

fmount (approved):

jistr:
recheck

PR #44, opened 2023-05-18
Document Development Environment based on install_yamls

fultonj:
This version of the documents describe how to use install_yamls/devsetup to configure edpm-compute-0 with isolated networks which can access the NG control plane on k8s. It then covers installing OpenStack Wallaby using TripleO Standalone on the edpm-compute-0 node with network isolation and Ceph and provides example environment files. It then explains how to add an NG deployment to the environment and leaves the reader linked to the adoption procedures they can follow in the new environment.

### DIFF
--- a/docs/contributing/development_environment.md
+++ b/docs/contributing/development_environment.md
@@ -3,8 +3,8 @@ Development environment
 
 -----
 
-*This is a guide for install_yamls based Adoption environment. It is
-a work in progress. For now please use the previous
+*This is a guide for an install_yamls based Adoption environment with
+network isolation as an alternative to the
 [CRC and Vagrant TripleO Standalone](https://gitlab.cee.redhat.com/rhos-upgrades/data-plane-adoption-dev/-/tree/main/crc-and-vagrant)
 development environment guide.*
 
@@ -13,7 +13,7 @@ development environment guide.*
 The Adoption development environment utilizes
 [install_yamls](https://github.com/openstack-k8s-operators/install_yamls)
 for CRC VM creation and for creation of the VM that hosts the original
-OpenStack (currently in Standalone configuration).
+Wallaby OpenStack in Standalone configuration.
 
 ## Environment prep
 
@@ -49,6 +49,7 @@ GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@latest
 
 ## Deployment of CRC with network isolation
 
+
 ```
 cd install_yamls/devsetup
 PULL_SECRET=$HOME/pull-secret.txt CPUS=12 MEMORY=40000 DISK=100 make crc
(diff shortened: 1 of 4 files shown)


jistr:
Thanks :+1:. I'm going to try and run through this ASAP. @fao89 can you please give this a try too? And maybe @booxter as i think you were waiting with OVN test suite implementation for a net-isolated dev env?

fmount:
Thanks @fultonj, I had a quick look and it seems very good, I'll try the setup in the Glance adoption context (with NFS) where I was stuck due to the missing network isolation!

fao89 (approved):
I haven't tested yet, but it looks good!
(I'm setting my env to test, but I'm ok to merge it)

fao89 (approved):
I've tested it and it worked pretty well

fmount (approved):

docs/contributing/development_environment.md
```diff
@@ -170,66 +188,354 @@ network:
...
+```
+sudo cephadm shell -- rbd -p images ls -l
+```
+
+## Complete the NG Deployment
+
+The remaining steps should be completed on the hypervisor hosting crc
+and edpm-compute-0.
+
+### Deploy NG Control Plane with Ceph
```
fmount, line 392: Not sure about this: I think services should be connected to the `Ceph` cluster as long as they're adopted, so if you follow the glance adoption doc there are steps (including the secrets creation) to run the service and connecting it to the external Ceph.
Shouldn't be the ctlplane almost empty at this stage? /cc @jistr 
Anyway the change itself is good, just wanted to clarify this point before merging.

fmount (changes requested):

docs/contributing/development_environment.md
jistr: Yes i think the dev env setup should produce an environment with "old" OpenStack (+ Ceph) and an empty OpenShift/CRC. Further steps should probably be part of the actual Adoption. (In the "new design" we don't expect to have OpenStack control plane services running at the same time on old and new.)

That said i think the PR looks good and it's great that it worked for Fabricio (i'm still yet to test manually). So i'd say we should merge and iterate from there.

jistr (approved):

docs/contributing/development_environment.md
fultonj: I agree we want the new control plane empty. 

I include these steps since I wanted to show people how easy it is to connect the NG system to the old ceph cluster. How to extract the ceph conf/keys and create the secret. I don't know if that part is as obvious to someone who doesn't use ceph regularly.

I then emphasized that they shouldn't be confused about two services writing to the same ceph and thus they might want to test the connection but that they shouldn't keep the testing data or just use the NG system. Line 436 explains this and then have them delete the test data so that the new control plane remains empty. I then point out that this is a part of the glance adoption and link to your document.

Thanks all for the revews!

PR #46, opened 2023-05-26
Dev env docs: fix network addresses on vlans

jistr:
On CRC side, vlan22 is for tenant network and uses 172.19.0.x. On our side it had 172.10.0.x. This is now being fixed in the sample dataplane CR [1] and we should fix it in Adoption dev envs too.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/209


jistr:
The deployment WFM, marking the PR ready for review.

fao89 (approved):

fultonj (approved):
/lgtm

10 -> 19
19 -> 21

PR #47, opened 2023-05-31
Add snapshot/revert instructions into dev env docs

jistr:


fmount (approved):
+1 on the useful commands

fultonj (approved):
Thanks Jistr. +2

jistr:
This is only affecting docs so i won't re-trigger the CI here.

PR #49, opened 2023-06-02
Clarify what's the critical path of dev env setup

jistr:
vs. what is extra with an additional Compute node.

This is something that we ran into today in our "office hours", we should probably be primarily directing folks towards deploying the simple "standalone + CRC" variant initially.


fultonj (approved):
+2 Thanks @jistr

fao89 (approved):

fmount (approved):

jistr:
Thanks, i'll merge regardless of CI as this is docs only.

PR #51, opened 2023-06-05
OVN procedure improvements

booxter:
- Document database schema version upgrade (using ovsdb-tool convert)
- Split northd deployment out to later stage, after backup is imported, for safety
- Remove troubleshooting step for OVN container version mismatch that is no longer needed


jistr (approved):
LGTM, thanks!

PR #60, opened 2023-06-27
Enable dataplane adotion test

fao89:


fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

fao89:
recheck

cescgina (commented):
looks good to me, will this change require creating a new vm for compute in CI?

fao89:
> looks good to me, will this change require creating a new vm for compute in CI?

it reuses the standalone node

jistr:
Great! One thing we should follow up on in another PR perhaps -- right now the last thing that the test suite does is creation of the OpenStackDataPlane CR [1] and beyond OpenShift not complaining that the CR has a formatting issue, we don't know what happens next -- e.g. whether Ansible can connect to the Standalone node or not. We should add some check that the DataPlane CR gets reconciled.

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/3232403e6baef84ec267593c98745cc01dfd9093/tests/roles/dataplane_adoption/tasks/main.yaml#L57

jistr (approved):

jistr:
We could also remove the compute role from the node configuration if that's possible now, so that EDPM would run essentially an "empty playbook" on the node. In my mind that still goes a long way in terms of CI-ing that EDPM is integrated correctly and we can control the Standalone node with it. And making it an actual working compute node can come later.

Or as you say we can wait for the Nova adoption. We could perhaps decide based on the ETA for the Nova implementation.

fao89:
if we enable it, we will the ansible running on the standalone node, but it will fail when trying to deploy libvirt and nova

fao89:
recheck

fao89:
recheck

fao89:
recheck

cescgina:
@fao89 the error in the last failure looks like what I'm trying to solve with https://github.com/openstack-k8s-operators/data-plane-adoption/pull/68, so we probably should get that in before, although I can't make any promises because we've been seeing a whole lot of crc errors as well

PR #59, opened 2023-06-27
Make rbd-only the default for the dev environment

fao89:
ref: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/58#issuecomment-1608192098

PR #61, opened 2023-06-28
Cinder adoption

Akrog:
Increase the coverage and detail of the cinder adoption.

To explain the cinder adoption it was necessary to add some other sections such as:

- Planning section before the adoption process is started.
- Explanation of the node selection mechanism to decide where OpenStack pods will run within OpenShift nodes.
- Explanation of how service configurations work, including when they have sensitive information.

There are sections that have TODO notes, but at least it serves as a reference/reminder for others that want to improve them.

### DIFF
--- a/docs/contributing/development_environment.md
+++ b/docs/contributing/development_environment.md
@@ -469,6 +469,20 @@ Confirm the image UUID can be seen in Ceph's images pool.
 sudo cephadm shell -- rbd -p images ls -l
 ```
 
+Create a Cinder volume, a backup from it, and snapshot it.
+```
+openstack volume create --image cirros --bootable --size 1 disk
+openstack volume backup create --name backup disk
+openstack volume snapshot create --volume disk snapshot
+```
+
+Boot a VM
+```
+openstack flavor create tiny --id auto --ram 256 --disk 0 --vcpus 1
+openstack --os-compute-api-version 2.37 server create --flavor tiny --image cirros --nic none --wait vm
+openstack server add volume vm disk
+```
+
 ## Performing the Data Plane Adoption
 
 The development environment is now set up, you can go to the [Adoption
--- a/docs/contributing/development_environment.md
+++ b/docs/contributing/development_environment.md
@@ -378,11 +378,12 @@ sudo openstack tripleo deploy \
   --standalone-role Standalone \
   -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
   -e /usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/deployed-network-environment.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/cinder-backup-active-active.yaml \
   -e ~/containers-prepare-parameters.yaml \
   -e standalone_parameters.yaml \
-  -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm-rbd-only.yaml \
   -e ~/deployed_ceph.yaml \
-  -e /usr/share/openstack-tripleo-heat-templates/environments/deployed-network-environment.yaml \
   -e deployed_network.yaml \
   -r /usr/share/openstack-tripleo-heat-templates/roles/Standalone.yaml \
   -n network_data.yaml \
--- a/docs/openstack/keystone_adoption.md
+++ b/docs/openstack/keystone_adoption.md
@@ -1,5 +1,3 @@
-# Keystone adoption
-
 ## Prerequisites
 
 * Previous Adoption steps completed. Notably, the service databases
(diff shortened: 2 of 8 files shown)


docs/contributing/development_environment.md
```diff
@@ -378,11 +378,13 @@ sudo openstack tripleo deploy \
   --standalone-role Standalone \
   -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
   -e /usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/deployed-network-environment.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/deployed-network-environment.yaml \
```
fultonj, line 383: There are two of these, I think you just want one.

fultonj (commented):
Thanks so much Gorka.

docs/contributing/development_environment.md
```diff
@@ -378,11 +378,13 @@ sudo openstack tripleo deploy \
   --standalone-role Standalone \
   -e /usr/share/openstack-tripleo-heat-templates/environments/standalone/standalone-tripleo.yaml \
   -e /usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml \
+  -e /usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml \
```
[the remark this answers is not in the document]
Akrog: Should we mix that change with this PR or submit an independent one?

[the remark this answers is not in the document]
fultonj: Whops, feel free to disregard. I thought https://github.com/openstack-k8s-operators/data-plane-adoption/pull/59 had merged already. If it merges first it can be addressed via rebase. If it merges after it will address it.

docs/openstack/planning.md
```diff
@@ -0,0 +1,197 @@
...
+cinder-backup, and maybe even the glance (if it's using Cinder as a backend)
+services to run on a OpenShift host that has an HBA.
+
+The OpenStack Operators allow a great deal of flexibility on where to run the
+OpenStack services, as we can use node labels to define which OpenShift nodes
+are eligible to run the different OpenStack services.  Refer to the [Node
+Selector guide](node-selector.md) to learn more about using labels to define
+placement of the OpenStack services.
+
+**TODO: Talk about Ceph Storage and Swift Storage nodes, HCI deployments,
```
fultonj, line 148: Ack, I can follow up on this (except Swift).

docs/openstack/planning.md
```diff
@@ -0,0 +1,197 @@
...
+etc.**
+
+## Network
+
+**TODO: Write about isolated networks, NetworkAttachmentDefinition,
```
fultonj, line 153: I'm OK with merging documentation which has a `todo` in it.

fultonj (commented):
I'm +2 to merge this. 

@jistr what do you think

Akrog:
> I'm +2 to merge this.

I just realized the flow for Cinder has a major flaw!!

Because of how the rolling upgrades mechanism works, when the adopted cinder services start on OpenShift they will pin down their RPC and OVO versions to the old release versions.

I'll look into the best way to change the flow.

Akrog:
Instead of modifying the adoption flow I have proposed a patch to the cinder-operator itself to fix this issue: https://github.com/openstack-k8s-operators/cinder-operator/pull/222

docs/openstack/keystone_adoption.md
```diff
@@ -31,51 +29,28 @@
...
 * Clean up old services and endpoints that still point to the old
   control plane (everything except Keystone service and endpoints):
 
-  ```
-  export OS_CLIENT_CONFIG_FILE=clouds-adopted.yaml
-  export OS_CLOUD=adopted
-
-  openstack endpoint list | grep keystone | awk '/admin/{ print $2; }' | xargs openstack endpoint delete || true
+  ```bash
+  openstack endpoint list | grep keystone | awk '/admin/{ print $2; }' | xargs ${BASH_ALIASES[openstack]} endpoint delete || true
```
jistr, line 42: TIL about BASH_ALIASES. A good trick to solve this :)

docs/openstack/node-selector.md
```diff
@@ -0,0 +1,144 @@
+Node Selector
```
jistr, line 1: This doc looks like it applies the same for greenfield deployments as well? Maybe it should eventually live in some "common docs place" and be referenced from Adoption docs.

Akrog: Yes, you are right, a lot of the stuff this doc adds needs to go into a generic doc that applies for both, but I thought I would add it first here and then figure out what we are doing for greenfield.

Akrog:
> We should also think about how to CI Cinder adoption. Currently we only have API/Scheduler deployed by the test suite. Perhaps adding volume and backup would be good once we have a CI environment that includes Ceph, so that we can test the Cinder services with a Ceph backend, which i assume is the dominant one.

That sounds good, having Ceph/RBD for volume and backup would be the ideal.  With time we may also add NVMe-oF or iSCSI if we figure out the LVM to eDPM connection, but those would be secondary.

jistr:
I take it that John's +2 is still applicable after the rebase, and i'm merging :). Thanks!

PR #62, opened 2023-06-28
Glance improvements

Akrog:
This PR adds 3 improvements/fixes to the glance adoption:
- Removes the pre-checks that will always fail.
- Fixes the name of the secret example for ceph and adds an automated mechanism to build the ceph secret.
- Fixes the name of the adopted glance configuration files and adds a way to do the checks automatically.

### DIFF
--- a/docs/openstack/glance_adoption.md
+++ b/docs/openstack/glance_adoption.md
@@ -22,22 +22,6 @@ This guide also assumes that:
 * Previous Adoption steps completed. Notably, MariaDB and Keystone
   should be already adopted.
 
-
-## Pre-check
-
-On the source Cloud, check that the service is active and works as expected,
-and list the existing images:
-
-
-```
-(openstack)$ image list
-+--------------------------------------+--------+--------+
-| ID                                   | Name   | Status |
-+--------------------------------------+--------+--------+
-| c3158cad-d50b-452f-bec1-f250562f5c1f | cirros | active |
-+--------------------------------------+--------+--------+
-```
-
 ## Procedure - Glance adoption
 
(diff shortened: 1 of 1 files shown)


docs/openstack/glance_adoption.md
```diff
@@ -22,22 +22,6 @@ This guide also assumes that:
 * Previous Adoption steps completed. Notably, MariaDB and Keystone
   should be already adopted.
 
-
-## Pre-check
```
fmount, line 26: This assumed there's a running service on the src cloud w/o any issue: it's redundant because if the service is not running on the src cloud there's no point on following this guide, but if you have no images it should be empty but not fail, right?
Perhaps this could be turned into a generic statement on the status of the src service, but if it's not consistent w/ the rest of the doc, feel free to remove

docs/openstack/glance_adoption.md
```diff
@@ -86,7 +70,7 @@ those files in `/etc/ceph`.
 apiVersion: v1
 kind: Secret
 metadata:
-  name: ceph-client-conf
+  name: ceph-conf-files
```
fmount, line 73: +1 on this, it's consistent w/ the rest of the doc

docs/openstack/glance_adoption.md
fmount: I missed the context here: the src service is already stopped at this time.
This is something that changed over the time, now I understand why you removed this section. Thanks, that makes a lot of sense now.

jistr (approved):
Looks good and there's probably nothing fundamental that should get reflected in the test suite right now. CI doesn't have Ceph yet, that should come with the new job that will also have isolated networks.

fultonj (approved):

jistr:
Merging, thanks!

PR #68, opened 2023-07-11
Set replicas for memcached

cescgina:
Set the number of replicas to 1, as otherwise no memcached pod is
created and it prevents keystone coming up.


marios (commented):
lgtm... as jistr mentioned in call please update docs too

fao89 (approved):

marios (approved):
lgtm

PR #74, opened 2023-07-20
docs: introduce a new page for neutron adoption

booxter:
This is largely the same info as captured in other_services page. The latter now goes away because it's not the right place to keep adoption steps documented for any service. Any new service should start with its own separate page.

Note: neutron instructions cover only starting API service for now. Later, the page will be expanded to cover SR-IOV and any other scenarios that may require additional adoption steps.


jistr (approved):

docs/openstack/neutron_adoption.md
```diff
@@ -0,0 +1,73 @@
...
+Patch OpenStackControlPlane to deploy Neutron:
+
+```
+oc patch openstackcontrolplane openstack --type=merge --patch '
+spec:
+  neutron:
+    enabled: true
+    template:
+      databaseInstance: openstack
+      containerImage: quay.io/podified-antelope-centos9/openstack-neutron-server:current-podified
```
jistr, line 35: It would be better to leave out the image and rely on the default?

fao89 (approved):

fao89:
merging as we have 2 approvals

fao89:
recheck

fao89:
recheck

PR #77, opened 2023-07-21
Rename EDPM SSH var

fao89:


cescgina:
recheck  data-plane-adoption-github-rdo-centos-9-crc-single-node  should work now

fao89:
recheck

fao89:
recheck

jistr (approved):

cescgina (approved):
lgtm

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -9,7 +9,7 @@
     metadata:
         name: dataplane-adoption-secret
     data:
-        ssh-privatekey: "{{ edpm_privatekey|b64encode }}"
+        ssh-privatekey: "{{ edpm_encoded_privatekey|default(edpm_privatekey) }}"
```
marios, line 12: ok cool thanks for adding this safely, i.e. with the |default(edpm_privatekey)

marios (approved):
lgtm

this should be good to merge

marios (changes requested):
i think we need to remove the depends-on though to be the other way around

If we merge https://review.rdoproject.org/r/c/rdo-jobs/+/49298 first, it will break the job

we should merge this first since it handles both cases with |default

fao89:
recheck

fao89:
recheck

fao89:
@marios changing the depends on broke CI in both PRs

fao89:
> @marios changing the depends on broke CI in both PRs

actually, the var need to be encoded, I've changed it now

fao89:
@marios could you please take another look?

tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -4,3 +4,4 @@ image_tag: "current-podified"
 ansibleSSHPrivateKeySecret: dataplane-adoption-secret
 deployStrategy:
   deploy: false
+edpm_encoded_privatekey: "{{ edpm_privatekey|b64encode }}"
```
marios, line 7: so edpm_privatekey will no longer be a thing, after we merge the rdo-jobs side https://review.rdoproject.org/r/c/rdo-jobs/+/49298

will you followup? but what will the default be?
 this is a _required_ var right so perhaps we should not even set a default?

marios (commented):
OK but inline question about having the default be edpm_privatekey
it should be OK like this but it would need a followup.

we can discuss in today's ci call? or here if you prefer before then ;)

tests/roles/dataplane_adoption/defaults/main.yaml
fao89: I needed this because I have to encode `edpm_privatekey` , I plan to remove this line after  https://review.rdoproject.org/r/c/rdo-jobs/+/49298 gets merged

marios: ok thanks sounds good

marios (approved):
lgtm

marios:
going to merge this ... only me and @fao89 around today no other cores.

This seems safe to go and we can merge rdo-jobs side https://review.rdoproject.org/r/c/rdo-jobs/+/49298 after this lands

PR #80, opened 2023-07-27
Cinder: Add helper script and instructions

Akrog:
This patch adds a simple script to help us convert a cinder.conf file into some drafts of the configurations required for podified adoption.

The script code itself is pure and utter garbage, and if I could I would remove my name from the git history, but I'm hoping this can help people out when doing their first adoptions.


fao89:
recheck

fao89:
recheck

cescgina:
recheck

fao89 (approved):

cescgina (approved):

jistr (approved):
LGTM. I wonder if going forward there's an opportunity for some integration with https://github.com/openstack-k8s-operators/os-diff/ , even if only on documentation level. Perhaps OS Diff could be the tool that fetches the config and then cinder-cfg.py is used to analyze it.

jistr:
FYI @matbu ^^

PR #97, opened 2023-08-09
Add stop OSP services in sliding menu

averdagu:
The data plane adoption procedure contains a doc on how to stop the OSP services right after the "Deploy podified backend services" doc. This is not reflect in the sliding menu.


jistr (approved):
Thanks Arnau. I think with a second +2 we can override CI and merge, the patch doesn't affect tests.

fao89 (approved):

fao89:
this change wasn't supposed to trigger the CI

PR #102, opened 2023-08-10
[wip] extend mariadb_copy to split out nova cell1 into a seperate db

SeanMooney:
This change modifes the default contolplane CR to deploy two mariadb
instances, the common openstack db and a seperate openstack-cell1
db. The later will be used to adopt the cell1 db called "nova" into
a seperate db instance.

This change extends the extends the mariadb_copy doc and test
to split up the resotre of the db into nova and non_nova dbs.
the non nova dbs are all adopted to the shared openstack db instance.
the nova_api and nova_cell0 db are adopted to the openstack db
and the nova db is adopted into the openstack-cell1 db.

TODO: while this may work as is we currntly have a limitation that
Db name and db user have to be the same. we shoudl be renaming the
nova db to nova-cell1 on import but that is not implemented in this
commit.


jistr (approved):
LGTM

fao89:
recheck

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -82,6 +82,17 @@
...
 
+- name: wait for mariadb to start up
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc get pod mariadb-openstack-cell1 -o jsonpath='{.status.phase}{"\n"}' | grep Running
+  register: mariadb_running_result
+  until: mariadb_running_result is success
+  retries: 60
+  delay: 2
```
SeanMooney, line 93: so this is timing out in ci which implies that my template changes are not working or i need to change it somewhere else.

tests/roles/backend_services/templates/container_overrides.j2
```diff
@@ -45,6 +45,8 @@ spec:
     templates:
       openstack:
         containerImage: {{ container_registry }}/{{ container_namespace }}/openstack-mariadb:{{ container_tag }}
+      openstack-cell1:
```
SeanMooney, line 48: this shoudl be creating the mariadb-openstack-cell1 pod

SeanMooney:
by the way i need to recreate my dev env to work on this again so i wont get back to this until Monday but ill work on this again next week

docs/openstack/mariadb_copy.md
```diff
@@ -111,4 +128,7 @@ SOURCE_MARIADB_IP=127.17.0.100
   ```
   oc run mariadb-client --image $MARIADB_IMAGE -i --rm --restart=Never -- \
      mysql -h "$PODIFIED_MARIADB_IP" -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;'
+  oc run mariadb-client --image $MARIADB_IMAGE -i --rm --restart=Never -- \
+      mysql -h "$PODIFIED_CELL1_MARIADB_IP" -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;' \
+      | grep nova
   ```
```
bogdando, line 134: do we also want to check for cell db mappings, connection/transport URIs etc. records? To compare it later on, if there were no templating issues introduced after adoption is done?

SeanMooney: i think that is out of scope

the intent of this PR is just to copy the data into the renamed db

updating the cell mapping is a separate step that will be likely implemented in the nova operator not in the adoption work.

i.e. when we deploy the conductors with the exting data i would expect the nova operator to correct the cell mappings with the new transport URL and db info so i don't think that should be covered in this pr.

tests/roles/backend_services/templates/container_overrides.j2
SeanMooney: we do not used this in the ci job. i have upstaed the template we use in the latest revision.

docs/openstack/mariadb_copy.md
```diff
@@ -35,14 +35,22 @@ Define the shell variables used in the steps below. The values are
 just illustrative, use values that are correct for your environment:
 
 ```
-PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
 MARIADB_IMAGE=quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
 
-SOURCE_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
+PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
+PODIFIED_CELL1_MARIADB_IP=$(oc get -o yaml pod ${CELL1_POD_NAME} | grep podIP: | awk '{ print $2; }')
```
jistr, line 41: We should mention CELL1_POD_NAME prior to using it here. Or can we just use `mariadb-openstack-cell1`?

docs/openstack/mariadb_copy.md
```diff
@@ -83,31 +93,60 @@ SOURCE_MARIADB_IP=127.17.0.100
...
+      fi
+      db_password=${db_server_password_map["default"]}
+      if [[ -v "db_server_password_map[${db_name}]" ]]; then
+          db_password=${db_server_password_map[${db_name}]}
+      fi
+      echo "creating ${db_name} in ${db_server}"
+      container_name=$(echo "mariadb-client-${db_name}-create" | sed 's/_/-/g')
+      oc run ${container_name} --image ${MARIADB_IMAGE} -i --rm --restart=Never -- \
+          mysql -h "${db_server}" -uroot "-p${db_password}" << EOF
+  CREATE DATABASE IF NOT EXISTS ${db_name} DEFAULT CHARACTER SET ${CHARACTER_SET} DEFAULT COLLATE ${COLLATION};
```
jistr, line 144: If we used the Neutron solution of editing the dump file on "CREATE DATABASE" and "USE" lines, we wouldn't need the explicit DB create and having variables for the char set and collation. Is there a reason not to go for that solution?

jistr (approved):
No blockers i think we can merge this but i do have some questions. If we figure out we'd want something done differently, it can be done in a follow-up patch.

docs/openstack/mariadb_copy.md
SeanMooney: ya thats a mistake i made in clean up

the doc version had CELL1_POD_NAME but i removed that in the test code

ill fix that to just be mariadb-openstack-cell1

docs/openstack/mariadb_copy.md
SeanMooney: that is correct but we considered that previously and considered it to be dangerous

sed -i '/^CREATE DATABASE/s/ovs_neutron/neutron/g;/^USE/s/ovs_neutron/neutron/g' ovs_neutron.sql

in principal this will only match lines starting with 
CREATE DATABASE and update ovs_neutron to neutron
and separately match lines starting with USE
and update ovs_neutron to neutron.

the nova db has user data in it and we do not want to assume 
"USE nova" is not in any data the user provides. i.e. cloud init user-data

in nova case we have to rename form nova to nova_cell1
while we likely can craft the regex such that its unlikely to break things we didn't feel comfortable with the neutron approach.
the explicit match of start of line might be sufficient but in general we dint want to take the risk.

i also want to have a large discussion around chartype and coaltation. i think we should be changing the type and normalising it going forward both for greenfild and adoption.

i have not looked at the galarea operator but it need to create
the db server with the default server/database correlation type set
to the same values as are used when adopting.

right now I'm using
```
CARACTER_SET=utf8
COLLATION=utf8_general_ci
```
since that is what happens to be deployed in the standalone cluster

we should likely be using
``` 
CARACTER_SET=utf8mb4
COLLATION=utf8mb4_unicode_ci
```

if your data contains emoji or other multi-bytye unicode
characters you can write the data to a charset of utf8 but it cannot be dumped and restored properly. i.e. it will break db backups.
that is why mysqldump changed its default form utf8 to utf8mb4
https://bugs.mysql.com/bug.php?id=71746

@gibizer what do you think should we recondier the sed approach.
we need to follow up on the green field approach anyway too.
I'm thinking if we are ever going to run an alter table to update
things during adoption is when we should be doing it. i have not tried to implement that as i was going to bring this up in the adoption sync tomorrow

docs/openstack/backend_services_deployment.md
```diff
@@ -167,6 +167,8 @@ podified OpenStack control plane services.
       templates:
         openstack:
           storageRequest: 500M
+        openstack-cell1:
+          storageRequest: 500M
```
gibizer, line 171: Do we need to support adopting a OSP17 multicell deployment with more than cell1? If so then this needs at least a note above that each cell needs its own DB create here.

docs/openstack/mariadb_copy.md
```diff
@@ -35,14 +35,22 @@ Define the shell variables used in the steps below. The values are
 just illustrative, use values that are correct for your environment:
 
 ```
-PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
 MARIADB_IMAGE=quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
 
-SOURCE_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
+PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
+PODIFIED_CELL1_MARIADB_IP=$(oc get -o yaml pod  mariadb-openstack-cell1  | grep podIP: | awk '{ print $2; }')
```
gibizer, line 41: if there is more than cell1 then we need a note to do this for each cell

docs/openstack/mariadb_copy.md
```diff
@@ -35,14 +35,22 @@ Define the shell variables used in the steps below. The values are
...
 PODIFIED_DB_ROOT_PASSWORD=$(oc get -o json secret/osp-secret | jq -r .data.DbRootPassword | base64 -d)
 
 # Replace with your environment's MariaDB IP:
 SOURCE_MARIADB_IP=127.17.0.100
+SOURCE_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
+
+# The CHARACTER_SET and collation should match the source DB
+# if the do not then it will break foreign key relationships
```
gibizer, line 49: > if the do not then

I think a word is missing here

docs/openstack/mariadb_copy.md
```diff
@@ -83,31 +93,60 @@ SOURCE_MARIADB_IP=127.17.0.100
...
 * Restore the databases from .sql files into the podified MariaDB:
 
   ```
-  for dbname in cinder glance keystone nova_api nova_cell0 nova ovs_neutron placement; do
-      echo "Importing $dbname"
-      oc run mariadb-client --image $MARIADB_IMAGE -i --rm --restart=Never -- \
-         mysql -h "$PODIFIED_MARIADB_IP" -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" < "$dbname.sql"
+  # db schemas to rename on import
+  declare -A db_name_map
+  db_name_map["nova"]="nova_cell1"
```
gibizer, line 112: I'm wondering what will be the naming scheme in the source DB if there are multiple cells there.

---
Also we need to note somewhere to the end user to make sure that if there is a cell db adopted with the name as nova_cell1 then there is a cellTemplate in the OpenStackControlPlane CR named "cell1".

docs/openstack/mariadb_copy.md
gibizer: I agree that if we need to run alter table then we should do it at import time.


On the sed approach:
```
stack@gibi-devstack-aio:~$ mysqldump --single-transaction --complete-insert --skip-lock-tables --lock-tables=0 --databases nova_api nova_cell1 > dump.sql
stack@gibi-devstack-aio:~$ grep 'INSERT INTO `instance_system_metadata`' dump.sql  | wc
      1     272   18749
stack@gibi-devstack-aio:~$ grep 'INSERT INTO `instance_system_metadata`' dump.sql  | head -c 100 && echo
INSERT INTO `instance_system_metadata` (`created_at`, `updated_at`, `deleted_at`, `id`, `instance_uu
```
So I think the sed that matches on the start of a line in the dump is safe as the insert into tables are a single line starting with INSERT INTO. So if sedding allows us to automatically copying the charset and collection from the source database then it might worth to do the sed.

docs/openstack/mariadb_copy.md
bogdando: what if there is already nova and nova_cell1 in the source DBs? That invalidates the mapping and nova_cell1 db will be lost?

bogdando: @stuggi if you recall tripleo cells DB naming scheme for more than >1 cell perchance?

fao89 (approved):

gibizer (approved):
I think all my comments can be fixed in a follow up so for me this is OK to merge.

Issue #86, opened 2023-08-08
Source MariaDB IP default as documented not working in documented dev env


jistr:
We try to make the default documented variables work in the documented development environment if possible. This doesn't seem to work with MariaDB in dev env:

```
SOURCE_MARIADB_IP=127.17.0.100
```

What works for me:

```
SOURCE_MARIADB_IP=192.168.122.100
```

jistr:
I think MariaDB not listening on 127.17.0.100 may be caused by stopping Haproxy in the "Stop OpenStack services" step. Perhaps instead of changing the IP we could remove Haproxy from the list of services to stop.

PR #106, opened 2023-08-13
Source MariaDB IP default patch for issue #86

cloudguruab:
The changes in this PR reflect a problem found in #86


jistr (approved):
Looks good, just the 3 commits are a bit superfluous, we should use "squash and merge" when merging.

fao89 (approved):

PR #111, opened 2023-08-16
Ceph backend configuration page, test-with-ceph target, and testing Glance with Ceph backend

jistr:


jistr:
(yet about to test it)

tests/roles/glance_adoption/tasks/main.yaml
```diff
@@ -22,6 +22,44 @@
...
+            default_backend=default_backend
+            [default_backend]
+            rbd_store_ceph_conf=/etc/ceph/ceph.conf
+            rbd_store_user=openstack
+            rbd_store_pool=images
+            store_description=Ceph glance store backend.
+          storageClass: "local-storage"
+          storageRequest: 10G
+          glanceAPIInternal:
+            externalEndpoints:
```
fmount, line 51: Looks good,  at some point we need to update this part according to [1], which is something applied to all the other operators as well. For now I think this looks very good and it's consistent with what we did in the doc. 

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/285

tests/roles/ceph_backend_configuration/tasks/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+      CEPH_CONF=$($CEPH_SSH "cat /etc/ceph/ceph.conf | base64 -w 0")
+
+- name: create ceph-conf-files secret
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    {{ ceph_backend_configuration_shell_vars }}
+
+    oc apply -f - <<EOF
```
fmount, line 16: I thought we could simplify this part by running ` oc create secret generic ceph-conf-files --from-file=/etc/ceph/ceph.client.openstack.keyring --from-file=/etc/ceph/ceph.conf` but I see you're grabbing the content by sshing to controller1 (where I assume Ceph is deployed).
The alternative would be:
- slurp files from controller1
- create secrets from files

This way you can catch issues in the file content and you don't have to convert them in `base64`, but in general I'm ok to keep this approach.

tests/playbooks/test_with_ceph.yaml
```diff
@@ -0,0 +1,28 @@
...
+  roles:
+    - prelude_local
+
+- name: Cleanup
+  hosts: local
+  gather_facts: no
+  roles:
+    - pcp_cleanup
+
+- name: Adoption
```
fmount, line 13: Do we have a CI log we can inspect?

tests/roles/glance_adoption/tasks/main.yaml
jistr: Tracking issue https://github.com/openstack-k8s-operators/data-plane-adoption/issues/124

tests/roles/ceph_backend_configuration/tasks/main.yaml
jistr: That makes sense, it's probably better to use --from-file, for now i created an issue to track it since i didn't have time to tweak & test this variant: https://github.com/openstack-k8s-operators/data-plane-adoption/issues/125

jistr:
There is no CI job for `make test-with-ceph` at this point, but i've now tested manually at least and it seems to be working. Perhaps the single virthost job that @cescgina implemented could be switched to `make test-with-ceph` after this is merged.

Marking the PR as ready.

jistr:
(Also rebased onto latest main branch to allow passing CI now (with #122) and gitignored tests/config/tmp where kustomize outputs are created.

fao89 (approved):

PR #123, opened 2023-08-25
Ensure that neutron db is adopted correctly

averdagu:
On a OSP Wallaby environment the neutron DB will be called "ovs_neutron" but neutron pod expects the database to be called "neutron"

### DIFF
--- a/docs/openstack/mariadb_copy.md
+++ b/docs/openstack/mariadb_copy.md
@@ -94,6 +94,13 @@ SOURCE_MARIADB_IP=127.17.0.100
   EOF
   ```
 
+* Before restoring the databases from .sql files, we need to ensure that the neutron
+database name is "neutron" and not "ovs_neutron" by running the command:
+
+  ```
+  sed -i '/^CREATE DATABASE/s/ovs_neutron/neutron/g;/^USE/s/ovs_neutron/neutron/g' ovs_neutron.sql
+  ```
+
 * Restore the databases from .sql files into the podified MariaDB:
 
   ```
--- a/tests/roles/mariadb_copy/tasks/main.yaml
+++ b/tests/roles/mariadb_copy/tasks/main.yaml
@@ -57,6 +57,29 @@
 
     EOF
 
+- name:  Replace neutron database name
+  ansible.builtin.shell: |
+    cd {{ mariadb_copy_tmp_dir }}
+    sed -i '/^CREATE DATABASE/s/ovs_neutron/neutron/g;/^USE/s/ovs_neutron/neutron/g' ovs_neutron.sql
+
+- name: Check name on creation
+  ansible.builtin.lineinfile:
+    path: "{{ mariadb_copy_tmp_dir }}/ovs_neutron.sql"
+    regexp: (^CREATE DATABASE.*`)(neutron)(`.*)
+    state: absent
+  check_mode: yes
+  register: create_database
+  failed_when: create_database.found == 0
+
+- name: Check name on usage
+  ansible.builtin.lineinfile:
+    path: "{{ mariadb_copy_tmp_dir }}/ovs_neutron.sql"
+    regexp: (^USE.*`)(neutron)(`.*)
+    state: absent
+  check_mode: yes
+  register: use_database
+  failed_when: use_database.found == 0
+
 - name: restore databases
   no_log: "{{ use_no_log }}"
   ansible.builtin.shell: |


averdagu:
I did only test the db-adoption section, need to adopt neutron to ensure that neutron can fetch db information.
Also, talking with @booxter maybe it'd be best to make sed more specific.

jistr:
Looks good but we should add the same thing to the test suite :)

https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/neutron_adoption/tasks/main.yaml

cescgina:
recheck

tests/roles/mariadb_copy/tasks/main.yaml
```diff
@@ -57,6 +57,15 @@
 
     EOF
 
+- name: Modify Neutron db name
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
```
booxter, line 63: do you need all the headers for a simple `sed` call? You don't call `oc` or `mariadb-client` here.

booxter: also, I see above that the other task that creates the file does `cd {{ mariadb_copy_tmp_dir }}`. Should you do the same here to make sure you are in the right directory?

tests/roles/mariadb_copy/tasks/main.yaml
```diff
@@ -57,6 +57,15 @@
...
+    {{ oc_header }}
+    {{ mariadb_copy_shell_vars }}
+
+    sed -i '/^CREATE DATABASE/s/ovs_neutron/neutron/g;/^USE/s/ovs_neutron/neutron/g' ovs_neutron.sql
```
booxter, line 67: you can use replace module from builtin to do that.

One problem with `sed` is that you assume it's installed. While it's part of POSIX, it's better to not make such assumptions. (This is a more theoretical problem.)

Another problem with `sed` is that it will return 0 even when no pattern was replaced. I think it's better to fail then than proceed. Because the missing pattern may indicate some problem with the dump. Better fail now than later?

tests/roles/mariadb_copy/tasks/main.yaml
averdagu: I don't need any of the headers. I thought it was common practice to add them to any task.
And good eyes, I'm on wrong directory, will fix on next patch! :)

tests/roles/mariadb_copy/tasks/main.yaml
averdagu: Agree, not having sed installed can be a problem. Will try different approach using some builtin module.

Regarding checking if string substitution was done correctly, since we are doing this bc we need neutron db name to be "neutron" I think it's best to check (on a different task after doing the replace) that the db name is "neutron".

But I agree on adding a check ensuring that the final value is correct.

tests/roles/mariadb_copy/tasks/main.yaml
booxter: It's not common practice for *any* script. These template substitutions will initialize some variables that are needed for `oc` and `mysql` client invocation. Since you don't use any of it, these are a waste.

tests/roles/mariadb_copy/tasks/main.yaml
jistr: Sorry for not jumping in earlier, i was on PTO last week. I'm not against merging what we have here but my feedback would be:

* If sed is pre-installed on the platforms we're interested in (RHEL, CentOS) then we can use it. If we need to specify some pre-requisite package installs for the Adoption procedure, that's fine as long as those packages come from the OS repos and not some additional ones.

* The test suite should follow the docs as closely as possible. We want to test that what we're telling users to run in the docs actually works. That can mean often using `shell` module instead of "pretty Ansible". (We talked about this only in calls, it's missing in the contrib docs, i reported #131 to fix it).

+2 as i don't want to go churning on the review :). Let's merge the DB import fix and we can refactor later.

jistr: To clarify:

> If we need to specify some pre-requisite package installs for the Adoption procedure, that's fine as long as those packages come from the OS repos and not some additional ones.

^ i wouldn't go too excessive/exotic with such package requirements but basic tools like sed should be totally fine.

jistr (approved):

jistr (commented):
The CI failure is relevant though:

```
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'mariadb_copy_temp_dir' is undefined. 'mariadb_copy_temp_dir'
```

It should be `tmp` instead of `temp`.

tests/roles/mariadb_copy/tasks/main.yaml
averdagu: Didn't know that test suite should follow docs as closely as possible, since I need to create new patch due to typo in variable, I'll address ansible bits and use sed again to mimic docs :)

jistr (approved):

docs/openstack/mariadb_copy.md
```diff
@@ -94,6 +94,13 @@ SOURCE_MARIADB_IP=127.17.0.100
   EOF
   ```
 
+* Before restoring the databases from .sql files, we need to ensure that the neutron
+database name is "neutron" and not "ovs_neutron" by running the command:
+
+  ```
+  sed -i '/^CREATE DATABASE/s/ovs_neutron/neutron/g;/^USE/s/ovs_neutron/neutron/g' ovs_neutron.sql
```
karelyatin, line 101: +1 I hit it today

Issue #112, opened 2023-08-21
Post cinder adoption, volume creation failing with  _DEFAULT_  volume type


archana203:
During cinder adoption [](https://openstack-k8s-operators.github.io/data-plane-adoption/openstack/cinder_adoption/
) we are passing default_volume_type=tripleo in customServiceConfig section of CinderApi service but the configuration is not reflected in cinder service pods and volume creation failed with following message :-

[ospng@dell-r640-089 osp2]$ openstack volume create --image cirros --bootable --size 1 disk_new
The request cannot be fulfilled as the default volume type __DEFAULT__ cannot be found. (HTTP 500) (Request-ID: req-621e986d-7220-474b-aaa2-5282a49f866c)
command terminated with exit code 1

Akrog:
We have merged PR#250 in the cinder-operator, so this bug should have been fixed.

Issue #90, opened 2023-08-08
A new page for Ceph backend configuration (parts which are common across multiple services)


jistr:
We should probably add a new document that would configure the common parts for connecting to Ceph storage backend, so that we don't have to repeat those instructions in individual service pages. And we should probably make connecting to the Ceph backend the primary documented Adoption path.

This is the content i have in mind for the "Configure external Ceph backend" page:

```
CEPH_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
CEPH_KEY=$($CEPH_SSH "cat /etc/ceph/ceph.client.openstack.keyring | base64 -w 0")
CEPH_CONF=$($CEPH_SSH "cat /etc/ceph/ceph.conf | base64 -w 0")

oc apply -f - <<EOF
apiVersion: v1
data:
  ceph.client.openstack.keyring: $CEPH_KEY
  ceph.conf: $CEPH_CONF
kind: Secret
metadata:
  name: ceph-conf-files
  namespace: openstack
type: Opaque
EOF

CEPH_FSID=$(oc get secret ceph-conf-files -o json | jq -r '.data."ceph.conf"' | base64 -d | grep fsid | sed -e 's/fsid = //')

oc patch openstackcontrolplane openstack --type=merge --patch '
spec:
  extraMounts:
    - name: v1
      region: r1
      extraVol:
        - propagation:
          - CinderVolume
          - CinderBackup
          - GlanceAPI
          extraVolType: Ceph
          volumes:
          - name: ceph
            projected:
              sources:
              - secret:
                  name: ceph-conf-files
          mounts:
          - name: ceph
            mountPath: "/etc/ceph"
            readOnly: true
'
```

fultonj:
Should we just link to https://github.com/openstack-k8s-operators/docs/blob/main/ceph.md ?

fultonj:
Perhaps this page (https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs/openstack/glance_adoption.md) should be shortened and linked to  https://github.com/openstack-k8s-operators/docs/blob/main/ceph.md

fultonj:
@jistr Of if you prefer to just create the proposed content, that's fine with me too.

jistr:
[...]

PR #129, opened 2023-08-30
Cinder: Remove duplicated cinderAPI

Akrog:
In one of the Cinder YAML examples we have `cinderAPI` twice, remove one of them.

Related #112


jistr (approved):

jistr:
@Akrog Regarding #112 FYI this is how we've configured the Ceph backend when adopting during the hackathon https://github.com/openstack-k8s-operators/data-plane-adoption/issues/90

PR #135, opened 2023-09-05
[WIP] /bin/sh is not /bin/bash

SeanMooney:
The ansible shell and command modules /bin/sh not the login
shell of a remote user.
command is alwasy /bin/sh where as the shell module
can sepcify a executable to use as the shell.

This change correct some of our missuse of the shell module
by explictly using /bin/bash in places that explictly need it
We proably should use this everywhere we currently have shell
but this change only updates the minimal set of task that
need to be changed.

this change also adds some docs for how to set
shell_header ot make most of the bash commands work
without explict code change. this will now not work in allcases
which is why some shell task are explictly updated to bash.

### DIFF
--- a/docs/contributing/tests.md
+++ b/docs/contributing/tests.md
@@ -29,3 +29,25 @@ samples are provided. To run the tests, follow this procedure:
   `tests/secrets.sample.yaml`).
 
 * Run `make test-minimal`.
+
+
+## Running tests on systems where /bin/sh is not /bin/bash
+
+ansible.builtin.command and ansible.builtin.shell use /bin/sh by default.
+for portability reason when using either command or shell modules you should
+not assume /bin/sh is bash and assume it is simply a posix compliant shell
+
+In this repo that requirement is relax for the ansible.builtin.shell
+module as we configure module_defaults
+
+```
+  module_defaults:
+    ansible.builtin.shell:
+      executable: /bin/bash
+```
+
+module_defaults are configure per playbook play so if you are adding a new
+play or playbook you will need to copy this to ensure compatibility
+
+note the command module always uses /bin/sh so we should avoid bash specific
+syntax when using command or just use the shell module instead.
--- a/tests/ansible.cfg
+++ b/tests/ansible.cfg
@@ -1,4 +1,3 @@
 [defaults]
-stdout_callback = yaml
 callback_whitelist = profile_json,profile_tasks
 roles_path = ./roles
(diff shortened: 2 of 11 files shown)


SeanMooney:
@jistr  im not really sure what the correct appcoh is to take

simlinking /bin/sh to /bin/bash is not supported and can break systems that do not expect it to be bash so that is not really an option

we could also take an approach fo running the adoption form a podman container or similar.
e.g. use a molecule env to run the adtopiton

the simplest thing would be to just set bash as the executable in all shells tasks but its not the nicest.

what do you think is a good way forward.

i also considered using the standalone VM to execute the tests playbooks via the inventory 
but it will not have oc installed

fao89:
recheck

fao89 (approved):

docs/contributing/tests.md
```diff
@@ -29,3 +29,25 @@ samples are provided. To run the tests, follow this procedure:
   `tests/secrets.sample.yaml`).
 
 * Run `make test-minimal`.
+
+
+## Running tests on systems where /bin/sh is not /bin/bash
```
jistr, line 34: I'm not sure we need this text now that it's properly defaulted? If something is not necessary to know to set the test suite up, i wouldn't put it here, to avoid "reading fatigue". It's not a blocker but IMO this may get removed in a test suite refactoring later.

tests/ansible.cfg
```diff
@@ -1,4 +1,3 @@
 [defaults]
-stdout_callback = yaml
```
jistr, line 2: I think the YAML output is much easier to read when e.g. debugging failed CI jobs... Personally i would leave this in. Not a blocker but unless there's a reason to avoid it, i think we may want to put this back in with upcoming test suite tweaks...

tests/roles/cinder_adoption/tasks/main.yaml
```diff
@@ -43,8 +43,8 @@
     {{ oc_header }}
     alias openstack="oc exec -t openstackclient -- openstack"
 
-    openstack endpoint list | grep cinder
-    openstack volume type list
+    ${BASH_ALIASES[openstack]} endpoint list | grep cinder
+    ${BASH_ALIASES[openstack]} volume type list
```
jistr, line 47: Good catch on these, thanks.

jistr (approved):

jistr:
Given that this touches a lot of files and the CI is green and my reservations are non-blocking, i'll merge and perhaps refactor the pieces i mentioned later.

Issue #118, opened 2023-08-23
Neutron adoption is missing network isolation parameters


jistr:
The template in the Neutron adoption doc should look more like this:

https://github.com/openstack-k8s-operators/openstack-operator/blob/beb079bad6c7be2ee8e18636d05055507d95f5bc/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml#L119-L129

PR #136, opened 2023-09-06
[Neutron] Add missing network isolation parameters

karelyatin:
This would be useful when accessing the adopted neutron service from the EDPM Nodes.

Closes https://github.com/openstack-k8s-operators/data-plane-adoption/issues/118

### DIFF
--- a/docs/openstack/neutron_adoption.md
+++ b/docs/openstack/neutron_adoption.md
@@ -33,6 +33,13 @@ spec:
     template:
       databaseInstance: openstack
       secret: osp-secret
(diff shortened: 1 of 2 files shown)


fao89 (commented):
could you please add this change to the test here: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/neutron_adoption/tasks/main.yaml

fao89 (approved):
thank you!

jistr (approved):
Thanks!

Issue #138, opened 2023-09-06
cinder-cfg.py generating invalid config


jistr:
We hit two issues with the cinder-cfg.py generated config:

* the serviceUser is meant to be a string rather than a list.

* the cinderVolumes section is invalid.

@karelyatin noticed that the problem with the the cinderVolumes section was incorrect networkAttachments nesting (under cinderVolumes rather than under tripleo-ceph).

After correcting these issues, Cinder seems to deploy but the cinder-volume pod is restarting.

Akrog:
I have fixed those 2 issues in PR#139 and I'm wondering if the restart is related to the missing `extraVolumes` section.

One of the patches of the PR fixes the `extraVolumes` sample not getting rendered for RBD backends.

PR #139, opened 2023-09-06
Fix cinder script

Akrog:
This PR fixes various bugs in the `cinder-cfg.py` script:

- Rendering `serviceUser` as a list instead of a string
- Rendering `networkAttachments` at the incorrect depth level
- Missing `extraMounts` when we have an RBD backend
- Incorrectly determining the default backend driver

Closes: #138


Akrog:
recheck

jistr (approved):
Thanks!

fao89 (approved):

PR #143, opened 2023-09-07
added updates to data copy procedure to include note on scc warning

cloudguruab:
This includes a note on the following: 

```
Warning: would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "example" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "example" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "database" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "example" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

### DIFF
--- a/docs/openstack/mariadb_copy.md
+++ b/docs/openstack/mariadb_copy.md
@@ -119,3 +119,9 @@ database name is "neutron" and not "ovs_neutron" by running the command:
   oc run mariadb-client --image $MARIADB_IMAGE -i --rm --restart=Never -- \
      mysql -h "$PODIFIED_MARIADB_IP" -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;'
   ```
+
+* During the pre/post checks the pod `mariadb-client` might have returned a pod security warning
+  related to the `restricted:latest` security context constraint. This is due to default security
+  context constraints and will not prevent pod creation by the admission controller. You'll see a
+  warning for the short-lived pod but it will not interfere with functionality.
+  For more info [visit here](https://learn.redhat.com/t5/DO280-Red-Hat-OpenShift/About-pod-security-standards-and-warnings/m-p/32502)


docs/openstack/mariadb_copy.md
```diff
@@ -112,3 +112,18 @@ SOURCE_MARIADB_IP=127.17.0.100
...
+
+    - For most workloads, the `restricted` SCC limits pod privileges
+      adequately. When creating a workload, you receive the warning,
+      but the workload creates pods and works correctly.
+
+    - You can also create/run your workloads by providing a full specification
+      of the security constraints to apply, without relying on the SCC to
+      provide the configuration. By following this approach, privileges are
+      more explicit, and you have greater control, so you can adjust more
+      precisely the specific privileges that are granted to workloads.
```
jistr, line 129: I would probably drop this second part as it does not apply to the MariaDB copy here. (If it did, we should probably just fix the bash snippets instead of documenting the warning.)

So the information could be a single paragraph where the gist is "You'll see a warning for the short-lived pod but it will not interfere with functionality." And for more info we could perhaps just link the article you've found, to keep the docs succinct. ( https://learn.redhat.com/t5/DO280-Red-Hat-OpenShift/About-pod-security-standards-and-warnings/m-p/32502 )

jistr:
Using the article i was actually able to get rid of the security warning this way:

```
(venv) [ospng@lenovo-p720-rdo-10 adoption-db]$ oc run mariadb-client --override-type=strategic --overrides='{"apiVersion": "v1", "spec": {"containers": [{"name": "mariadb-client", "securityContext": {"allowPrivilegeEscalation": false, "runAsNonRoot": true, "runAsUser": 42434, "runAsGroup": 42434, "capabilities": {"drop": ["ALL"]}, "seccompProfile": {"type": "RuntimeDefault"}}}]}}' --image $MARIADB_IMAGE -i --rm --restart=Never --    mysql -h "$PODIFIED_MARIADB_IP" -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;'
If you don't see a command prompt, try pressing enter.
warning: couldn't attach to pod/mariadb-client, falling back to streaming logs: Internal error occurred: error attaching to container: container is not created or running
Database
cinder
glance
information_schema
keystone
mysql
neutron
nova
nova_api
nova_cell0
performance_schema
placement
pod "mariadb-client" deleted
```

But you can see somehow that creates a new warning which was *not* present previously:

```
warning: couldn't attach to pod/mariadb-client, falling back to streaming logs: Internal error occurred: error attaching to container: container is not created or running
```

And i had to rely on digging out the numeric mysql user ID from the container. So i think we should still just document that the warning is not problematic.

jistr (approved):
Thanks

PR #146, opened 2023-09-08
Add sample env file to persist adoption settings

bogdando:
Add sample env file to persist adoption settings
    
    Provide a sample env file to recover settings after rebooting the
    hypervisor. It is mostly useful for resuming of the interrupted manual
    steps, when following the documentation. Especially, when the
    hypervisor is a VM in a cloud provider, like PSI.
    
    The helper env file ensures that the already deployed crc and edpm
    standalone host (compute 0) are running, logs in to crc, exports env
    vars and makes aliases for openstack and podman CLI to exec it on the
    standalone edpm host via ssh.
    
    This saves time, but requires the documented manual steps and the
    sample env file contents to be in sync. There is already such a
    requirement for the documentated scripts and test suits. We could
    leverage code generation in the future to update docs and tests
    automatically, when the scripted commands get updated in a commit.
    While the scripts and env files, like that sample env file, should
    then become the only source of truth.
    
    NOTE: Stability of CRC and EDPM hosts after restarting in unrelated,
    but it shows reliable enough for simple adoption scenarios.
    
    Additionally, follow-up multi-cell DB adoption patch with service
    IP selection and trivial typos fixing.

### DIFF
--- a/samples/env
+++ b/samples/env
@@ -1,0 +1,41 @@
+sudo virsh start edpm-compute-0
+crc start
+eval $(crc oc-env)
+oc login -u kubeadmin -p 12345678 https://api.crc.testing:6443
+
+export EDPM_COMPUTE_DISK_SIZE=40
+export EDPM_COMPUTE_CEPH_ENABLED=false
+export EDPM_COMPUTE_VCPUS=4
+export EDPM_COMPUTE_RAM=8
+
+export ADMIN_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AdminPassword:' | awk -F ': ' '{ print $2; }')
+export CINDER_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' CinderPassword:' | awk -F ': ' '{ print $2; }')
+export GLANCE_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' GlancePassword:' | awk -F ': ' '{ print $2; }')
+export IRONIC_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' IronicPassword:' | awk -F ': ' '{ print $2; }')
+export NEUTRON_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' NeutronPassword:' | awk -F ': ' '{ print $2; }')
+export NOVA_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' NovaPassword:' | awk -F ': ' '{ print $2; }')
+export OCTAVIA_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' OctaviaPassword:' | awk -F ': ' '{ print $2; }')
+export PLACEMENT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' PlacementPassword:' | awk -F ': ' '{ print $2; }')
+export CONTROLLER_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
+export COMPUTE_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
+export CONTROLLER1_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
+export CONTROLLER2_SSH=""
+export CONTROLLER3_SSH=""
+export PODIFIED_MARIADB_IP=$(oc get svc --selector "cr=mariadb-openstack" -ojsonpath='{.items[0].spec.clusterIP}')
+export MARIADB_IMAGE=quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
+export PODIFIED_DB_ROOT_PASSWORD=$(oc get -o json secret/osp-secret | jq -r .data.DbRootPassword | base64 -d)
+export PODIFIED_CELL1_MARIADB_IP=$(oc get svc --selector "cr=mariadb-openstack-cell1" -ojsonpath='{.items[0].spec.clusterIP}')
+export SOURCE_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
+export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:?"Enter source mariadb IP"}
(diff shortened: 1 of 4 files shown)


samples/env
```diff
@@ -0,0 +1,27 @@
...
+export PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
+export MARIADB_IMAGE=quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
+export PODIFIED_DB_ROOT_PASSWORD=$(oc get -o json secret/osp-secret | jq -r .data.DbRootPassword | base64 -d)
+export CELL1_POD_NAME=${CELL1_POD_NAME:-"mariadb-openstack-cell1"}
+export PODIFIED_CELL1_MARIADB_IP=$(oc get -o yaml pod ${CELL1_POD_NAME} | grep podIP: | awk '{ print $2; }')
+export SOURCE_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
+export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:?"Enter source mariadb IP"}
+export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:-192.168.122.100}
+
+alias podman="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100 podman"
```
bogdando, line 26: maybe not this one...

samples/env
```diff
@@ -0,0 +1,27 @@
...
+export CONTROLLER1_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
+export CONTROLLER2_SSH=""
+export CONTROLLER3_SSH=""
+export PODIFIED_MARIADB_IP=$(oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }')
+export MARIADB_IMAGE=quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
+export PODIFIED_DB_ROOT_PASSWORD=$(oc get -o json secret/osp-secret | jq -r .data.DbRootPassword | base64 -d)
+export PODIFIED_CELL1_MARIADB_IP=$(oc get svc --selector "cr=mariadb-openstack-cell1" -ojsonpath='{.items[0].spec.clusterIP})
+export SOURCE_DB_ROOT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
+export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:?"Enter source mariadb IP"}
+export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:-192.168.122.100}
```
SeanMooney, line 23: is this intentionaly 

export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:?"Enter source mariadb IP"}
export SOURCE_MARIADB_IP=${SOURCE_MARIADB_IP:-192.168.122.100}

samples/env
```diff
@@ -0,0 +1,27 @@
+sudo virsh start edpm-compute-0
+crc start
+eval $(crc oc-env)
+oc login -u kubeadmin -p 12345678 https://api.crc.testing:6443  
+
+export ADMIN_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AdminPassword:' | awk -F ': ' '{ print $2; }')
```
SeanMooney, line 6: also can yo ufafctor this out into a variable

i had script that assume they can put stuff in my home dir arbitarally

samples/env
```diff
@@ -0,0 +1,27 @@
...
+export CINDER_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' CinderPassword:' | awk -F ': ' '{ print $2; }')
+export GLANCE_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' GlancePassword:' | awk -F ': ' '{ print $2; }')
+export IRONIC_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' IronicPassword:' | awk -F ': ' '{ print $2; }')
+export NEUTRON_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' NeutronPassword:' | awk -F ': ' '{ print $2; }')
+export NOVA_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' NovaPassword:' | awk -F ': ' '{ print $2; }')
+export OCTAVIA_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' OctaviaPassword:' | awk -F ': ' '{ print $2; }')
+export PLACEMENT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' PlacementPassword:' | awk -F ': ' '{ print $2; }')
+export CONTROLLER1_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
```
SeanMooney, line 14: same for isntall yamls

all my repos are in ~/repos/

SeanMooney (changes requested):
how does this relate to the var files we use to store info for the ci

really i think we should be doing this in one way for both so it would be good to converge the too

samples/env
bogdando: sort of... A warning to tweak it, but still providing some default

samples/env
bogdando: do you mean the kubeadmin password?

samples/env
bogdando: ah I see, makes sense

bogdando: the intent although is to keep those the same as in docs, assuming the same paths used there. Not automating variables to allow fully customized runs (there are test suits for that)

samples/env
bogdando: the intent although is to keep those the same as in docs, assuming the same paths used in both places. Not automating variables to allow fully customized runs (there are test suits for that)

samples/env
bogdando: I see. The intention of having this file is not for any kind of automation, CI, or test suits in this repo.
It is to catch up with the manual adoption sessions, to eliminate the need re-entering all the export commands when resuming the manual steps from some intermediate state

bogdando:
> how does this relate to the var files we use to store info for the ci
 
It does not, please see my answer inline

bogdando:
PTAL

samples/env
bogdando: vars is not that we use in documentation, but env vars and shell commands.

Issue #132, opened 2023-09-05
The sample workload should have networking attached


jistr:
To be able to test workload networking aspects, the sample workload that we are creating prior to Adoption [1] should have a port on a private network attached, and the network should probably have a router on it connecting it with a provider network.

[1] https://openstack-k8s-operators.github.io/data-plane-adoption/contributing/development_environment/#create-a-workload-to-adopt

PR #149, opened 2023-09-11
[Workload Setup] reuse edpm-deploy-instance.sh script

karelyatin:
The script creates an instance with all the dependencies, let's reuse the script.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/530 Closes https://github.com/openstack-k8s-operators/data-plane-adoption/issues/132

### DIFF
--- a/docs/contributing/development_environment.md
+++ b/docs/contributing/development_environment.md
@@ -112,22 +112,9 @@ the section:
 
 ### Create a workload to adopt
 
-For this example we'll upload a Glance image and confirm it's using
-the Ceph cluster. Later you can adopt the Glance image in the NG
-deployment.
-
-Download a cirros image and convert it to raw format for Ceph.
-```
-IMG=cirros-0.5.2-x86_64-disk.img
-URL=http://download.cirros-cloud.net/0.5.2/$IMG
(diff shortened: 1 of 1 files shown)


fao89 (approved):

SeanMooney:
That script is inteded to be used with a greenfiled deployment
for adoption all the networks and images should be created by tripleo before the adoption
the install yamlls script should not be used with the tripleo deploy OpenStack so I'm not really sure this is a good idea.

karelyatin:
<< That script is inteded to be used with a greenfiled deployment
<< for adoption all the networks and images should be created by tripleo before the adoption
<< the install yamlls script should not be used with the tripleo deploy OpenStack so I'm not really sure this is a good idea.

@SeanMooney I reused that to avoid duplicated steps being done in workload preparation for adoption(environment prepared as part of "make standalone").
What part of tripleo handles images/network preparation as standalone setup didn't had those pre created? iirc there are some tempest roles that used to create it in TripleO jobs.
Me not sure of all the use cases but i noticed many scripts used for adoption document are from install_yamls.

jistr (approved):

PR #151, opened 2023-09-12
Update sample vars and secrets and test doc

bogdando:
Add missing dependencies insntallation step into testing docs (based on the adoption hackathon demos recoreded)

Improve samples default values so that user can make less manual editions to start the test suits.

### DIFF
--- a/tests/vars.sample.yaml
+++ b/tests/vars.sample.yaml
@@ -1,23 +1,24 @@
 # Path to install_yamls repo
-install_yamls_path: /CUSTOMIZE_THIS
+install_yamls_path: ~/install_yamls #CUSTOMIZE_THIS
 # Whether to use 'make crc_storage_cleanup; make crc_storage' before the test
 reset_crc_storage: true
 
 # Snippet to get the desired 'oc' command onto $PATH.
 oc_header: |
-  source CUSTOMIZE_THIS.sh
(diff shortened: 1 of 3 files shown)


bogdando:
still testing this

tests/vars.sample.yaml
```diff
@@ -1,23 +1,24 @@
...
-source_ovndb_ip: CUSTOMIZE_THIS
+source_ovndb_ip: 192.168.122.100 #CUSTOMIZE_THIS
 
 # EDPM node IP
-edpm_node_ip: CUSTOMIZE_THIS
+edpm_node_ip: 192.168.122.100 #CUSTOMIZE_THIS
 
 # NTP servers list
 edpm_chrony_ntp_servers:
+  - clock.redhat.com
```
jistr, line 21: I don't think we should default to this in public code... E.g. the upstream CI won't be able to reach that. Other than this LGTM, this PR is a nice UX improvement for local testing, thanks :+1:

bogdando: commented it out, but it is crucial to use it when testing from the intranet, I can recall multiple problems related to missing that source from downstream tripleo background

jistr (approved):
Thanks!

fao89 (approved):

PR #152, opened 2023-09-12
Filter by mariadb service cluster IP instead

bogdando:
Filtered pod IP may change, if pod restarts. Use the service cluster IP which never changes.

### DIFF
--- a/tests/roles/mariadb_copy/tasks/main.yaml
+++ b/tests/roles/mariadb_copy/tasks/main.yaml
@@ -1,15 +1,15 @@
-- name: get podified MariaDB IP
+- name: get podified MariaDB service cluster IP
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }'
+    oc get svc --selector "cr=mariadb-openstack" -ojsonpath='{.items[0].spec.clusterIP}'
   register: podified_mariadb_ip_result
 
 - name: get podified cell1 MariaDB IP
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc get -o yaml pod mariadb-openstack-cell1 | grep podIP: | awk '{ print $2; }'
+    oc get svc --selector "cr=mariadb-openstack-cell1" -ojsonpath='{.items[0].spec.clusterIP}'
   register: podified_cell1_mariadb_ip_result
 
 - name: set MariaDB copy shell vars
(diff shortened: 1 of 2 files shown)


tests/roles/mariadb_copy/tasks/main.yaml
```diff
@@ -1,8 +1,8 @@
-- name: get podified MariaDB IP
+- name: get podified MariaDB service cluster IP
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc get -o yaml pod mariadb-openstack | grep podIP: | awk '{ print $2; }'
+    oc get svc --selector "cr=mariadb-openstack" -ojsonpath='{.items[0].spec.clusterIP}'
```
jistr, line 5: Won't this break if we have 2 MariaDB instances? (Wouldn't cell 1 instance also match the CR selector?)

+1 on the clusterIP but i think we should keep selecting by pod name?

jistr (commented):
+1 on the direction, in addition to the inline comment please let's make a matching amendment to the docs as well.

Also, it may be worth coordinating with Sean regarding the order in which you guys would like to merge PRs that touch the MariaDB procedure.

bogdando:
> +1 on the direction, in addition to the inline comment please let's make a matching amendment to the docs as well.
> 
> Also, it may be worth coordinating with Sean regarding the order in which you guys would like to merge PRs that touch the MariaDB procedure.

please let's go with Sean's PR first, then I'll rebase

tests/roles/mariadb_copy/tasks/main.yaml
bogdando: I checked on my env, it works OK for main and cell db svc

jistr: :+1: thanks

bogdando:
/hold
should come after https://github.com/openstack-k8s-operators/data-plane-adoption/pull/102

fao89 (approved):

fao89 (commented):
Actually, I think we also need to update it here: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/bc58d557bf1f70c767e5c299de9ce9d72a49decb/docs/openstack/mariadb_copy.md?plain=1#L40C1-L40C1

bogdando:
> Actually, I think we also need to update it here: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/bc58d557bf1f70c767e5c299de9ce9d72a49decb/docs/openstack/mariadb_copy.md?plain=1#L40C1-L40C1

done

bogdando:
rebased on top of the recent cells db as well

bogdando:
@SeanMooney @jistr PTAL

jistr (approved):

PR #153, opened 2023-09-13
Move nodes to top level of nodeset spec

fao89:
ref: https://github.com/openstack-k8s-operators/dataplane-operator/pull/409


fao89:
recheck

jistr:
:+1: Thanks

fao89:
recheck

PR #157, opened 2023-09-18
Rely on automatic ConfigMap/Secret configuration of EDPM OVN services

fao89:
ref: https://github.com/openstack-k8s-operators/dataplane-operator/pull/412

### DIFF
--- a/docs/openstack/edpm_adoption.md
+++ b/docs/openstack/edpm_adoption.md
@@ -95,6 +95,19 @@ EOF
   $(cat ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa | base64 | sed 's/^/        /')
   EOF
   ```
+* Create the Nova Metadata secret (Workaround while nova isn't adopted yet):
+
+  ```bash
+  oc apply -f - <<EOF
+  apiVersion: v1
+  kind: Secret
+  metadata:
+      name: nova-metadata-neutron-config
+  data:
+      05-nova-metadata.conf: |
+  $(echo "[DEFAULT]\nnova_metadata_host = 1.2.3.4\nnova_metadata_port = 8775\nnova_metadata_protocol = http\nmetadata_proxy_shared_secret = 1234567842\n" | base64 | sed 's/^/        /')
+  EOF
+  ```
 
 * Stop the nova services.
 
@@ -143,6 +156,7 @@ done
         - ctlplane
     preProvisioned: true
     services:
+      - download-cache
       - configure-network
       - validate-network
       - install-os
@@ -207,16 +221,9 @@ done
           edpm_nodes_validation_validate_controllers_icmp: false
           edpm_nodes_validation_validate_gateway_icmp: false
 
-          edpm_ovn_metadata_agent_DEFAULT_transport_url: $(oc get secret rabbitmq-transport-url-neutron-neutron-transport -o json | jq -r .data.transport_url | base64 -d)
-          edpm_ovn_metadata_agent_metadata_agent_ovn_ovn_sb_connection: $(oc get ovndbcluster ovndbcluster-sb -o json | jq -r .status.dbAddress)
-          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_nova_metadata_host: $(oc get svc nova-metadata-internal -o json |jq -r '.status.loadBalancer.ingress[0].ip')
-          edpm_ovn_metadata_agent_metadata_agent_DEFAULT_metadata_proxy_shared_secret: $(oc get secret osp-secret -o json | jq -r .data.MetadataSecret  | base64 -d)
-          edpm_ovn_metadata_agent_DEFAULT_bind_host: 127.0.0.1
           edpm_chrony_ntp_servers:
           - clock.redhat.com
           - clock2.redhat.com
-
-          edpm_ovn_dbs: $(oc get ovndbcluster ovndbcluster-sb -o json | jq -r '.status.networkAttachments."openstack/internalapi"')
 
           edpm_ovn_controller_agent_image: quay.io/podified-antelope-centos9/openstack-ovn-controller:current-podified
           edpm_iscsid_image: quay.io/podified-antelope-centos9/openstack-iscsid:current-podified
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -97,6 +97,21 @@
         ssh-privatekey: "{{ edpm_encoded_privatekey|default(edpm_privatekey.content) }}"
     EOF
 
+- name: "[hack] create nova-metadata-neutron-config-secret.yaml"
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc apply -f - <<EOF
+    apiVersion: v1
+    kind: Secret
+    metadata:
+        name: nova-metadata-neutron-config
+    data:
+        05-nova-metadata.conf: |
+    $(echo "[DEFAULT]\nnova_metadata_host = 1.2.3.4\nnova_metadata_port = 8775\nnova_metadata_protocol = http\nmetadata_proxy_shared_secret = 1234567842\n" | base64 | sed 's/^/        /')
+    EOF
+
 - name: set Nova copy shell vars
   no_log: "{{ use_no_log }}"
   ansible.builtin.set_fact:
@@ -153,6 +168,7 @@
         - ctlplane
       preProvisioned: true
       services:
+        - download-cache
         - configure-network
         - validate-network
         - install-os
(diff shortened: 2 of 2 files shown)


fao89:
recheck

fao89:
recheck

fao89:
Actually, the CI failed because Neutron is expecting Nova:
https://github.com/openstack-k8s-operators/neutron-operator/blob/8ef2e76869c3580aaa1a8f99cd95f6a2e8a9c8e2/controllers/neutronapi_controller.go#L633-L637
and we haven't adopted Nova yet

fao89:
@booxter @gibizer does Neutron can know whether it can expect a secret from Nova?
Currently, Nova isn't adopted yet so it isn't enabled in the Adoption CI, the CI logs aren't much helpful but I suspect Neutron is failing because Nova isn't running

gibizer:
> @booxter @gibizer does Neutron can know whether it can expect a secret from Nova? Currently, Nova isn't adopted yet so it isn't enabled in the Adoption CI, the CI logs aren't much helpful but I suspect Neutron is failing because Nova isn't running

neutron-operator does not depend on the secret from nova-operator. Only the DataPlaneService representing the neutron-metadata-agent depends on a secret from nova-operator.

docs/openstack/edpm_adoption.md
```diff
@@ -95,6 +95,18 @@ EOF
...
+* Create the Nova Cell secret (Workaround while nova isn't adopted yet):
+
+  ```bash
+  oc apply -f - <<EOF
+  apiVersion: v1
+  kind: Secret
+  metadata:
+      name: nova-metadata-neutron-config
+  data:
+      05-nova-metadata.conf: $("[DEFAULT]\nnova_metadata_host = http://nova-metadata-internal.openstack.svc:8775\nmetadata_proxy_shared_secret = 1234567842\n" | base64)
```
booxter, line 107: I don't think this is correct, the format is

```
[DEFAULT]
nova_metadata_host = 1.2.3.4
nova_metadata_port = 8775
nova_metadata_protocol = http
metadata_proxy_shared_secret = 12345678
```

booxter (changes requested):
Please fix the contents of the snippet included in the Secret to reflect what metadata agent actually consumes.

docs/openstack/edpm_adoption.md
fao89: I got it from a running env:
```
[dpadev@osp-dev-05 devsetup]$ oc get secret nova-metadata-neutron-config -o json | jq '.data | map_values(@base64d)' 
{
  "05-nova-metadata.conf": "[DEFAULT]\nnova_metadata_host = http://nova-metadata-internal.openstack.svc:8775\nmetadata_proxy_shared_secret = 1234567842\n",
  "httpd.conf": "ServerTokens Prod\nServerSignature Off\nTraceEnable Off\nPidFile run/httpd.pid\nServerRoot \"/etc/httpd\"\nServerName \"localhost.localdomain\"\n\nUser apache\nGroup apache\n\nListen 8775\n\nTypesConfig /etc/mime.types\n\nInclude conf.modules.d/*.conf\n\nLogFormat \"%h %l %u %t \\\"%r\\\" %>s %b \\\"%{Referer}i\\\" \\\"%{User-Agent}i\\\"\" combined\nLogFormat \"%{X-Forwarded-For}i %l %u %t \\\"%r\\\" %>s %b \\\"%{Referer}i\\\" \\\"%{User-Agent}i\\\"\" proxy\n\nSetEnvIf X-Forwarded-For \"^.*\\..*\\..*\\..*\" forwarded\nCustomLog /dev/stdout combined env=!forwarded\nCustomLog /dev/stdout proxy env=forwarded\n## set default apache log level to info from warning\nLogLevel info\n\n<VirtualHost *:8775>\n  <IfVersion >= 2.4>\n    ErrorLogFormat \"%M\"\n  </IfVersion>\n  SetEnvIf X-Forwarded-For \"^.*\\..*\\..*\\..*\" forwarded\n  ErrorLog /dev/stdout\n  CustomLog /dev/stdout combined env=!forwarded\n  CustomLog /dev/stdout proxy env=forwarded\n  ## set nova vhost log level to debug\n  LogLevel debug\n\n  ## WSGI configuration\n  WSGIProcessGroup nova-metadata\n  WSGIApplicationGroup %{GLOBAL}\n  WSGIPassAuthorization On\n  ## In general we want nova-metadata to scale via k8s replicas but we need\n  ## two processes per replica to always has a room for a healthecheck query\n  WSGIDaemonProcess nova-metadata processes=2 threads=1 user=nova group=nova display-name=nova-metadata-api\n  WSGIScriptAlias / /usr/bin/nova-metadata-wsgi\n</VirtualHost>\n\nAlias /nova-metadata /usr/bin/nova-metadata-wsgi\n<Location /nova-metadata>\n  SetHandler wsgi-script\n  Options +ExecCGI\n  WSGIProcessGroup nova-metadata\n  WSGIApplicationGroup %{GLOBAL}\n  WSGIPassAuthorization On\n</Location>\n",
  "nova-metadata-config.json": "{\n  \"command\": \"/usr/sbin/httpd -DFOREGROUND\",\n  \"config_files\": [\n    {\n      \"source\": \"/var/lib/openstack/config/nova-blank.conf\",\n      \"dest\": \"/etc/nova/nova.conf\",\n      \"owner\": \"nova\",\n      \"perm\": \"0600\"\n  },\n  {\n      \"source\": \"/var/lib/openstack/config/01-nova.conf\",\n      \"dest\": \"/etc/nova/nova.conf.d/01-nova.conf\",\n      \"owner\": \"nova\",\n      \"perm\": \"0600\"\n  },\n  {\n      \"source\": \"/var/lib/openstack/config/02-nova-override.conf\",\n      \"dest\": \"/etc/nova/nova.conf.d/02-nova-override.conf\",\n      \"owner\": \"nova\",\n      \"perm\": \"0600\",\n      \"optional\": true\n  },\n    {\n      \"source\": \"/var/lib/openstack/config/httpd.conf\",\n      \"dest\": \"/etc/httpd/conf/httpd.conf\",\n      \"owner\": \"apache\",\n      \"perm\": \"0644\"\n    }\n  ],\n  \"permissions\": [\n    {\n      \"path\": \"/var/log/nova\",\n      \"owner\": \"nova:apache\",\n      \"recurse\": true\n    }\n  ]\n}\n"
}
```
and the CI got green

booxter: It got "green" because ansibleee role was executed and the container for the agent was created. But if you look at the logs for the agent, you will see that it crashes over and over because it can't parse the config option value for nova_metadata_host (that should be a host, without protocol / port). I am not sure where to find the logs for EDP service containers in adoption CI artifacts though. Do you have an idea?

fao89: it is under "controller": https://logserver.rdoproject.org/57/157/fbb5a720a18ee16a2c442aa0686de2e4302e20bc/github-check/data-plane-adoption-github-rdo-centos-9-crc-single-node/d89f7af/controller/pod/

booxter: As to the fact that you got it from a running env, this is because there's a bug (fixed now) in nova-operator, see https://github.com/openstack-k8s-operators/nova-operator/pull/534

fao89: gotcha!
I'll update the PR

booxter: > it is under "controller"
I am not sure. These seem like pod logs. I am talking about ovn_metadata_agent container running on EDP nodes (not on CP nodes).

fao89: yes, https://review.rdoproject.org/r/c/rdo-jobs/+/50082 was wrong and didn't catch those logs, I've updated it now
we will have the logs in the next run

booxter: @fao89 this is helpful, but I think it still won't capture the logs for actual containers that run services like ovn-controller or metadata-agent on EDP nodes. Your patch will collect logs from pods that run operators, not from containers running on EDP.

fao89: do you want something like this? https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/artifacts/tasks/edpm.yml#L58

booxter: @fao89 yes, this would do.

cescgina (approved):

jistr (approved):

docs/openstack/edpm_adoption.md
fao89: @cescgina do you know how can we collect these logs from the standalone node?

cescgina: @fao89 we could add an extra task here https://review.rdoproject.org/r/plugins/gitiles/rdo-jobs/+/refs/heads/master/playbooks/data_plane_adoption/standalone_collect_logs.yaml to run the needed comands and pull the logs out of the standalone vm/node

PR #158, opened 2023-09-18
Use cluster service network IPs for OVN sb/nb and switch from podman and hostpath to oc run/exec and pvc

bogdando:
Pod IP may change, invalidating the automated test suit execution.
OVN DB south/north-bound endpoints should be as well accessible on
the standard k8s clusterIP services network. Use that instead of
fetching IPs on internalAPI openstack network.

Switch from podman to oc run, in order to access the clusterIP
service network. This also solves connectivity problems when
deploying openstack CR that uses isolnet, and podman fails
accessing IPs on internalAPI service network. Use storageclass
(defaults to crc-csi-hostpath-provisioner) insetad of Instead of local
host directories.

Add storage_reclaim_policy and storage_class_name ansible var to
control that PVC behavior.                                                                                                                                                       

Use force_handlers=true for the minimal test playbook so
that the handlers trigger on errors as well.

Partial JIRA https://issues.redhat.com/browse/OSPRH-365

### DIFF
--- a/docs/openstack/ovn_adoption.md
+++ b/docs/openstack/ovn_adoption.md
@@ -31,6 +31,7 @@ Define the shell variables used in the steps below. The values are
 just illustrative, use values that are correct for your environment:
 
 ```bash
+STORAGE_CLASS_NAME=crc-csi-hostpath-provisioner
 OVSDB_IMAGE=quay.io/podified-antelope-centos9/openstack-ovn-base:current-podified
 SOURCE_OVSDB_IP=172.17.1.49
 
@@ -54,13 +55,64 @@ grep -rI 'ovn_[ns]b_conn' /var/lib/config-data/puppet-generated/
 ```bash
 ${CONTROLLER_SSH} sudo systemctl stop tripleo_ovn_cluster_northd.service
 ```
+- Prepare the OVN DBs copy dir and the adoption helper pod (pick the storage requests to fit the OVN databases sizes)
+
+```yaml
+oc apply -f - <<EOF
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: ovn-data
+spec:
+  storageClassName: $STORAGE_CLASS_NAME
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ovn-copy-data
+  annotations:
+    openshift.io/scc: anyuid
+  labels:
+    app: adoption
+spec:
+  containers:
+  - image: $OVSDB_IMAGE
+    command: [ "sh", "-c", "sleep infinity"]
+    name: adoption
+    volumeMounts:
+    - mountPath: /backup
+      name: ovn-data
+  securityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop: ALL
+    runAsNonRoot: true
+    seccompProfile:
+      type: RuntimeDefault
+  volumes:
+  - name: ovn-data
+    persistentVolumeClaim:
+      claimName: ovn-data
+EOF
+```
+
+- Wait for the pod to come up
+
+```bash
+oc wait --for=condition=Ready pod/ovn-copy-data --timeout=30s
+```
 
 - Backup OVN databases.
 
 ```bash
-client="podman run -i --rm --userns=keep-id -u $UID -v $PWD:$PWD:z,rw -w $PWD $OVSDB_IMAGE ovsdb-client"
-${client} backup tcp:$SOURCE_OVSDB_IP:6641 > ovs-nb.db
-${client} backup tcp:$SOURCE_OVSDB_IP:6642 > ovs-sb.db
+oc exec ovn-copy-data -- bash -c "ovsdb-client backup tcp:$SOURCE_OVSDB_IP:6641 > /backup/ovs-nb.db"
+oc exec ovn-copy-data -- bash -c "ovsdb-client backup tcp:$SOURCE_OVSDB_IP:6642 > /backup/ovs-sb.db"
 ```
 
 - Start podified OVN database services prior to import.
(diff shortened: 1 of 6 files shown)


bogdando:
Note, if we want to verify connectivity exactly via internaAPI, let's make sure we check a VIP that never changes, instead of pod IPs on isolated networks

booxter:
There are plans to introduce stable IPs on internalapi for SB (but not for NB - I actually want to detach NB from internalapi completely unless there's a good reason to have it exposed), but 1) it's not there yet and 2) adoption should work fine with restoring using cluster IPs.

bogdando:
tested on my adoption env with a custom CR

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -73,6 +69,30 @@
...
+          volumeMounts:
+          - mountPath: /backup
+            name: backup
+        template:
+          volumes:
+          - hostPath:
+              path: "{{ ovn_copy_tmp_dir }}"
+              type: Directory
+            name: backup
+
```
SeanMooney, line 88: so this might work for crc but its probably not ok for anything else
can we just use a PVC instead.

SeanMooney: bacialy we shoudl just create a volume with the same storage class we use for the control plan

yes that happens to ba a local volume with the default crc setup but
we should not actually force it tobe a hostPath volume

bogdando:
now looks correct. Will address comments and update

tests/roles/ovn_adoption/tasks/main.yaml
bogdando: that's a good idea, although is out of the patch scope, and belongs to future refactoring. But could be a single unit of work as soon as we're going to switch podman for oc run/exec globally (and start using the volumes, indeed)

SeanMooney: well honestly i would not merge this as is until this is fixed which is why i left the comment os i would like to see what other think

booxter:
/lgtm

I think it's ok to leave volume switch to a separate PR.

bogdando:
@jistr I guess then it's good to go

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -29,31 +29,77 @@
...
+- name: set fact for oc run command
+  ansible.builtin.set_fact:
+    oc_run_cmd: >-
+      oc delete pod adoption --ignore-not-found=true;
+      oc run adoption
+      --image {{ ovn_image }}
+      --overrides='$(cat {{ playbook_dir }}/{{ ovn_copy_tmp_dir }}/oc_run_args.yaml)';
+      oc wait --for=condition=Ready pod adoption --timeout=30s
+
+- name: start an adoption helper pod
```
averdagu, line 86: Shouldn't it be removed once the adoption had been completed?

bogdando: it should be, right, and there is a handler to do so. But the teardown needs improvement (the pod remains if playbooks exits with errors)

bogdando: so I've added that pre-start command to remove the helper pod left behind in such a failure scenario. That looks simpler than changing all the sections with rescue blocks to remove it. We can think on a better solution

averdagu: Thanks! After looking a bit I discovered that you can add "force_handlers: true" to a task to ensure that even if there were some error during the play the handler will run. 

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#handlers-and-failure

Just sharing it, I like the current implementation :) Thanks!

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -29,31 +29,77 @@
...
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc get pod ovsdbserver-sb-0 -o jsonpath='{.metadata.annotations.k8s\.v1\.cni\.cncf\.io/network-status}' | \
-    jq 'map(. | select(.name=="openstack/internalapi"))[0].ips[0]' | \
-    tr -d '"'
+    oc get svc --selector "statefulset.kubernetes.io/pod-name=ovsdbserver-sb-0" -ojsonpath='{.items[0].spec.clusterIP}'
   register: podified_ovn_sb_ip_result
 
+- name: set fact for oc run args
```
jistr, line 46: This and the relevant `oc exec` part is not reflected in the docs. We should keep the parity, ideally with verbatim copy/paste (exact same content in docs and tests as much as possible -- using shell vars for parametrization where necessary).

I will also bring this up to others to make sure we're comfortable with the `oc exec` approach in general. This is a bit double bladed, it does make the manual workflow more complex, but the procedure also becomes better insulated from possible quirks of the "users laptop" where the procedure is being executed, so i'm leaning +1.

jistr (changes requested):

tests/roles/ovn_adoption/tasks/main.yaml
fmount: +1

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -29,31 +29,77 @@
...
 
-- name: get podified OVN NB ovsdb-server IP
+- name: get podified OVN NB ovsdb-server service cluster IP
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc get pod ovsdbserver-nb-0 -o jsonpath='{.metadata.annotations.k8s\.v1\.cni\.cncf\.io/network-status}' | \
-    jq 'map(. | select(.name=="openstack/internalapi"))[0].ips[0]' | \
-    tr -d '"'
+    oc get svc --selector "statefulset.kubernetes.io/pod-name=ovsdbserver-nb-0" -ojsonpath='{.items[0].spec.clusterIP}'
```
cescgina, line 36: This is failing in CI because it seems that it takes a few seconds to assing this IP to the pod and it's not ready when it queries it. I tested it in a held node  adding a few retries like in the task above and it worked after a couple of attempts

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -29,31 +29,77 @@
...
+        volumes:
+        - hostPath:
+            path: "{{ playbook_dir }}/{{ ovn_copy_tmp_dir }}"
+            type: Directory
+          name: backup
+
+- name: prepare oc run args file
+  copy:
+    content: "{{ oc_run_args }}"
+    dest: "{{ ovn_copy_tmp_dir }}/oc_run_args.yaml"
```
cescgina, line 74: the `ovn_copy_tmp_dir` is not created until the `dump OVN databases` task later on, we need to move the creation before this task

tests/roles/ovn_adoption/tasks/main.yaml
bogdando: this is weird as we wait for the pod's status.phase==Running

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -83,21 +129,17 @@
     mkdir -p {{ ovn_copy_tmp_dir }}
     cd {{ ovn_copy_tmp_dir }}
 
-    client="podman run -i --rm --userns=keep-id -u $UID -v $PWD:$PWD:z,rw -w $PWD $OVSDB_IMAGE ovsdb-client"
-    ${client} backup tcp:$SOURCE_OVSDB_IP:6641 > ovs-nb.db
-    ${client} backup tcp:$SOURCE_OVSDB_IP:6642 > ovs-sb.db
-
-- name: upgrade OVN databases to the latest schema from podified ovsdb-servers
```
cescgina, line 90: I think this line was deleted by mistake?

tests/roles/ovn_adoption/tasks/main.yaml
bogdando: but I'm updating other places now, let's keep it synced, right

bogdando:
PTAL, the minimal test suit has passed the ovn part, also I think I've addressed all the comments

cescgina:
recheck

docs/openstack/ovn_adoption.md
```diff
@@ -48,13 +48,50 @@ COMPUTE_SSH="ssh -F ~/director_standalone/vagrant_ssh_config vagrant@standalone"
...
+  securityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop: ALL
+    runAsNonRoot: true
+    runAsUser: 1000
+    seccompProfile:
+      type: RuntimeDefault
+  volumes:
+  - hostPath:
```
jistr, line 77: As Sean previously commented, can we avoid using hostPath? Perhaps we should use a PVC like we use for other persistent control plane storage? We should aim for something that would map well to production use.

tests/roles/ovn_adoption/handlers/main.yaml
```diff
@@ -0,0 +1,2 @@
+- name: stop adoption helper pod
+  command: oc delete pod adoption --ignore-not-found=true
```
jistr, line 2: We should allow for "debug mode" where the pod would stay around if the test suite fails, or maybe even just if the user wants it to. But this one is not a blocker and can be done in a different PR.

jistr (changes requested):

tests/roles/ovn_adoption/handlers/main.yaml
bogdando: ack, let's do in follow ups

bogdando:
this worked now on my testing env

tests/roles/ovn_adoption/handlers/main.yaml
bogdando: done in the updated revision

bogdando:
recheck no logs collected

bogdando:
the pod waiting by ``--selector=service=ovsdbserver-nb`` failed immediately, w/o retrying it. That's not what I've got on my testing env, where this just worked OK...

bogdando:
PTAL, we can merge this

bogdando:
@booxter could you please re-approve?

bogdando:
recheck

bogdando:
@jistr please merge

fao89 (approved):

PR #159, opened 2023-09-19
Adding OpenStackDataPlaneDeployment CR

fao89:


fao89:
merging as it fixes the CI

PR #161, opened 2023-09-20
Run sudo cephadm shell from the Standalone VM

matbu:


jistr:
+1 on that edit but i think we were mainly interested in the other occurrence of the `cephadm shell` command which is in the "Create a workload to adopt" section:

https://openstack-k8s-operators.github.io/data-plane-adoption/contributing/development_environment/#create-a-workload-to-adopt

jistr (changes requested):
(commented above)

jistr (approved):
The other part has been done in #163, so we can merge this.

fao89 (approved):

Issue #133, opened 2023-09-05
"sudo cephadm shell" command in dev env setup needs to be run from the standalone VM


jistr:
Instead of
```
sudo cephadm shell -- rbd -p images ls -l
```

we should have
```
ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100 sudo cephadm shell -- rbd -p images ls -l
```

Or we can perhaps drop the command from there entirely because it's a "just in case" check that we haven't seen fail yet.

matbu:
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/161

Issue #114, opened 2023-08-22
Improve Development environment docs by making explicit which commands should run on host/VM


fao89:
The following section should be executed inside the standalone node:
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs/contributing/development_environment.md#create-a-workload-to-adopt
`ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100`

jistr:
This is a duplicate of #133. This one came earlier but i'd leave #133 open as it goes into more depth about the problem that we hit during the hackathon.

PR #163, opened 2023-09-20
set explicit commands to be executed in standalone node

cloudguruab:


fao89 (approved):

jistr:
Not a -1 but i'm curious what doesn't work on the previous solution? We do have `alias openstack="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100 OS_CLOUD=standalone openstack"` in effect i think, so any `openstack` commands should work?

docs/contributing/development_environment.md
```diff
@@ -111,11 +111,18 @@ the section:
...
 ```
+ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100
 export OS_CLOUD=standalone
-source ~/install_yamls/devsetup/scripts/edpm-deploy-instance.sh
+source ~/edpm-deploy-instance.sh
 ```
+
 Confirm the image UUID can be seen in Ceph's images pool.
 ```
 sudo cephadm shell -- rbd -p images ls -l
```
jistr, line 128: But the cephadm command here isn't fixed and instead we're editing things above it?

jistr: Ah yes there is #161 which aims to fix #133. I'm not sure if #114 meant we need to move the edpm-deploy-instance.sh. It may be that those are actually duplicate issues and we have duplicate PRs but neither is editing the intended spot? :)

jistr: Ah yes #114 was reported at a time when this `edpm-deploy-instance.sh` part didn't exist yet, that's why it's not more specific on what needs to be done. So to my knowledge we only need to fix that `cephadm shell` part as #133 suggests. And none of the open PRs presently address that part. So IIUC we only need a oneliner edit there.

PR #148, opened 2023-09-11
Fix container images used in periodic CI job for some services

marios:
Some of the podified openstack services are still using the current-podified tag for the container image in the CI periodic line. This adds the overrides for glance and ovn.

PR #165, opened 2023-09-22
remove hard-coded container images from docs and check tests

cloudguruab:
Wondering if we want to update hard coded images [here](https://openstack-k8s-operators.github.io/data-plane-adoption/openstack/cinder_adoption/) as well? They seem to be more explicit for the surrounding context/example that covers these container image references, if it's not the case I'll migrate to using same approach with other `containerImage` formats.

### DIFF
--- a/docs/openstack/glance_adoption.md
+++ b/docs/openstack/glance_adoption.md
@@ -80,7 +80,6 @@ spec:
     enabled: true
     template:
       databaseInstance: openstack
-      containerImage: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-glance-api:{{ container_tag | default("current-podified") }}
       customServiceConfig: |
         [DEFAULT]
         enabled_backends=default_backend:rbd
--- a/docs/openstack/ovn_adoption.md
+++ b/docs/openstack/ovn_adoption.md
@@ -67,12 +67,10 @@ spec:
     template:
       ovnDBCluster:
         ovndbcluster-nb:
-          containerImage: {{ container_registry }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-ovn-nb-db-server:{{ container_tag | default("current-podified") }}
           dbType: NB
           storageRequest: 10G
           networkAttachment: internalapi
         ovndbcluster-sb:
-          containerImage: {{ container_registry }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-ovn-sb-db-server:{{ container_tag | default("current-podified") }}
           dbType: SB
           storageRequest: 10G
           networkAttachment: internalapi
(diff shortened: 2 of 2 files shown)


docs/openstack/edpm_adoption.md
```diff
@@ -223,12 +223,12 @@ done
...
+          edpm_iscsid_image: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-iscsid:{{ container_tag | default("current-podified") }}
+          edpm_logrotate_crond_image: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-cron:{{ container_tag | default("current-podified") }}
+          edpm_nova_compute_container_image: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-nova-compute:{{ container_tag | default("current-podified") }}
+          edpm_nova_libvirt_container_image: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-nova-libvirt:{{ container_tag | default("current-podified") }}
+          edpm_ovn_metadata_agent_image: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-neutron-metadata-agent-ovn:{{ container_tag | default("current-podified") }}
```
jistr, line 231: The jinja variables can't be used in documentation. If these variables aren't defaulted inside EDPM then we'll need to keep them spelled out in docs.

docs/openstack/glance_adoption.md
```diff
@@ -74,7 +74,7 @@ spec:
     enabled: true
     template:
       databaseInstance: openstack
-      containerImage: quay.io/podified-antelope-centos9/openstack-glance-api:current-podified
+      containerImage: {{ container_registry|default("quay.io") }}/{{ container_namespace|default("podified-antelope-centos9") }}/openstack-glance-api:{{ container_tag | default("current-podified") }}
```
jistr, line 77: Ditto here, we can't use jinja in docs. So this issue was about automation (test suite) mainly. I think it got fixed here but we forgot to close the issue: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/148/files

jistr: Or actually, we may be able to remove the hardcoded images from docs by removing the `containerImage` lines completely. I think the operators should be defaulting to the correct images. I think that's what i had in mind when i reported the issue. I'm not sure if that approach will work for the EDPM doc too though, i don't know if the variables are defaulted in EDPM Ansible.

jistr (commented):
LGTM, but we should squash it into one commit. Alternatively we can squash when merging.

jistr:
(this will likely need a rebase anyway to pass CI after #166 merges)

jistr:
This doesn't touch tests and i can see the defaults being present, e.g. https://github.com/openstack-k8s-operators/ovn-operator/blob/7dd41df22fa2e910ee8e554ec094af94b5434f31/api/v1beta1/ovndbcluster_types.go#L39 .

So i'll merge. Thanks!

PR #170, opened 2023-09-27
WIP: Initial addition of the VANS dataplane adoption instructions

beagles:
This is an initial description of how dataplne adoption would work. This is a work-in-progress that is pending completion of OpenShift cluster side design of pod networking for the designate and octavia services.


docs/openstack/octavia-adoption.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Get the IPs for the load balancer management network
+
+*TODO: We will need the details of how the OpenShift OpenStack control
+plane makes the  IPs available. e.g. oc describe octavia
+octaviaHealthManagerS  -o json | jq -r '.status.healthManager.managementEndpoints'*
+
+### Modify the running load balancer's management network information.
+
+ *TODO: Allegedly there's an API for that *
```
weinimo, line 122: Yes, it seems to be [PUT /v2/octavia/amphorae/{amphora_id}/config](https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=configure-amphora-detail#configure-amphora). I haven't checked how the request needs to look like yet.

weinimo: Actually the example there shows that the request body is empty, so the PUT request just tell Octavia to update the config and the update is done automatically.

weinimo: `openstack loadbalancer amphora configure <ampid>`

docs/openstack/octavia-adoption.md
```diff
@@ -0,0 +1,122 @@
...
+oc patch openstackcontrolplane openstack --type=merge --patch '
+spec:
+  redis:
+    enabled: true
+	templates:
+    - something-octavia-specific-tbd
+'
+```
+
+## Shutdown Director Deployed Octavia Services
```
weinimo, line 37: I forgot you added this here already and added it in #174. A few questions:

1. Shouldn't we stop Octavia services before MariaDB data copy already in order prevent inconsistencies?
2. Your list of services below does not contain the driver agent server and the rsyslog services. I assume you did that on purpose, but I wanted to highlight that in case it was missed somehow.

docs/openstack/octavia-adoption.md
weinimo, line 1: For consistency with other files we should only use underscores in file names.

weinimo (commented):
Some comments/questions.

jistr (commented):
Overall LGTM. The Designate procedure is inherently complex it seems :). And we should figure out if we can do some basic testing in the test suite, but it doesn't necessarily need to be within this PR, especially since the procedures for theses services aren't trivial.

PR #174, opened 2023-10-04
Updates for Octavia adoption

weinimo:
- Add Octavia services to the services to stop

I was able to confirm that octavia databases get copied as part of the MariaDB data copy step already.


weinimo:
Superseded by #170

PR #180, opened 2023-10-12
Properly patch AEE image

fao89:


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -4,14 +4,17 @@
...
+      --type='json' -p='[{
+      "op":"replace", "path":"/spec/install/spec/deployments/0/spec/template/spec/containers/1/env/0",
+      "value": {"name": "RELATED_IMAGE_ANSIBLEEE_IMAGE_URL_DEFAULT", "value": "quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"}}]'
 
 - name: ensure oc login
   no_log: "{{ use_no_log }}"
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    {{ oc_login_command }}
```
marios, line 14: this one is intentional? why though

marios (changes requested):
sanity check please otherwise lgtm

tests/roles/dataplane_adoption/tasks/main.yaml
fao89: because we are doing the login in the previous task, no need to login 2 times in a row

jistr: I think we should be able to rely on the login which happens during the `prelude` role execution [1], just like the control plane adoption parts do? So we can probably remove both of the oc_login_command appearences from this file. Not a blocker of this patch though IMO.

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/8ceca3837fae2374b3ab024b7663eac94868692d/tests/roles/prelude_local/tasks/main.yaml#L12

jistr (approved):

tests/roles/dataplane_adoption/tasks/main.yaml
fao89: I don't know if one login is enough, as we have more and more services adopted, the login can expire in between

PR #183, opened 2023-10-17
Document ansible 8.5 and jmespath on py11

bogdando:
When running adoption on the 'laptop' with Centos 9 Stream, ansible 8.5 and ansible-core 2.15.x require python 3.11, while the system python is 3.9. For jmespath module to function in ansbile lookup queries, provide a workaround to install pip3.11 into the user space and use it with ansible automation for running test suits.


docs/contributing/tests.md
```diff
@@ -50,19 +50,29 @@ samples are provided. To run the tests, follow this procedure:
...
-  source venv/bin/activate
-  pip install openstackclient osc_placement jmespath
+  pip install --user openstackclient osc_placement jmespath
   ansible-galaxy collection install community.general
   ```
 
+* (on Centos9 Stream) Apply `jmespath` workarounds for ansible-core 2.15 and ansible 8.5:
+  ```
+  pip install --user ansible yq
+  sudo su
```
jpodivin, line 60: Is sudo really necessary here?

bogdando: yes, it didn't work for non root user shell

bogdando:
PTAL

docs/contributing/tests.md
```diff
@@ -50,19 +50,29 @@ samples are provided. To run the tests, follow this procedure:
 
   ```
   sudo dnf -y install python-devel
-  python3 -m venv venv
-  source venv/bin/activate
-  pip install openstackclient osc_placement jmespath
+  pip install --user openstackclient osc_placement jmespath
```
jistr, line 53: Hmm at least personally i'd probably not want to use this over the virtualenv... Maybe we should provide more options here? E.g. if someone is using Fedora + venv vs. CentOS + non-venv? At least for Adoption we've been using Fedora (38) for the lab machines.

jistr: Is there a reason to avoid virtualenv?

PR #190, opened 2023-11-06
Remove Heat hack for keystone endpoint

bshephar:
Heat was using the Public Keystone endpoint instead of internal. This caused issues with the self-signed certificate validation. This is fixed by: https://github.com/openstack-k8s-operators/heat-operator/pull/270

Depends-On: https://github.com/openstack-k8s-operators/heat-operator/pull/270

### DIFF
--- a/tests/roles/heat_adoption/tasks/main.yaml
+++ b/tests/roles/heat_adoption/tasks/main.yaml
@@ -15,11 +15,6 @@
         apiOverride:
           route: {}
         template:
-          customServiceConfig: |
-            [DEFAULT]
-            debug = true
-            [keystone_authtoken]
-            insecure = true
           databaseInstance: openstack
           databaseUser: heat
           rabbitMqClusterName: rabbitmq


bshephar:
The Heat Operator PR mentioned in the commit message needs to merge first. Leaving this as a draft until that happens.

bshephar:
recheck

bshephar:
Heat operator patch is merged now. Let's re-check and see if we can get that job passing without the hack,

bshephar:
recheck

bshephar:
Ah, we probably need to update the heat-operator in openstack-operator for this to work actually

bshephar:
recheck

bshephar:
Looks like the version of heat-operator still isn't updated in openstack-operator. So this will continue to fail until that bump is merged:

https://review.rdoproject.org/zuul/build/6b749009641240f0bf6ae63d8e2eade8/log/controller/pod/heat-api-5d5c5848d8-ckqxq-logs.txt#233

fao89 (approved):

cescgina (approved):

PR #196, opened 2023-11-10
Remove variables with external prefix

cescgina:
In order to make them clearer, variables that point to the source
(standalone) environment begin by "source" and not "external". After
removing the old "external" vars from CI, removing the remainig ones as
workaround here.

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/50752


cescgina:
This is working, but need to merge the depends-on rdo-jobs patch first

jistr (approved):

fao89 (approved):

PR #200, opened 2023-11-17
Add overrides for dataplane adoption os-net-config template

marios:
Needed by [1] this allows us to override the nic and routes for the os-net-config template passed with the dataplane deployment. Also see [2]

[1] https://review.rdoproject.org/r/c/rdo-jobs/+/50590/
[2] https://issues.redhat.com/browse/OSP-30319

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -205,6 +205,9 @@
           ansiblePort: 22
           ansibleUser: root
           ansibleVars:
+            os_net_config_iface: {{ dataplane_os_net_config_iface | default ('nic1') }}
+            os_net_config_set_route: {{ dataplane_os_net_config_set_route | default(true) | bool }}
+            os_net_config_dns: {{ dataplane_os_net_config_dns | default("") }}
             service_net_map:
               nova_api_network: internal_api
               nova_libvirt_network: internal_api
@@ -220,19 +223,20 @@
               {{ mtu_list.append(lookup('vars', networks_lower[network] ~ '_mtu')) }}
               {%- endfor %}
               {% set min_viable_mtu = mtu_list | max %}
+
               network_config:
               - type: ovs_bridge
                 name: {{ neutron_physical_bridge_name }}
                 mtu: {{ min_viable_mtu }}
                 use_dhcp: false
-                dns_servers: {{ ctlplane_dns_nameservers }}
+                dns_servers: {{ os_net_config_dns | default(ctlplane_dns_nameservers, true) }}
                 domain: {{ dns_search_domains }}
                 addresses:
                 - ip_netmask: {{ ctlplane_ip }}/{{ ctlplane_subnet_cidr }}
-                routes: {{ ctlplane_host_routes }}
+                routes: {{ ctlplane_host_routes if os_net_config_set_route else '[]' }}
                 members:
                 - type: interface
-                  name: nic1
+                  name: {{ os_net_config_iface }}
                   mtu: {{ min_viable_mtu }}
                   # force the MAC address of the bridge to this interface
                   primary: true
@@ -252,7 +256,8 @@
             # These vars are for the network config templates themselves and are
             # considered EDPM network defaults.
             neutron_physical_bridge_name: br-ctlplane
-            neutron_public_interface_name: eth0
+            neutron_public_interface_name: {{ dataplane_public_iface | default('eth0') }}
+
             role_networks:
             - InternalApi
             - Storage
@@ -280,7 +285,7 @@
             enable_debug: false
             # edpm firewall, change the allowed CIDR if needed
             edpm_sshd_configure_firewall: true
-            edpm_sshd_allowed_ranges: ['192.168.122.0/24']
+            edpm_sshd_allowed_ranges: {{ ['192.168.122.0/24'] if dataplane_os_net_config_set_route|default(true)|bool else ['0.0.0.0/0'] }}
             # SELinux module
             edpm_selinux_mode: enforcing
             plan: overcloud


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -261,7 +261,7 @@
             # These vars are for the network config templates themselves and are
             # considered EDPM network defaults.
             neutron_physical_bridge_name: br-ctlplane
-            neutron_public_interface_name: eth0
+            neutron_public_interface_name: {{ dataplane_os_net_config_iface | default('eth0') }}
```
marios, line 264: and make this explicitly eth1

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -229,6 +229,11 @@
               {{ mtu_list.append(lookup('vars', networks_lower[network] ~ '_mtu')) }}
               {%- endfor %}
               {% set min_viable_mtu = mtu_list | max %}
+              {% set os_net_config_routes = dataplane_os_net_config_routes | default(ctlplane_host_routes) %}
+              {% set os_net_config_iface = dataplane_os_net_config_iface | default('nic1') %}
+              {% set neutron_public_iface = dataplane_public_iface | default('eth1') %}
+
```
fao89, line 235: this template consumes the ansible vars from the dataplane inventory generated by the dataplane-operator, so for having these variables injected there, you need to add them here under `ansibleVars`

fao89:
you need something like this:
https://github.com/openstack-k8s-operators/data-plane-adoption/compare/main...fao89:data-plane-adoption:vars_demo?expand=1

marios:
> you need something like this: https://github.com/openstack-k8s-operators/data-plane-adoption/compare/main...fao89:data-plane-adoption:vars_demo?expand=1

thank you. yesterday for sanity i hard-coded them to see if the config was correct and if there was something further (seems ok, network config passed) so will add these vars and try with that today

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -207,6 +207,9 @@
           ansiblePort: 22
           ansibleUser: root
           ansibleVars:
+            os_net_config_iface: {{ dataplane_os_net_config_iface | default ('nic1') }}
+            os_net_config_routes: {{ dataplane_os_net_config_routes | default(ctlplane_host_routes) }}
```
fao89, line 211: `ctlplane_host_routes` is created on dataplane-operator reconcile loop

fao89: https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/pkg/deployment/inventory.go#L114

marios: so you mean, i should not try to default that here it doesn't exist yet. perhaps |default(omit) will work instead here then.

and I can have |default ctlplane_host_routes below in the template, as before WDYT?

fao89: exactly!

marios: and same for dnsnameservers too then i think https://github.com/openstack-k8s-operators/dataplane-operator/blob/0874b6a673e9fcc6ac7e96896bf3a7047b05fe45/pkg/deployment/inventory.go#L113

marios: thanks for reviewing @fao89 ++

fao89:
recheck

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -205,6 +205,9 @@
           ansiblePort: 22
           ansibleUser: root
           ansibleVars:
+            os_net_config_iface: {{ dataplane_os_net_config_iface | default ('nic1') }}
+            os_net_config_routes: {{ dataplane_os_net_config_routes | default(omit) }}
+            os_net_config_dns: {{ dataplane_os_net_config_dns | default(omit) }}
```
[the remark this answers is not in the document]
marios: ok. so we should be able to check if this does the right thing or not, on the run on this pull request, where we aren't passing any value for dataplane_os_net_config_routes or _dns. we could run a recheck here to see, but I'm about to rebase this and what you suggest sounds safe either way so will update with that anyway. we'll still see if that works well here before we merge

marios:
recheck 


        * https://logserver.rdoproject.org/00/200/0b503031a9d416047c167f26a34a17f1621bdd81/github-check/data-plane-adoption-github-rdo-centos-9-extracted-crc/cda58e5/job-output.txt
        * 2023-11-28 18:45:43.376937 | TASK [Accept standalone vm ssh host keys to avoid prompt connecting for the first time]
2023-11-28 18:45:48.921564 | controller | ERROR

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -252,7 +256,8 @@
             # These vars are for the network config templates themselves and are
             # considered EDPM network defaults.
             neutron_physical_bridge_name: br-ctlplane
-            neutron_public_interface_name: eth0
+            neutron_public_interface_name: {{ dataplane_public_iface | default('eth0') }}
```
marios, line 259: might still be some issue with these values.

the tests are passing but we are losing connectivity to the standalone over the main interface (eth0). i can ssh into it from the other nodes e.g. crc node using the 192.168 address.

marios: the problem i think is that I want to pass routes as []

but this is being treated as 'empty' and defaulting to the ctlplane_host_routes

fao89 (approved):

tests/roles/dataplane_adoption/tasks/main.yaml
marios: yes using the boolean flag for setting the route seems to work in last test

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -252,7 +256,8 @@
             # These vars are for the network config templates themselves and are
             # considered EDPM network defaults.
             neutron_physical_bridge_name: br-ctlplane
-            neutron_public_interface_name: eth0
```
marios, line 255: also need to add edpm_sshd_allowed_ranges: ['0.0.0.0/0']

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -280,7 +285,7 @@
             enable_debug: false
             # edpm firewall, change the allowed CIDR if needed
             edpm_sshd_configure_firewall: true
-            edpm_sshd_allowed_ranges: ['192.168.122.0/24']
+            edpm_sshd_allowed_ranges: "{{ ['192.168.122.0/24'] if dataplane_os_net_config_set_route|default(true)|bool else ['0.0.0.0/0'] }}"
```
marios, line 288: i think the quotes here are a mistake, and i will also remove the /0

`edpm_sshd_allowed_ranges: \"['0.0.0.0/0']\"\n        # SELinux module\n        edpm_selinux_mode: enforcing\n        plan: overcloud\nEOF\n", "delta": "0:00:00.761304", "end": "2023-11-30 08:13:50.213396", "msg": "", "rc": 0, "start": "2023-11-30 08:13:49.452092", "stderr": "+ oc apply -f -", "stderr_lines": ["+ oc apply -f -"], "stdout": "openstackdataplanenodeset.dataplane.openstack.org/openstack created", "stdout_lines": ["openstackdataplanenodeset.dataplane.openstack.org/openstack`

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -205,6 +205,9 @@
           ansiblePort: 22
           ansibleUser: root
           ansibleVars:
+            os_net_config_iface: {{ dataplane_os_net_config_iface | default ('nic1') }}
+            os_net_config_set_route: {{ dataplane_os_net_config_set_route | default(true) | bool }}
+            os_net_config_dns: {{ dataplane_os_net_config_dns | default("") }}
```
marios, line 210: sanity check for earlier conversation with @jistr about these values.

in the CI run here where we are not passing any of those 'dataplane_os_net_config_ ' vars the defaults are working as before:

https://logserver.rdoproject.org/00/200/c74c028d20b5831fa4bd0f92b90bf1721f900265/github-check/data-plane-adoption-github-rdo-centos-9-extracted-crc/54688aa/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/test_with_ceph_out_2023-11-30T10:23:56EST.log

```
TASK [dataplane_adoption : deploy dataplane] ***********************************
...

ansibleVars:\n        os_net_config_iface: nic1\n        os_net_config_set_route: True\n        os_net_config_dns: \n        service_net_map

```

and tests complete successfully

PR #202, opened 2023-11-20
Getting AEE unstuck

fao89:
ref: https://github.com/ansible/ansible-runner/issues/1327
Trying to unstuck EDPM playbooks
ref: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/191#issuecomment-1816810256

PR #204, opened 2023-11-21
[DNM] Disable sftp on file sending for ansible execution

rebtoor:
this is a test, just like https://github.com/openstack-k8s-operators/data-plane-adoption/pull/202


fao89:
recheck

fao89:
recheck

rebtoor:
not needed anymore, the problem was elsewhere

PR #206, opened 2023-11-22
[edpm]Do not run discover_hosts after adoption

gibizer:
The compute nodes being adopted has already been discovered in the 17.1 side so and that information is adopted via the DB adoption. So we should not run discovery again as it might mask an error in adoption.


fao89 (approved):

bogdando (approved):

jistr (approved):

fao89:
closing as it got reverted https://github.com/openstack-k8s-operators/data-plane-adoption/pull/207

PR #208, opened 2023-11-24
Drop container image ansible vars from samples

fao89:
These image vars are not needed in the samples. There are defaults already set in the corresponding roles in edpm-ansible. ref:  https://github.com/openstack-k8s-operators/dataplane-operator/pull/522

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -269,13 +269,6 @@
 
             edpm_chrony_ntp_servers: {{ edpm_chrony_ntp_servers | default(default_edpm_chrony_ntp_servers) }}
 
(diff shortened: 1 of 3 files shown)


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -269,13 +269,6 @@
 
             edpm_chrony_ntp_servers: {{ edpm_chrony_ntp_servers | default(default_edpm_chrony_ntp_servers) }}
 
-            edpm_ovn_controller_agent_image: "{{ registry_name }}/{{ registry_namespace }}/openstack-ovn-controller:{{ image_tag }}"
```
marios, line 272: we need to consider how we override this for the periodic job, i.e. when we need to set 'podified-ci-testing' instead of current-podified and also downstream job where we have a different registry altogether.

if we use defaults from edpm-ansible we  have to make sure we can override them i.e. lets work that out before doing this

marios (changes requested):

PR #192, opened 2023-11-06
Nova adoption ffu (no extra cell)

bogdando:
Extracted nova services FFU step being done from https://github.com/openstack-k8s-operators/data-plane-adoption/pull/176

Update EDPM adoption docs and tests to execute Nova compute post-FFU.

Verify no dataplane disruptions during the FFU/adoption process.

Verify Nova services still control pre-created VM workload after
FFU/adotpion is done.

Jira: [OSPRH-338](https://issues.redhat.com/browse/OSPRH-338)

TODO (rebaes after merged deps):
- [ ] Depends-on: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/191

Issue #109, opened 2023-08-15
OVN northd starts together with DBs before DB data is imported


jistr:
Northd is meant to start at the end of the procedure after both NB and SB databases are imported, but currently it actually starts alongside the DB pods.

PR #210, opened 2023-11-24
Fix OVN template for isolnet scenarios

bogdando:
Align CR for OVN adoption with its isolnet CR sample in openstack operator.

Fix communication between OSP control plane pods on OCP and standalone OSP 17.1 host.

### DIFF
--- a/docs/openstack/backend_services_deployment.md
+++ b/docs/openstack/backend_services_deployment.md
@@ -197,11 +197,20 @@ podified OpenStack control plane services.
     ovn:
       enabled: false
       template:
+        ovnDBCluster:
+          ovndbcluster-nb:
+            dbType: NB
+            storageRequest: 10G
+            networkAttachment: internalapi
+          ovndbcluster-sb:
+            dbType: SB
+            storageRequest: 10G
+            networkAttachment: internalapi
+        ovnNorthd:
+          networkAttachment: internalapi
+          replicas: 1
         ovnController:
-          external-ids:
-            system-id: "random"
-            ovn-bridge: "br-int"
-            ovn-encap-type: "geneve"
+          networkAttachment: tenant
 
     placement:
       enabled: false
--- a/tests/config/base/openstack_control_plane.yaml
+++ b/tests/config/base/openstack_control_plane.yaml
@@ -83,14 +83,18 @@ spec:
     enabled: false
     template:
       ovnController:
-        external-ids:
-          system-id: "random"
-          ovn-bridge: "br-int"
-          ovn-encap-type: "geneve"
+        networkAttachment: tenant
+      ovnNorthd:
+        networkAttachment: internalapi
+        replicas: 1
       ovnDBCluster:
         ovndbcluster-nb:
+          dbType: NB
+          networkAttachment: internalapi
           containerImage: quay.io/podified-antelope-centos9/openstack-ovn-nb-db-server:current-podified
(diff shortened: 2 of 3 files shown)


docs/openstack/backend_services_deployment.md
```diff
@@ -197,11 +197,23 @@ podified OpenStack control plane services.
...
+              storageRequest: 10G
+              networkAttachment: internalapi
+          ovnNorthd:
+            networkAttachment: internalapi
+          ovnController:
+            external-ids:
+              system-id: "random"
+              ovn-bridge: "br-int"
+              ovn-encap-type: "geneve"
+            networkAttachment: tenant
```
bogdando, line 216: see for the context https://github.com/openstack-k8s-operators/data-plane-adoption/pull/192#issuecomment-1825336591

docs/openstack/backend_services_deployment.md
```diff
@@ -197,11 +197,23 @@ podified OpenStack control plane services.
...
+              networkAttachment: internalapi
+            ovndbcluster-sb:
+              dbType: SB
+              storageRequest: 10G
+              networkAttachment: internalapi
+          ovnNorthd:
+            networkAttachment: internalapi
+          ovnController:
+            external-ids:
+              system-id: "random"
```
karelyatin, line 213: external ids are defalt so not needed to set, but shouldn't harm

karelyatin (commented):
/lgtm

fao89 (commented):
ansible needs to be updated
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/config/base/openstack_control_plane.yaml


https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/config/periodic_ci/container_image_overrides.yaml

fao89 (approved):

bogdando:
ok shall we merge this?
It still doesn't help to move https://github.com/openstack-k8s-operators/data-plane-adoption/pull/192 forward, but I think the direction this goes is correct?

docs/openstack/backend_services_deployment.md
```diff
@@ -197,11 +197,20 @@ podified OpenStack control plane services.
...
+            dbType: NB
+            storageRequest: 10G
+            networkAttachment: internalapi
+          ovndbcluster-sb:
+            dbType: SB
+            storageRequest: 10G
+            networkAttachment: internalapi
+        ovnNorthd:
+          networkAttachment: internalapi
+          replicas: 1
```
jistr, line 211: OVN northd is supposed to start after DB imports are done, but it's already not the case ( https://github.com/openstack-k8s-operators/data-plane-adoption/issues/109 ) so not a blocker. Eventually we should probably have replicas 0 here and replicas 1 in the OVN adoption.

However, let's please make the same updates to docs as we're doing to the test suite.

jistr: Since the CI is green here the doc update can be a different PR but it would be nice to have it posted before we merge?

bogdando: I'm not sure what exactly should be in that follow-up, sorry

jistr: Sorry i see you're editing both, not sure how i missed that earlier, i guess e_too_many_things :)

jistr (approved):

PR #211, opened 2023-11-26
add ospdo test_minimal

pinikomarov:
Adds a ospdo env on the controller nodes to be the source of adoption.
Runs tests agains it


tests/playbooks/test_minimal_ospdo.yaml
```diff
@@ -0,0 +1,43 @@
...
+- name: Cleanup
+  hosts: local
+  gather_facts: false
+  module_defaults:
+    ansible.builtin.shell:
+      executable: /bin/bash
+  roles:
+    - pcp_cleanup
+
+# TODO need to execute the first 4 roles against the "controller" host
```
pinikomarov, line 19: @marios @cescgina @jistr Hi guys can you take a look at the options I have of ansible execution here ? 
The two hosts are localhost (crc adoption dest) and controller (ospdo adoption source), also specified in this patch as "{{ adoption_ospdo_host }}".
some of the commands require execution on the loccalhost (crc) and some execute on the controller , get some mdata and use it on the crc adopting host.
usage can be seen here  : 
https://github.com/rdo-infra/rdo-jobs/blob/master/playbooks/data_plane_adoption/deploy_standalone_run_repo_tests.yaml#L247-L248

wdyt ?

pinikomarov:
continued here : https://github.com/openstack-k8s-operators/data-plane-adoption/pull/292

PR #186, opened 2023-10-27
Use service selector for openstackclient pod

cescgina:
Recently the openstackclient pod moved to using app to service as
a selector, this change updates the check in the tests to use the right
label.

It appears this change in openstack-operator was the trigger:
https://github.com/openstack-k8s-operators/openstack-operator/pull/502

PR #212, opened 2023-11-27
Fix waiting for openstackclient pod

bogdando:
Use selector app=openstackclient instead of
service=openstackclient

### DIFF
--- a/tests/roles/keystone_adoption/tasks/main.yaml
+++ b/tests/roles/keystone_adoption/tasks/main.yaml
@@ -37,7 +37,7 @@
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc get pod --selector=service=openstackclient -o jsonpath='{.items[0].status.phase}{"\n"}' | grep Running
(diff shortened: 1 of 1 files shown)


cescgina (commented):
I'm not sure which is the right way of waiting for the openstackclient, but this change does not work in CI, I just tested it in a held node

bogdando:
> I'm not sure which is the right way of waiting for the openstackclient, but this change does not work in CI, I just tested it in a held node

ugh... that's interesting. On my dev env, it helped to fix a wrong waiting condition...

cescgina:
probably, I changed it here https://github.com/openstack-k8s-operators/data-plane-adoption/pull/186. This seems like the change that broke it last time https://github.com/openstack-k8s-operators/openstack-operator/pull/502/files#diff-f6e8b90519d4c43e1d61937b0cfd3ef6f477adbf0ea82470281629ff56f6c4acR178. This https://github.com/openstack-k8s-operators/lib-common/blob/6bd6025ade373131acb7737c58cb61b43dafeee4/modules/common/const.go#L22 seems also relevant

PR #216, opened 2023-11-29
Fernet keys adoption

xek:
Add copying of fernet keys to Keystone adoption procedure

### DIFF
--- a/docs_user/modules/openstack-keystone_adoption.adoc
+++ b/docs_user/modules/openstack-keystone_adoption.adoc
@@ -12,6 +12,28 @@ must already be imported into the podified MariaDB.
 
 == Pre-checks
 
+== Copy over fernet keys
+
+* Create the `keystone` secret, containing fernet keys:
++
+[,bash]
+----
+oc apply -f - <<EOF
+apiVersion: v1
+data:
+  CredentialKeys0: $($CONTROLLER1_SSH sudo cat /var/lib/config-data/puppet-generated/keystone/etc/keystone/credential-keys/0 | base64 -w 0)
+  CredentialKeys1: $($CONTROLLER1_SSH sudo cat /var/lib/config-data/puppet-generated/keystone/etc/keystone/credential-keys/1 | base64 -w 0)
+  FernetKeys0: $($CONTROLLER1_SSH sudo cat /var/lib/config-data/puppet-generated/keystone/etc/keystone/fernet-keys/0 | base64 -w 0)
+  FernetKeys1: $($CONTROLLER1_SSH sudo cat /var/lib/config-data/puppet-generated/keystone/etc/keystone/fernet-keys/1 | base64 -w 0)
+kind: Secret
+metadata:
+  name: keystone
+  namespace: openstack
+type: Opaque
+EOF
+----
+
+
 == Procedure - Keystone adoption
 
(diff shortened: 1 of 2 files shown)


docs/openstack/keystone_adoption.md
```diff
@@ -10,6 +10,29 @@
...
+Create the `keystone` secret, containing fernet keys:
+
+```
+oc apply -f - <<EOF
+apiVersion: v1
+data:
+  CredentialKeys0: $(cat /tmp/collect_tripleo_configs/keystone/etc/keystone/credential-keys/0 | base64 -w 0)
+  CredentialKeys1: $(cat /tmp/collect_tripleo_configs/keystone/etc/keystone/credential-keys/1 | base64 -w 0)
+  FernetKeys0: $(cat /tmp/collect_tripleo_configs/keystone/etc/keystone/fernet-keys/0 | base64 -w 0)
+  FernetKeys1: $(cat /tmp/collect_tripleo_configs/keystone/etc/keystone/fernet-keys/1 | base64 -w 0)
```
jistr, line 26: I wonder if we should utilize `$CONTROLLER_SSH` explicitly instead of reusing what OS Diff pulled into /tmp... That would make it easier to add this into the test suite as it would be decoupled from OS Diff which is not currently there.

jistr:
Still a draft?

docs/openstack/keystone_adoption.md
jistr: We discussed yesterday that initially OS Diff will be used after deploying the podified control plane to validate its configuration, but it will not be used before deploying the control plane (to not make the documentation confusing to users). This means we shouldn't rely on OS Diff prior/during deployment of the control plane, we should fetch these keys with $CONTROLLER1_SSH.

https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs/openstack/stop_openstack_services.md#variables

jistr (changes requested):

docs_user/modules/openstack-keystone_adoption.adoc
```diff
@@ -12,6 +12,29 @@ must already be imported into the podified MariaDB.
 
 == Pre-checks
 
+== Copy over fernet keys
+
+Pull your [openstack configuration using os-diff](pull_openstack_configuration.md) from the old environment.
+
+Create the `keystone` secret, containing fernet keys:
+
+```
```
jistr, line 21: AsciiDoc syntax -- see line 42 below for a code block.

docs_user/modules/openstack-keystone_adoption.adoc
```diff
@@ -12,6 +12,29 @@ must already be imported into the podified MariaDB.
 
 == Pre-checks
 
+== Copy over fernet keys
+
+Pull your [openstack configuration using os-diff](pull_openstack_configuration.md) from the old environment.
```
jistr, line 17: AsciiDoc syntax - see line 6 above for local link syntax. But now that we pull the config directly, we can probably remove this line altogether?

docs/openstack/keystone_adoption.md
xek: I changed it to use $CONTROLLER_SSH

jistr (approved):

fao89 (approved):

PR #217, opened 2023-11-30
Asciidoc rendering (local and in gh-pages)

jistr:
This adds stubs for the new AsciiDoctor docs (one document for user-facing content, one document for developer-facing content). It also edits the Github Actions workflow for docs rendering so that the legacy docs are rendered still the same (into the top-level directory of the gh-pages branch) and the new docs are rendered into `user` and `dev` subdirectories.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -15,10 +15,52 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: '3.9'
+      - uses: ruby/setup-ruby@v1.160.0
+        with:
+          ruby-version: '3.2'
+
       - name: Install MkDocs
         run: pip install -r docs/doc_requirements.txt
-      - name: Publish docs
+      - name: Build legacy mkdocs documentation
+        run: make mkdocs
+
+      - name: Install Asciidoc
+        run: make docs-dependencies
+      - name: Build user docs
+        run: make docs-user
+      - name: Build dev docs
+        run: make docs-dev
+
+      - name: Prepare gh-pages branch
         run: |
           git config user.name github-actions
           git config user.email github-actions@github.com
-          mkdocs gh-deploy --force --clean --verbose
+
+          git branch -D gh-pages &>/dev/null || true
+          git checkout -b gh-pages aabb4803f89a3b006094d8ae7a933650f644d036
+
+      - name: Commit asciidoc docs
+        run: |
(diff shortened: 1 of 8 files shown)


docs_user/main.adoc
```diff
@@ -0,0 +1,3 @@
+= Data Plane Adoption
```
jistr, line 1: I wonder if we'll want the OpenStack + Ceph procedures in one AsciiDoc document, or whether we'll want to separate them...

But anyway maybe this is something that should be relatively easy to tweak later if needed.

.github/workflows/docs.yaml
```diff
@@ -15,10 +15,51 @@ jobs:
...
+      - name: Build dev docs
+        run: make docs-dev
+
+      - name: Prepare gh-pages branch
         run: |
           git config user.name github-actions
           git config user.email github-actions@github.com
-          mkdocs gh-deploy --force --clean --verbose
+
+          git checkout -b gh-pages aabb4803f89a3b006094d8ae7a933650f644d036
```
jistr, line 39: I should probably put here a gh-pages branch delete here to be sure. I guess only 'main' branch gets created locally and the 'gh-pages' is only remote so that's why it wasn't complaining in my testing and would probably work, but i'll make it explicit.

docs_user/main.adoc
jistr: cc @klgill @fmount

fao89 (commented):
looking good so far

docs_user/main.adoc
klgill: I want to clarify that I agree with keeping the Ceph procedure within the overall adoption guide, but as its own chapter.  Maybe I misunderstood Jiri. 
That's similar to the current FFU guide. The customer follows the FFU guide in sequence, and follows the Ceph chapter if that applies to their environment.

marios:
merging as discussed just now on ci call. there are no updates to tests here (besides Makefile change) and the new OSP17 job is green anyway.

jistr:
Thanks, looks like it worked as intended, original docs look the same [1], stubs for new docs are in place [2][3].

[1] https://openstack-k8s-operators.github.io/data-plane-adoption/
[2] https://openstack-k8s-operators.github.io/data-plane-adoption/user/
[3] https://openstack-k8s-operators.github.io/data-plane-adoption/dev/

PR #218, opened 2023-12-05
Modify the upstream dns server in the control plane

cescgina:
Patch the control plane dns upstream server to point to the
dns server address in the running environment.

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/50590


jistr (approved):

fao89 (approved):

marios (approved):
lgtm

PR #213, opened 2023-11-28
Parameterize EDPM

fao89:

PR #219, opened 2023-12-07
Removing dataplane_rhev_adoption role

fao89:
dataplane_adoption should support RHEV: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/213
`make test-minimal` should work on RHEV


fao89:
recheck

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -75,7 +75,7 @@
           ssh \
             -i {{ edpm_privatekey_path | default("/tmp/ansible_private_key") }} \
             {{ edpm_user }}@"${computes[$name]}" \
-            "echo $uuid > /var/lib/nova/compute_id"
+            "sudo echo $uuid > /var/lib/nova/compute_id"
```
fao89, line 78: @archana203 I've added it here

fao89:
recheck

tests/roles/dataplane_adoption/tasks/main.yaml
archana203: I still see permission denied with this:
[stack@undercloud-0 ~]$ ssh -i ~/.ssh/id_rsa stack@192.168.23.10 "sudo echo test > /var/lib/nova/compute_id"
bash: line 1: /var/lib/nova/compute_id: Permission denied

fao89: @bogdando @gibizer @SeanMooney do you have any clue on how to make this work?
In our CI we use `root` so it works fine, but on RHEV we have the `stack` user and we are hitting this error

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -78,6 +78,7 @@
             "echo $uuid > /var/lib/nova/compute_id"
 
         done
+      when: not platform_rhev|default(false)
```
fao89, line 81: @archana203 could you check if this PR works on RHEV when we skip this step?

tests/roles/dataplane_adoption/tasks/main.yaml
bogdando: As @sean-k-mooney have recommended for a similar case, consider using `edpm_user` role to create an `openstack` user (or stack?), then make sure when a task uses ``become:true`` (instead of sudo) it sets the user and group to ansible user, like this https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nova/tasks/configure.yml#L3-L19

tests/roles/dataplane_adoption/tasks/main.yaml
archana203: @fao89 Yeah, it worked without this step.

tests/roles/dataplane_adoption/tasks/main.yaml
fao89: something like this? https://github.com/openstack-k8s-operators/edpm-ansible/pull/534/files

fao89:
recheck

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -75,7 +75,7 @@
           ssh \
             -i {{ edpm_privatekey_path | default("/tmp/ansible_private_key") }} \
             {{ edpm_user }}@"${computes[$name]}" \
-            "echo $uuid > /var/lib/nova/compute_id"
+            "echo $uuid | sudo tee /var/lib/nova/compute_id"
```
gibizer, line 78: if this keeps read access for the nova-compute service to this file then I'm fine with it.

archana203: compute_id is registered with this command. But the id does not match with the one registered in compute service list and computes state is down.
https://privatebin.corp.redhat.com/?2a67ecedadf5a3a2#3qmFnemnvWxERgykPRhiE3FQ3JjR96hiTRzftm49ohSh

fao89: @bogdando @gibizer any idea why the compute state is down?

gibizer: @archana203 could you point me to the job logs?

gibizer: Without logs or troubleshooting I can only guess.
A) the nova-compute service cannot read the compute_id file as it is now written via sudo
B) for some reason the openstack client is different in this case and instead of defaulting to the maximum nova api microversion it defaults to the minimum. If this is the case then you can add `--os-compute-api-version 2.53` to the openstack hypervisor command to force the newer microversion. (Before 2.53 the hypervisor show CLI returned the 'id' not the 'uuid' of the compute node object)

gibizer: I looked at the deployment provided by @archana203 . The OSP18 deployment failed at network config
```
TASK [osp.edpm.edpm_users : Ensure user is present on the host [ openvswitch ]] ***
Monday 22 January 2024  18:18:17 +0000 (0:00:00.630)       0:00:03.397 ********
fatal: [compute-0]: FAILED! => {"changed": false, "msg": "usermod: user openvswitch is currently used by process 15864\n", "name": "openvswitch", "rc": 8}
fatal: [compute-1]: FAILED! => {"changed": false, "msg": "usermod: user openvswitch is currently used by process 15950\n", "name": "openvswitch", "rc": 8}
```
So the new nova_compute container is not yet deployed.

The compute_id is generate properly. It has o+r rights so nova-compute will be able to read it.

bogdando: there was a fix https://github.com/openstack-k8s-operators/edpm-ansible/pull/540

fao89: does it mean sudo isn't needed anymore?

SeanMooney: no sudo is still required

but the issue gibi noted is fixed.

it was unrelated to this change

SeanMooney: /var/lib/nova/compute_id is chown'd by kolla when the nova_compute container starts so this will become owned by nova at that point so yes nova will have read access to it.

fao89:
recheck

jistr:
LGTM but i think we need an ack from @archana203 before merging.

archana203:
LGTM, The changes work on RHEV setup without an issue.

jistr (approved):

jistr:
One thing tho -- the PR has been here for a while and the branch is now 10 commits behind main. We should probably rebase and let the CI run again before merging?

fao89:
just rebased it!

tests/roles/dataplane_adoption/vars/rhev.yaml
```diff
@@ -37,21 +37,48 @@ netconfig_networks:
...
+        subnetName: subnet1
+      - name: InternalApi
+        fixedIP: 172.15.2.11
+        subnetName: subnet1
+      - name: Tenant
+        fixedIP: 172.15.0.11
+        subnetName: subnet1
+      - name: External
+        fixedIP: 10.4.0.11
+        subnetName: subnet1
```
fao89, line 78: 2nd node

PR #220, opened 2023-12-11
Barbican adoption procedure

xek:
- Fix command output examples
- Barbican adoption procedure

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/51920


jistr (commented):
Looks good :+1: . It would be good to add tests too but it doesn't have to be done in the initial PR.

jistr:
Still a draft? (Also needs a rebase now.)

docs/openstack/backend_services_deployment.md
```diff
@@ -127,9 +128,13 @@ podified OpenStack control plane services.
   account passwords from the original deployment:
 
   ```
+<<<<<<< HEAD
   oc set data secret/osp-secret "AodhPassword=$AODH_PASSWORD"
   oc set data secret/osp-secret "CeilometerMeteringSecret=$CEILOMETER_METERING_SECRET"
   oc set data secret/osp-secret "CeilometerPassword=$CEILOMETER_PASSWORD"
+=======
+  oc set data secret/osp-secret "BarbicanPassword=$BARBICAN_PASSWORD"
+>>>>>>> 5bd34c5 (Barbican adoption procedure)
```
jistr, line 137: Here's a rebase artifact, and the CI failure looks like Barbican failed to start, but other than that the patch LGTM.

docs/openstack/backend_services_deployment.md
```diff
@@ -54,6 +54,7 @@ podified OpenStack control plane services.
   AODH_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AodhPassword:' | awk -F ': ' '{ print $2; }')
   CEILOMETER_METERING_SECRET=$(cat ~/tripleo-standalone-passwords.yaml | grep ' CeilometerMeteringSecret:' | awk -F ': ' '{ print $2; }')
   CEILOMETER_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' CeilometerPassword:' | awk -F ': ' '{ print $2; }')
+  BARBICAN_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' BarbicanPassword:' | awk -F ': ' '{ print $2; }')
```
vakwetu, line 57: You are missing the barbican kek (assuming that this is for simple crypto).  This is of course assuming that the backend of the adopted system is simple_crypto.  We probably  need something about backends in the doc in the barbican section.

docs/openstack/backend_services_deployment.md
vakwetu: Need to set the KEK here too.  As of recent commits, this is possible as well.  This absolutely must be the same as what is in the existing barbican config or you will not be able to decrypt existing secrets.

docs/openstack/barbican_adoption.md
```diff
@@ -0,0 +1,83 @@
...
+is initialy disabled, should be patched to start the service with the configuration
+parameters provided by the source environment.
+
+When the procedure is over, the expectation is to see the `BarbicanAPI`, `BarbicanWorker`,
+`BarbicanKeystoneListener` services up and running.
+The `Keystone endpoints` should also be updated and the same crypto plugin of the source Cloud
+will be available. If the conditions above are met, the adoption is considered concluded.
+
+> **NOTE** Barbican doesn't yet support all of the crypto plugins available in TripleO.
+
```
vakwetu, line 13: We should specify here the plugins we do support -- ie. simple crypto

docs/README.md
```diff
@@ -21,6 +21,8 @@ Perform the actions from the sub-documents in the following order:
```
vakwetu, line 21: Just adding a comment here to make sure we capture this work in a Jira somewhere.  The pull_openstack_configuration.md file discusses the use of os-diff to capture config differences for different services.  We are going to need to update that tool to add code for barbican and keystone.  In particular, see https://github.com/openstack-k8s-operators/os-diff/tree/main#add-service

vakwetu: Resolve this comment once the relevant Jira are created.

vakwetu: In stop_openstack_services.md, there is a script that is used to stop openstack services.  Does it need an addition to stop barbican too?

docs/README.md
```diff
@@ -21,6 +21,8 @@ Perform the actions from the sub-documents in the following order:
 
 * [Keystone adoption](openstack/keystone_adoption.md)
```
vakwetu, line 22: In the keystone_adoption.md , there are commands to remove old endpoints and services.  We need to add barbican here too.

docs/openstack/barbican_adoption.md
```diff
@@ -0,0 +1,83 @@
...
+```bash
+openstack endpoint list | grep key-manager
+```
+
+### List secrets
+```bash
+openstack secret list
+```
+
```
vakwetu, line 82: We can do even better in terms of testing by creating a test secret prior to the database migration.  Then, at this stage, we can retrieve this test secret and its payload.  This tests the database migration - but also is a good test of the barbican plugin migration, because the plugin would need to be migrated correctly in order for the secret to be decrypted properly.  We can then delete the test secret.

vakwetu: Oh, I see we do that in the test code ..  Feels like a good thing to document here too, eh?

tests/roles/barbican_adoption/tasks/main.yaml
```diff
@@ -0,0 +1,54 @@
+- name: deploy podified Barbican
```
vakwetu, line 1: Adding here -- need to look at keystone_adoption.  There is a task there to remove old endpoints.  You need to remove the barbican endpoint too.

vakwetu: ditto for stop_openstack_services

tests/roles/barbican_adoption/tasks/main.yaml
```diff
@@ -0,0 +1,54 @@
...
+    {{ oc_header }}
+
+    alias openstack="oc exec -t openstackclient -- openstack"
+
+    ${BASH_ALIASES[openstack]} endpoint list | grep key-manager
+    ${BASH_ALIASES[openstack]} secret list
+  register: barbican_responding_result
+  until: barbican_responding_result is success
+  retries: 60
+  delay: 2
```
vakwetu, line 54: same question as above -- is there a way to create a secret before the migration - and then test for its existence here?  Also, we should probably do a test for creating a secret - that will test the operation of the worker.

It might be nice also to test the keystone-listener but that's going to involve a lot more setup.

docs/openstack/barbican_adoption.md
vakwetu: Can also do a test for creating a secret (generating a key) - that will test the barbican-workers

vakwetu (changes requested):
The kek is one thing that is likely causing an decrypts to fail.  See other changes as commented.

docs/README.md
xek: I added barbican to the list of stopped services

docs_user/modules/openstack-barbican_adoption.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+____
+*NOTE* This procedure configures barbican to use the simple_crypto backend.
+Additional backends are available, such as PKCS11 and DogTag, however they are
+not supported in this release.
+____
+
+== Prerequisites
+
+* Previous Adoption steps completed. Notably, MariaDB and Keystone and xref:ovn_adoption.adoc[OVN DBs]
```
vakwetu, line 20: xref:ovn_adoption.adoc[OVN DBs]  not sure what this is supposed to be, but we don't depend on OVN , do we?

xek:
We fixed an issue with setting the barbican user password in the tests - it should be good to go.

docs/openstack/barbican_adoption.md
xek: We will follow up with additional tests in a separate patch. I created a jira story here: https://issues.redhat.com/browse/OSPRH-5317

xek:
We need https://review.rdoproject.org/r/c/rdo-jobs/+/51920 in, to make the rdo job pass

cescgina:
recheck

xek:
recheck

xek:
@cescgina The data-plane-adoption-osp-17-to-extracted-crc job passed with the Depends-On

jistr (approved):

PR #226, opened 2023-12-19
Add Telemetry and Autoscaling adoption guide

yadneshk:


fao89:
recheck

jistr:
Needs a rebase but it's looking good :+1: 

It would be good to add testing as well but that may not be absolutely necessary for the initial PR.

jistr (approved):

jistr:
:+1: Thanks, please follow up with a test suite PR to reflect the new services. I'm merging as we want to move from Markdown to AsciiDoc, so any bigger chunks of docs that are ready to land should land.

PR #227, opened 2023-12-22
Use default CRC sc instead of local store

bogdando:
Remove dependencies on install_yaml make crc_storage* targets and switch storageClasses in the adoption related specs to use the default CRC sc. Note to chose the real storage classes in production scenarios with real OCP and OSP clusters.

Replace the make targets for storage to direct ocp commands and PVC targets for it.

Related [JIRA](https://issues.redhat.com/browse/OSPRH-861)


docs/openstack/node-selector.md
```diff
@@ -52,7 +52,7 @@ metadata:
   name: openstack
 spec:
   secret: osp-secret
-  storageClass: local-storage
+  storageClass: crc-csi-hostpath-provisioner
```
cescgina, line 55: shouldn't this use `$STORAGE_CLASS_NAME` as the other files?

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -32,6 +32,8 @@
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    cd {{ install_yamls_path }}
-    for i in {1..3}; do make crc_storage_cleanup crc_storage && break || sleep 5; done
+    oc delete pvc ovn-data
```
cescgina, line 35: these delete commands should include `--ignore-not-found=true` to avoid errors when running on a clean environment

docs/openstack/node-selector.md
bogdando: this document looks detached from the ohter places and there is no env variables for it, so I am not sure if it should, or this is good as is. @Akrog WDYT?

tests/roles/pcp_cleanup/tasks/main.yaml
bogdando: done

bogdando:
So, we should configure LVMS for CI, and probably as a default as well (instead of CRC storage).

fultonj (changes requested):
Ultimately we want the PVCs used by the control plane to use LVMS [1] and we want to be prescriptive about that (OSPRH-861) and not just say "use any storage class" which is part of the proposed documentation change here.

The other motivation for this patch, based on my chat with the author, is to have dependency on less tools. That's nice but it's a move to somewhere we will then have to move from again since our ultimate goal is to get to LVMS [1].

The storage targets provided by `make crc_storage` should be removed, I agree, but the plan is to replace them with a method to get LVMS working (OSPRH-861). I'll probably do that via both install_yamls and ci-framework unless everyone is off of install_yamls by then. LVMS will be local-storage.

[1] https://docs.openshift.com/container-platform/4.14/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html

bogdando:
the longer I think on this easy win, the more I prefer an option to ditch ``make crc_storate / cleanup`` dances earlier than later. Even before we start implementing LVMs support, that simple change could start benefittinc RHOSO devs on CRC just today

docs/openstack/glance_adoption.md
```diff
@@ -22,13 +22,21 @@ This guide also assumes that:
...
+  ```
+
 ## Procedure - Glance adoption
 
 As already done for [Keystone](https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/keystone_adoption.md), the Glance Adoption follows the same pattern.
 
 ### Using local storage backend
 
-When Glance should be deployed with local storage backend (not Ceph),
+When Glance should be deployed without Ceph,
```
fultonj, line 39: This change is wrong.

Even when glance is deployed with Ceph we still need local storage for import plugins.

  https://github.com/openstack-k8s-operators/glance-operator/tree/main/config/samples/import_plugins

So Glance should always be deployed with a PVC regardless of the storage backend.

docs/openstack/node-selector.md
fultonj: The CR should always have `storageClass: local-storage`.

fultonj (dismissed):

fultonj:
> the longer I think on this easy win, the more I prefer an option to ditch `make crc_storate / cleanup` dances earlier than later. Even before we start implementing LVMs support, that simple change could start benefitting RHOSO devs on CRC just today
> 
> @fultonj @jistr @gfidente @Akrog fyi

What exactly is the benefit?

This patch makes a number of these changes:
```
-  storageClass: local-storage
+  storageClass: crc-csi-hostpath-provisioner
```
But we want everyone to be testing `local-storage` since that's going to be our default. We cannot require fancy RWX or external storage for control plane pods.

`local-storage` comes from install_yamls today but will come LVMS in the future. Defaulting it to something that is not local-storage and is CRC-centric now just forces people to make a temporary move to a place we are not going to be anyway.

bogdando:
> What exactly is the benefit?

from my dev experience, it became much simpler to re-spin the adoption test case (without ceph), with CRC default storage, omitting all the local-storage manipulations. And Glance adoption used to work for me with that patch as is...

docs/openstack/node-selector.md
fultonj: Today, `install_yamls` provides a `local-storage` class which we can use for testing. It uses `make crc_storage` for now. 

In the future `ci-framework` will provide `local-storage` and it will be implemented via LVMS. I hope everyone will be on ci-framework by then, but I expect I'll update install_yamls too so that it has something like `make lvms_storage`.

fultonj:
> > What exactly is the benefit?
> 
> from my dev experience, it became much simpler to re-spin the adoption test case (without ceph), with CRC default storage, omitting all the local-storage manipulations. 

So the make file in install_yamls creates local-storage for now via `make crc_storage`. That command takes <1 minute to run on my system. I guess you're trying to save that extra step since crc comes with with `crc-csi-hostpath-provisioner`.

The CRs have `local-storage` in them. If you want the CRs to have `crc-csi-hostpath-provisioner` as the storage class you could have your own kustomize for that but people should be testing local-storage instead since that's what's going to be our default. I don't think we should change that default.

> And Glance adoption used to work for me with that patch as is...

Glance should be tested with import_plugins. When glance is deployed with Ceph the conversion import plugin will be enabled by default (this is a missing feature that is coming).

If you want to test glance adoption with local storage as the storage backend and with no import plugins enabled, that's fine. It still adds value since adoption is getting tested. Just know that we won't be running Glance that way in production (storing all images on a PVC).

bogdando:
> `make crc_storage`. That command takes <1 minute to run on my system. I guess you're trying to save that extra step since crc comes with with `crc-csi-hostpath-provisioner`.

Not exactly, the creation time was OK, while the cleanup step via removing *everything* is used to be a bit of overkill for me, and misc finalizers were often making the cleanup flaky

bogdando:
FTR we are already using this CRC SC for local PVC to store imported data from the source cloud during OVN adoption. This provided a 1st example of removing podman host-mounts from the adoption procedure. Other components' adoption still uses podman and host-mounts, and that was supposed to change earlier, than waiting for production ready LVMS support.

fultonj:
> > `make crc_storage`. That command takes <1 minute to run on my system. I guess you're trying to save that extra step since crc comes with with `crc-csi-hostpath-provisioner`.
> 
> Not exactly, the creation time was OK, while the cleanup step via removing _everything_ is used to be a bit of overkill for me, and misc finalizers were often making the cleanup flaky

I get it. Yes, the clean up of that is overkill.

fultonj:
Perhaps this PR can be updated to keep `local-storage` by default but to make it easy to use `crc-csi-hostpath-provisioner` by passing a parameter?

bogdando:
> Perhaps this PR can be updated to keep `local-storage` by default but to make it easy to use `crc-csi-hostpath-provisioner` by passing a parameter?

As we've discussed on slack, we could hopefully use a modification of the default CRC storage class, re-created with RWO instead of RWX

docs/openstack/backend_services_deployment.md
```diff
@@ -120,7 +126,7 @@ podified OpenStack control plane services.
     name: openstack
   spec:
     secret: osp-secret
-    storageClass: local-storage
+    storageClass: $STORAGE_CLASS_NAME
```
SeanMooney, line 129: you need to update the glance template on line 159 to use this too

 glance:
    template:
      storageClass: $STORAGE_CLASS_NAME

docs/openstack/glance_adoption.md
```diff
@@ -40,7 +48,7 @@ spec:
       route: {}
     template:
       databaseInstance: openstack
-      storageClass: "local-storage"
+      storageClass: $STORAGE_CLASS_NAME
```
SeanMooney, line 51: oh ok you already have that here in this step

docs/openstack/node-selector.md
SeanMooney: so when customer deploy this in production i would not expect them to use a storage class called "local-storage" in general. themy may but i would expect them to use a storage class that is preconfigured in there openshift which may have another name.

we discussed this a bit on slack and I'm not convinced we whoudl move our default testing to lvms.

we use crc-csi-hostpath-provisioner in ci for non adoption jobs because make crc storage was unreliable in the past when using the nested virt crc.

that may have resolved it self with the move to split out crc.

to me deviating form crc-csi-hostpath-provisioner in the zuul based jobs is an unnecessary risk and maintenance burden so i would reserve testing with lvms for baremetal jobs that do not use crc.
i.e. jobs that use SNO or similar.

I'm not convinced that we should have make lvms_storage target in install yamls. we could but presumably that would need to either
modify the CRC VM with an additional disk to use for the lvm topology operator and install and configure that operator
or it would have to create a file in the CRC VM and mount that as a loopback device which is then used by the lvm topology operator.

the loop back device concerns me due to the io overhead with that configuration. we do use that for cinder lvm with devstack but my experience with the nova gate is we do have io issues in the lvm based jobs when they run on a slow node. i would be concerned with recreating that problem in our zuul ci with CRC.

our operator ci is sufficiently different that perhaps we wont see the lvm issues we see with loop back devices there however its a concern. I'm not convinced that modifying the ci-framework to add cinder volumes to the zuul vms is a good approach either nor do i think modifying the CRC bundle image to add an additional portion that can be used for the lvm topology operator is a good approach. 
both add technical debt that i think we we would be better to avoid.

unless the is a very strong reason not continue to use crc-csi-hostpath-provisioner as our storage class for all crc based zull jobs i would be strongly in favor of adopting that for the adoption jobs too.

bogdando: addressed @fultonj 's @cescgina 's comments

bogdando:
> Perhaps this PR can be updated to keep `local-storage` by default but to make it easy to use `crc-csi-hostpath-provisioner` by passing a parameter?

done

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -1,3 +1,46 @@
...
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc apply -f - <<EOF
+    {{ local_sc_result.stdout | from_yaml | combine({"reclaimPolicy": storage_reclaim_policy, "resourceVersion": omit, "uuid": omit, "creationTimestamp": omit, "metadata": {"annotations": omit}}) }}
+    EOF
+
+- name: create a local-storage SC from crc-csi-hostpath-provisioner SC
+  when: local_sc_result.changed
```
marios, line 24: add  # noqa: no-handler

```
when: local_sc_result.changed # noqa: no-handler
```

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -1,3 +1,46 @@
...
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc get sc {{ storage_class_name }} -o yaml
+  register: local_sc_result
+  failed_when:
+    - local_sc_result.rc != 0
+    - ('Error from server (NotFound)' not in local_sc_result.stdout)
+  changed_when: local_sc_result.rc != 0
```
marios, line 11: i had to think about that a bit... like does it need ignore_errors. but I think it is ok: 

It will fail if there is no such SC and the error is something other than "not found" eg there was some real issue talking to the kluster or some internal error there

marios (commented):
first pass

bogdando:
@fultonj I believe I've addressed your comments, PTAL

docs/openstack/backend_services_deployment.md
```diff
@@ -63,10 +63,38 @@ podified OpenStack control plane services.
...
 ## Pre-checks
 
 ## Procedure - backend services deployment
 
+* Create a storage class named `local-storage`. For real deployments, you may want to
+  create it from
+  [LVMS](https://docs.openshift.com/container-platform/4.14/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html)
+  SC instead (or reference it directly in `openstackcontrolplanes` CR and its
+  `spec.glanceAPIs`). In this guide, we create a `local-storage` SC from the
+  default `crc-csi-hostpath-provisioner` SC for CRC:
```
fultonj, line 81: ack

fultonj:
> @fultonj I believe I've addressed your comments, PTAL

Yes, LGTM. Thanks.

fultonj (commented):
I'm fine with these proposed changes as they are right now.

tests/roles/backend_services/tasks/main.yaml
fultonj: We should fix these linter issues though

docs/openstack/node-selector.md
fultonj: > so when customer deploy this in production i would not expect them to use a storage class called "local-storage" in general. themy may but i would expect them to use a storage class that is preconfigured in there openshift which may have another name.
> 
> we discussed this a bit on slack and I'm not convinced we whoudl move our default testing to lvms.

We don't have to resolve this debate in order to merge this patch. I'm OK with merging it as @bogdando has right now and we can talk more about LVMS vs other backends. I think it depends on the job.

> we use crc-csi-hostpath-provisioner in ci for non adoption jobs because make crc storage was unreliable in the past when using the nested virt crc.
> 
> that may have resolved it self with the move to split out crc.
> 
> to me deviating form crc-csi-hostpath-provisioner in the zuul based jobs is an unnecessary risk and maintenance burden so i would reserve testing with lvms for baremetal jobs that do not use crc. i.e. jobs that use SNO or similar.

I wouldn't limit it only to baremetal. There are going to be jobs deployed by ci-framework which use devscripts-deployed OCP on VMs. I think LVMS would be good for them to test with.

For the [HCI VA](https://github.com/openstack-k8s-operators/architecture/tree/main/examples/va/hci) I would like to move the control-planes PVCs to LVMS and I'd like that also for jobs which will test it, specifically ci-framework jobs. I'm less concerned with crc jobs for that.

> I'm not convinced that we should have make lvms_storage target in install yamls. we could but presumably that would need to either modify the CRC VM with an additional disk to use for the lvm topology operator and install and configure that operator or it would have to create a file in the CRC VM and mount that as a loopback device which is then used by the lvm topology operator.
> 
> the loop back device concerns me due to the io overhead with that configuration. we do use that for cinder lvm with devstack but my experience with the nova gate is we do have io issues in the lvm based jobs when they run on a slow node. i would be concerned with recreating that problem in our zuul ci with CRC.
> 
> our operator ci is sufficiently different that perhaps we wont see the lvm issues we see with loop back devices there however its a concern. I'm not convinced that modifying the ci-framework to add cinder volumes to the zuul vms is a good approach either nor do i think modifying the CRC bundle image to add an additional portion that can be used for the lvm topology operator is a good approach. both add technical debt that i think we we would be better to avoid.
> 
> unless the is a very strong reason not continue to use crc-csi-hostpath-provisioner as our storage class for all crc based zull jobs i would be strongly in favor of adopting that for the adoption jobs too.

fultonj: > so when customer deploy this in production i would not expect them to use a storage class called "local-storage" in general. themy may but i would expect them to use a storage class that is preconfigured in there openshift which may have another name.

I think we should have a storage class recommendation for customers to use in production which uses the local storage of the OCP worker nodes and does not require them to purchase an additional storage solution. To require such a purchase would be hardware scope creep because we have supported the local disks of our controller nodes in previous OpenStacks for control plane services. I think we should use LVMS to implement this. I also want to ensure that any limitation from using this local storage (e.g. no support for RWX, but other issues could exist)  does not affect our control plane. I.e. we need to test what we recommend. I would call this storage class "local-storage" but it could be called anything. In our HCI VA its name is kustomizable:

https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/values.yaml#L201
https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/values-ci-framework.yaml#L216

I can imagine useful crc-based jobs which don't use LVMS too so we don't need to use it in 100% of the jobs, but if the control plane does not work with LVMS I'd consider it a blocker.

fultonj (commented):
I'm fine with merging this. Please just fix the linter errors first.

cescgina:
recheck added depends-on https://review.rdoproject.org/r/c/rdo-jobs/+/51177 to expose storage_class_name var to the tests

cescgina:
Moved a bit further now, failing with 
```
TASK [backend_services : create a local-storage SC from the given source SC] ***
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "set -euxo pipefail\n\n\noc apply -f - <<EOF\n{'apiVersion': 'storage.k8s.io/v1', 'kind': 'StorageClass', 'metadata': {'annotations': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d'}, 'parameters': {'storagePool': 'local'}, 'provisioner': 'kubevirt.io.hostpath-provisioner', 'reclaimPolicy': 'Delete', 'volumeBindingMode': 'WaitForFirstConsumer', 'resourceVersion': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d', 'uuid': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d', 'creationTimestamp': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d'}\nEOF\n", "delta": "0:00:00.665911", "end": "2024-01-10 05:57:52.732299", "msg": "non-zero return code", "rc": 1, "start": "2024-01-10 05:57:52.066388", "stderr": "+ oc apply -f -\nerror: error parsing STDIN: json: offset 2: invalid character '\\'' looking for beginning of object key string", "stderr_lines": ["+ oc apply -f -", "error: error parsing STDIN: json: offset 2: invalid character '\\'' looking for beginning of object key string"], "stdout": "", "stdout_lines": []}
``

File: docs/openstack/node-selector.md (hunk 1 above)
Comment (fultonj) [reply docs/openstack/node-selector.md]: > if it did not work with lvms it would be a regression

agreed.

> perhaps we shoudl expect operator lint with a check to catch any direct request for a PVC or type RWX and omit a warning or error? that shoudl prevent us from regressing?

That's an interesting idea. Do you know of any existing patterns which have that implemented?

Comment (SeanMooney) [reply docs/openstack/node-selector.md]: so @gibizer wrote the tool but i have not looked tat the details

https://github.com/gibizer/operator-lint

here is an example rule for crds
https://github.com/gibizer/operator-lint/tree/main/linters/crd/C003

you can see its using the golang ast tooling to inspect the code

https://github.com/gibizer/operator-lint/blob/main/linters/crd/C003/C003.go#L36-L50

i haven't played with that interface before but i suspect it would not be that difficult to check for the access mode

https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glance/pvc.go#L42-L45

we would need to parse that and validate the access mode

gibi will be back on Monday so perhaps we can discuss it with them next week and see how feasible that is

**Comment by cescgina** [comment]
recheck crc died in the last attempt: `dial tcp 10.217.0.116:9443: connect: connection refused"`

**Comment by bogdando** [comment]
> Moved a bit further now, failing with
> 
> ```
> TASK [backend_services : create a local-storage SC from the given source SC] ***
> fatal: [localhost]: FAILED! => {"changed": true, "cmd": "set -euxo pipefail\n\n\noc apply -f - <<EOF\n{'apiVersion': 'storage.k8s.io/v1', 'kind': 'StorageClass', 'metadata': {'annotations': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d'}, 'parameters': {'storagePool': 'local'}, 'provisioner': 'kubevirt.io.hostpath-provisioner', 'reclaimPolicy': 'Delete', 'volumeBindingMode': 'WaitForFirstConsumer', 'resourceVersion': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d', 'uuid': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d', 'creationTimestamp': '__omit_place_holder__c8ceeca0031379887f9fe2797350d7e93135e16d'}\nEOF\n", "delta": "0:00:00.665911", "end": "2024-01-10 05:57:52.732299", "msg": "non-zero return code", "rc": 1, "start": "2024-01-10 05:57:52.066388", "stderr": "+ oc apply -f -\nerror: error parsing STDIN: json: offset 2: invalid character '\\'' looking for beginning of object key string", "stderr_lines": ["+ oc apply -f -", "error: error parsing STDIN: json: offset 2: invalid character '\\'' looking for beginning of object key string"], "stdout": "", "stdout_lines": []}
> ```

It should be working now, at least I have passed that earlier on my dev setup

**Comment by bogdando** [comment]
recheck [download-cache-openstack-dnk8q](https://logserver.rdoproject.org/27/227/522e2402473971d6ff301105caa6bd87be1e833c/github-check/data-plane-adoption-OSP-17-to-extracted-crc/ec93054/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/test_with_ceph_out_2024-01-10T10:28:11EST.log)

**Comment by cescgina** [comment]
recheck

**Comment by bogdando** [comment]
it worked, PTAL

File: tests/roles/backend_services/tasks/main.yaml [hunk 9]
```diff
@@ -1,3 +1,54 @@
...
+        resourceVersion: ""
+        uuid: ""
+        creationTimestamp: ""
+        annotations: ""
+        name: local-storage
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc delete --ignore-not-found=true sc local-storage
+    echo -e '{{ local_sc_result.stdout | from_yaml | combine(_sc_params) | to_nice_yaml }}' | oc apply -f -
```
jistr, line 29: Maybe i don't fully grasp how this creation works but it's weird to me that there is no mention of the `storage_class_name` variable here at all? How does it refer to the "given source SC" then?

Other than that i'm ok with the change. I was fine with doing the crc storage cleanup/recreate manually but if this approach makes the storage recreation work "out of the box" then +1.

jistr: Was just confused with `changed_when: local_sc_result.rc != 0` which i parsed wrongly as if it was `  changed_when: local_sc_result.rc == 0`. I understand now :).

jistr (approved):

tests/roles/backend_services/tasks/main.yaml
bogdando: as explained on slack:
It copied SC from the source SC,
or creates one based on a default CRC SC, if no source to copy from.
To override this 'storage_class_name: lvms-magic-sc', that lvms will become a new source for local-storage SC
The next iteration will be to support direct renaming SC in all places

jistr:
Got a positive review from @fultonj as well but he doesn't have access to the merge button right now. Merging myself and i'll rebase #229 afterwards.

PR #229, opened 2024-01-05
Transform contributor documentation into AsciiDoc + some docs rendering improvements

jistr:
This moves the contributor documentation from the mkdocs site to the new asciidoc site. The mkdocs site now links to the new site for the contributor part. Also i've done a bunch of improvements in the Makefiles related to local docs rendering.

What's worth mentioning is that the docs are transformed from a multi-page into a single-page rendering, following the STF docs example. https://infrawatch.github.io/documentation/

The single-page vs multi-page aspect is something that we'll need to discuss and solve before moving the main user-facing documentation.


jistr:
I don't think this absolutely needs a green CI as the changes to the Makefile are pretty self-contained to docs only, but let's see if we can get it on the next attempt anyway...

jistr:
recheck

jistr:
Looks like the previous recheck didn't actually trigger a re-run, trying again.

jistr:
recheck

fao89 (approved):

fao89:
I think merge and iterate is the best way to go

jistr:
Thanks, this will collide with #227 so we'll merge that one first and then i'll rebase this.

klgill (approved):

PR #231, opened 2024-01-09
Adds the github-check layout to the data-plane-adoption repo

marios:
The github-check for this repo is currently defined at [1]. Instead we should move this in repo.

https://issues.redhat.com/browse/OSPRH-2997

[1] https://github.com/rdo-infra/review.rdoproject.org-config/blob/b13c3f957d6c4352c3cf976a732364074752c768/zuul.d/projects.yaml#L139

### DIFF
--- a/zuul.d/jobs-layout.yaml
+++ b/zuul.d/jobs-layout.yaml
@@ -1,0 +1,6 @@
+---
+- project:
+    name: openstack-k8s-operators/data-plane-adoption
+    github-check:
+      jobs:
+        - data-plane-adoption-OSP-17-to-extracted-crc


cescgina (approved):
lgtm, a couple of questions about the differences between the this project definition and the one in rdo config repo:
1. here we don't have a `default-branch`, I guess that zuul will by default pick up the branch used in the PR ( the docs seems to imply so: https://zuul-ci.org/docs/zuul/latest/config/project.html#attr-project.default-branch)
2. the rdo config uses a `system-required`, on a quick search I could not find its defintions, will we need that?

fao89 (approved):

cescgina:
> > lgtm, a couple of questions about the differences between the this project definition and the one in rdo config repo:
> > ```
> > 1. here we don't have a `default-branch`, I guess that zuul will by default pick up the branch used in the PR ( the docs seems to imply so: https://zuul-ci.org/docs/zuul/latest/config/project.html#attr-project.default-branch)
> > ```
> 
> we will keep the job definition and also existing project definition in rdo-jobs/config repos i.e. https://github.com/rdo-infra/review.rdoproject.org-config/blob/b13c3f957d6c4352c3cf976a732364074752c768/zuul.d/projects.yaml#L139 (and see https://review.rdoproject.org/r/c/config/+/51172
> 
> You can see all the operators have the project definition in the projects.yaml with default-branch. On the github side we are omitting that e.g. https://github.com/openstack-k8s-operators/install_yamls/blob/6ff6f43c2b45e536d5a78d9506bb5bd758e3a3d7/zuul.d/projects.yaml#L2-L4 and also https://github.com/openstack-k8s-operators/ci-framework/blob/acbb40b2deb4555ea8e0bfed578b6f2a42f20451/zuul.d/project-templates.yaml#L28-L32
> 
> > ```
> > 2. the rdo config uses a `system-required`, on a quick search I could not find its defintions, will we need that?
> > ```
> 
> I had to go looking for that. So we _will_ still have that, in the config repo i.e. https://review.rdoproject.org/r/c/config/+/51172 so we'll keep the project definition with the system-required.
> 
> As to what that is exactly :) and i had to dig, looks like system-required is defined there https://github.com/rdo-infra/rdo-jobs/blob/11fc06ca88616f30d1a94fcd2ce174b66487108d/zuul.d/project-templates.yaml#L4 But is seems to be a noop. I wonder if we still need to keep that. Digging a bit further I found that merge-check is defined there https://review.rdoproject.org/r/plugins/gitiles/config/+/refs/heads/master/zuul.d/pipelines.yaml#204
> 
> Anyway that is beyond the scope of just this change as we have that system-required on all the operators as linked above.
> 
> One thing I'm not sure about is merge order this vs https://review.rdoproject.org/r/c/config/+/51172
> 
> I think this one first as zuul seems to be running here without any complaints so far lets see if it reports without issue and if so we can go ahead with this, then try merging 51172 and hopefully not see any explosions.

thanks for the explanations, now I know were the noop job come from :)
I think it should be safe to merge this first, as it's duplicating the layout

PR #236, opened 2024-01-11
Fix resetting storage dependency on PCP

bogdando:
This is an attempt to fix the storage cleanup regression after https://github.com/openstack-k8s-operators/data-plane-adoption/pull/227

Storage cannot be reset until podified openstack control plane CRs
deleted, because of finalizers.

Fix condition to allow resetting storage only when resetting PCP
as well. Remove hacks to erase finalizers (but one for keystoneapis,
which should normally be deleted during oscp CR deletion).

Add removal of osp-secret, and osdpns steps.

Fix the ordering and waiting of events when resetting storage and PCP:

* Disable galera operator by scaling-in its operator deployment to 0
  (this is the only way as openstack controller webhook prohibits
  disabling/scaling-in galera replicas to 0, because this would
  break all openstack components under normal circumstances)
* Fully scale-in statefull sets for galera instances of all cells
* Remove galera databases pvcs, for all Nova cells
* Enable galera operator back to start DB instances back with cleaned
  up storage PVs. (It will recreate PVCs back, after the oscp CR applied)

### DIFF
--- a/docs_dev/assemblies/development_environment.adoc
+++ b/docs_dev/assemblies/development_environment.adoc
@@ -306,12 +306,21 @@ against your environment.
 
 The development environment must be rolled back in case we want to execute another Adoption run.
 
-Delete the data-plane and control-plane resources from the CRC vm
+Delete the data-plane and control-plane resources
 
 [,bash]
 ----
-oc delete osdp openstack
-oc delete oscp openstack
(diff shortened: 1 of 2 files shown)


bogdando:
I keep getting keystone-db-sync errors when respinning the adoption tests (CI does not cover that)

"Access denied for user 'keystone'@'<job-pod-ip>' (using password: YES)")

PR #239, opened 2024-01-15
Revert "Make local-storage SC options more flexible"

bogdando:
This reverts commit 4c0e1683ea8356076980d2760324df9f0d3354b2.

There is a problem with re-spinning databases PVCs for the adoption deveolpment setups, which should be addressed in order to restore this change back.


bogdando:
@jistr before we revert this, I would firstly check, if developers could still keep using make crc storage targets and getting away with a ``storage_class_name: local-storage`` setting in tests/vars.yaml instead of crc-csi-hostpath-provisioner

In CI, no need to change anything

bogdando:
so, the old data, picked up by recreated PVC, and it sits in /var/lib/csi-hostpath-data/

bogdando:
> @jistr before we revert this, I would firstly check, if developers could still keep using make crc storage targets and getting away with a `storage_class_name: local-storage` setting in tests/vars.yaml instead of crc-csi-hostpath-provisioner. I would simply change the default then, and restored the make targets in docs
> 
> In CI, no need to change anything

there is still this option to try

jistr (approved):
I would prefer this over #240, to not diverge CI and dev env config more than necessary.

bogdando:
@jistr I set it back to hold as I need a confirmation if the subject regression is real, or is it my env specific.
After "reverted" (by https://github.com/openstack-k8s-operators/data-plane-adoption/pull/240) on my env, with make crc_storage_clean approach in place (confirmed by logs), I do still observe exactly the same problem with keystone not accessing DB cuz of missing creds!

See the ansible log from my build set based on patches https://github.com/openstack-k8s-operators/data-plane-adoption/pull/240 and https://github.com/openstack-k8s-operators/data-plane-adoption/pull/236

```
TASK [pcp_cleanup : reset CRC storage] *****************************************
changed: [localhost] => {"changed": true, "cmd": "set -euxo pipefail

eval $(crc oc-env)
GALERA_OPERATOR_VERSION=mariadb-operator.v0.0.1

oc delete --ignore-not-found=true pod ovn-copy-data
oc delete --ignore-not-found=true pvc ovn-data

oc patch csv $GALERA_OPERATOR_VERSION -n openstack-operators  --type=json -p=\"[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 0}]\"

oc delete --ignore-not-found=true mariadbdatabases openstack
oc delete --ignore-not-found=true galera openstack
oc delete --ignore-not-found=true galera openstack-cell1
oc scale statefulsets openstack-galera --replicas=0 || true
oc scale statefulsets openstack-cell1-galera --replicas=0 || true

while oc get pod | grep rabbitmq-server-0; do
    sleep 2
done
while oc get pod | grep openstack-galera-0; do
    sleep 2
done
while oc get pod | grep openstack-cell1-galera-0; do
    sleep 2
done

oc delete --ignore-not-found=true pvc mysql-db-openstack-galera-0
oc delete --ignore-not-found=true pvc mysql-db-openstack-cell1-galera-0

oc patch csv $GALERA_OPERATOR_VERSION -n openstack-operators  --type=json -p=\"[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 1}]\"
cd ~/install_yamls
for i in {1..3}; do make crc_storage_cleanup crc_storage && break || sleep 5; done
",
"delta": "0:00:56.450020",
"end": "2024-01-17 12:52:31.890938",
"rc": 0, "start": "2024-01-17 12:51:35.440918",
```
^^ you can see that make crc storage cleanup has been done, albeit galera PVCs created later on that local-storage still do have that 'dirty state' problem for me.

So please kindly confirm if this issue is real or not (w/o reverting this yet).

bogdando:
> So please kindly confirm if this issue is real or not (w/o reverting this yet).

I re-tested re-spinning of an adoption env with a clean revert done, after CRC reverted to clean state as well. Then the issue is gone. So I'm affraid the revert is still needed
(that was just the state of crc becomes a such, that after it became 'broken', none of the clean up approaches helps but reverting CRC)

jistr:
:+1: Thank you.

PR #240, opened 2024-01-15
Undo local-storage default SC change

bogdando:
There is a problem with re-spinning databases PVCs for the adoption
deveolpment setups with storage_class_name crc-csi-hostpath-provisioner

Until it is addressed, use the pre-created local-storage with
install_yamls make storage targets.

Change nothing for CI builds, as it doesn't re-spin PVCs for galera
DBs, and storage_class_name crc-csi-hostpath-provisioner works for it.

Remove no longer existing rhev sample vars file.
Set back ovn helper pod to use its own default SC.


jistr (changes requested):
This one splits the dev env vs. CI storage configuration which is IMO undesirable unless absolutely necessary. If we hit any issues related to storage, it would be difficult to debug/reproduce. I'd rather go with #240.

bogdando:
no need for this, neither to revert the base feature

PR #245, opened 2024-01-19
Update/os diff/doc

matbu:
This PR moves and updates the documatation related to os-diff for pulling the configuration from TripleO environment.


docs_user/assemblies/openstack_adoption.adoc
```diff
@@ -6,8 +6,8 @@ include::../modules/openstack-planning.adoc[leveloffset=+1]
 include::../modules/openstack-node-selector.adoc[leveloffset=+1]
 
 include::../modules/openstack-backend_services_deployment.adoc[leveloffset=+1]
-include::../modules/openstack-pull_openstack_configuration.adoc[leveloffset=+1]
 include::../modules/openstack-stop_openstack_services.adoc[leveloffset=+1]
+include::../modules/openstack-pull_openstack_configuration.adoc[leveloffset=+1]
```
jistr, line 10: We probably want the OS Diff step after the full control plane adoption, right before the `include::../modules/openstack-edpm_adoption.adoc[leveloffset=+1]` line?

matbu: OS Diff could be use before, for  example here: https://openstack-k8s-operators.github.io/data-plane-adoption/user/#_glance_adoption
But maybe we want a full check after the control plane adoption and then remove those parts from the services adoption (like Glance).
If so, yes the this line could be move right before the edpm adoption.

In anycase, the part for configurating the tool and pulling config can be done any time.

jistr: Ok i didn't recall the Glance procedure is referring to OS Diff already. So let's merge this PR as it is now, but we should probably increase consistency between how OS Diff use is documented across services in another PR.

We could either add an OS Diff section to all other service documents, or remove OS Diff from the Glance doc and put  the comprehensive OS Diff document before the data plane adoption. I'm thinking having all OS Diff content in that one document before data plane adoption may make it easier to maintain, and maybe also easier for users to orient themselves in the docs.

jistr (approved):

jistr:
I'll click "update branch with merge commit with the base branch" to trigger the new docs CI job and if that's green, i'll merge.

PR #246, opened 2024-01-22
Optionally deploy cinder-volume with a Ceph backend

fmount:
This patch patch the ctlplane to deploy cinder-volume with a ceph backend. If cinder_backend is not defined or is empty, no cinder-volume is deployed, preserving the current behavior.

Fixes: [OSPRH-2442](https://issues.redhat.com/browse/OSPRH-2442)


fmount:
Closing as @Akrog is going to address it w/ more improvements

PR #248, opened 2024-01-22
Tests: Cinder volume and backup support

Akrog:
This patch adds the cinder volume and backup services to the tests.

For now only Ceph is supported in both services, and the backup can also be deployed with swift.

By default none of these cinder services are deployed and we must use `cinder_volume_backend` and `cinder_backup_backend` to deploy them, as illustrated in `tests/playbooks/test_with_ceph.yaml`.

For cinder volume and backup services we'll not only confirm that the pods are running, but also that the services are reporting as being `up`.

Jira: OSPRH-2442

### DIFF
--- a/tests/playbooks/test_with_ceph.yaml
+++ b/tests/playbooks/test_with_ceph.yaml
@@ -22,6 +22,8 @@
   vars:
     glance_backend: ceph
     manila_backend: cephfs
+    cinder_volume_backend: ceph
+    cinder_backup_backend: ceph
   module_defaults:
     ansible.builtin.shell:
       executable: /bin/bash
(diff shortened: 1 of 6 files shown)


cescgina:
recheck

jistr (approved):
Nice, thanks!

Akrog:
Looks like we are not deploying cinder backup for some reason on the standalone, so there should be no backup service to adopt, and that's I can see an error in the cinder backup service logs.
The job doesn't fail because there seem to be a race condition, because cinder services always start as `up` and only after a minute or show they show as down if there are some initialization failures.
Will disable the backup in the tests.

fao89 (approved):

Akrog:
Do I need to rebase this PR to have it merged?

jistr:
No i think it's ok as-is. The "do-not-merge/hold" can be removed?

Akrog:
> No i think it's ok as-is. The "do-not-merge/hold" can be removed?

Yes, I've removed it now

jistr:
Thanks, merging :)

PR #254, opened 2024-01-26
Fix osdpd cleanup procedure to respin dev envs

bogdando:
Also delete other osdpds related to FFU.
Align the commands in tests with those given in docs.

### DIFF
--- a/docs_dev/assemblies/development_environment.adoc
+++ b/docs_dev/assemblies/development_environment.adoc
@@ -311,8 +311,23 @@ Delete the data-plane and control-plane resources from the CRC vm
 
 [,bash]
 ----
-oc delete osdp openstack
-oc delete oscp openstack
+oc delete --ignore-not-found=true --wait=false openstackdataplanedeployment/openstack
+oc delete --ignore-not-found=true --wait=false openstackdataplanedeployment/openstack-nova-compute-ffu
+oc delete --ignore-not-found=true --wait=false openstackcontrolplane/openstack
+oc patch openstackcontrolplane openstack --type=merge --patch '
+metadata:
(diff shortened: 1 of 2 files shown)


docs_dev/assemblies/development_environment.adoc
```diff
@@ -311,8 +311,23 @@ Delete the data-plane and control-plane resources from the CRC vm
 
 [,bash]
 ----
-oc delete osdp openstack
-oc delete oscp openstack
+oc delete --ignore-not-found=true --wait=false openstackdataplanedeployment/openstack
+oc delete --ignore-not-found=true --wait=false openstackdataplanedeployment/openstack-nova-compute-ffu
+oc delete --ignore-not-found=true --wait=false openstackcontrolplane/openstack
```
fao89, line 316: why setting wait to false if you wait it later?

jistr: This pattern was first introduced in e523c7175cb2b6ee99d48473e77ab27712215aea i think. The purpose was that we need to get the shell back to be able to clear the finalizers, otherwise the deletion gets stuck.

Maybe we only need --wait=false for `openstackcontrolplane/openstack` then... But also maybe it's not a blocker since these are being run manually.

Alternatively we could run the delete in background with `&`, then clear finalizers, then run `wait`. That's perhaps better, i just didn't think of it back then.

jistr: So this could be improved but my 2 cents would be it does not have to be a blocker given that it is not customer-facing.

jistr (approved):

fao89 (approved):

PR #260, opened 2024-01-30
Link to downstream instructions through corp url shortener

jistr:


jistr:
This should make it easier to find that page.

fao89 (approved):
thank you!

PR #263, opened 2024-01-31
Update EDPM

fao89:


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -200,9 +200,12 @@
         - install-os
         - configure-os
         - run-os
+        - reboot-os
```
jistr, line 203: I guess that reboot is conditional based on something and we wouldn't actually reboot during Adoption, right?

jistr (approved):
:+1: Thanks!

tests/roles/dataplane_adoption/tasks/main.yaml
fao89: it is conditional, but I don't remember what is the condition, so I'm not sure if it would happen during adoption

fao89:
recheck

fao89:
recheck

PR #264, opened 2024-01-31
Adding mod docs metadata

klgill:
Experimenting with adding metadata to files and testing the GitHub workflow.


docs_user/assemblies/openstack_adoption.adoc
```diff
@@ -1,5 +1,12 @@
+ifdef::context[:parent-context: {context}]
+
+[id="openstack-adoption_{context}"]
+
 = OpenStack adoption
-:toc: left
+
+:context: openstack-adoption
+//kgilliga: I don't think we need the toc and toclevels attributes.
```
jistr, line 8: Will we still get a sidebar for navigation though? I find the sidebar useful.

jistr: I guess we will because it's in main.adoc: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/main.adoc?plain=1#L2

thatdocslady:
@jistr is it possible to generate a mid-stream preview of the doc updates to make sure that the build works and the docs look as expected? I don't know if this is possible while a PR is in WIP mode or if it needs to be tagged as Ready for Review. Thx!

docs_user/assemblies/openstack_adoption.adoc
klgill: I was comparing this file with another "master.adoc"" (aka main.adoc) file, and I didn't see the toc attributes. I'm happy to leave them as is if they're necessary.

jistr:
@thatdocslady I forgot to mention -- we do have a check job that verifies that the build works at least. That's the "Docs Build" check with a green checkmark on this PR. There just doesn't seem to be a reasonably easy/practical solution to publish that HTML somewhere for viewing.

thatdocslady:
@jistr thanks for the clarification, I understand the limitations. Wondering if we could leverage the downstream GitLab solution that we use (Jenkins job with bccutil running inside it somehow). If you go to any MR link you should be able to see it, but of course we should consider whether it's something that we can manage publicly. I can also check with OCP to see what they use, since they author in GH in AsciiDoc.

fao89:
@jistr @thatdocslady I started this: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/269

jistr:
@fao89 That looks exactly like what we need, thanks a lot!!

docs_user/modules/openstack-backend_services_deployment.adoc
```diff
@@ -1,15 +1,21 @@
-= Backend services deployment
+[id="deploying-backend-services_{context}"]
+
+//:context: backend-services
+
+//kgilliga: The backend services deployment intro will be converted to an assembly. Most of the rest of the module will be converted to a procedure.
+
+= Deploying backend services
```
jistr, line 7: Maybe we could rename the files to match the headings, but that can be done in another PR.

docs_user/modules/openstack-backend_services_deployment.adoc
```diff
@@ -36,14 +42,12 @@ you've run `make crc_storage`.
 * Set the desired admin password for the podified deployment. This can
 be the original deployment's admin password or something else.
 +
-[,bash]
```
jistr, line 39: When we remove this, the block renders with white background, while the blocks which have this "header" render with black background. Not an issue if it's not an issue downstream, just making sure we know about it. Here's the preview:

https://logserver.rdoproject.org/64/264/dccf1357734825bc1672d7160e12a2926772cb95/github-check/adoption-docs-preview/29a7a68/docs_build/docs_build/adoption-user/index-upstream.html

jistr (approved):

fao89 (approved):

docs_user/modules/openstack-backend_services_deployment.adoc
klgill: Yes, I will be renaming files for Beta.

PR #267, opened 2024-02-01
Tests: Fix cinder defaults

Akrog:
The `defaults/main.yaml` was incorrectly called `defaults/main.yam` so it wasn't being used.

This means that running `test_minimal` would fail with:
  'cinder_volume_backend' is undefined

This patch renames the file to have the right name.


bogdando (commented):
lgtm

jistr (approved):

PR #268, opened 2024-02-02
[glance_adoption] Ensure replicas is set for glance CR

karelyatin:
With [1] replicas needs to be explicitly set for glanceAPI instances while patching it along with backend configuration. This is needed for cases where empty spec for glanceAPIs is used during initial creation.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/417


fmount (approved):
/lgtm

jistr (approved):

PR #271, opened 2024-02-05
Store source cloud pulled mariadb data in env file

bogdando:
Allow resuming/starting test cases from the tasks like 'Verify MariaDB data imported', or 'Adopted Nova FFU post-checks' without loosing the facts for pulled source cloud data, which then is used for post-adoption verifications for MariaDB and Nova.

Allow using the source cloud data from env vars locally in each chapter, without carrying it over from another chapters. That is agreed aproach for documentation.  Achieve that by storing it in an env file, which can be sourced to get prepared values afterwards (those cannot be re-fetched again, after stopping the source cloud services).


fao89 (approved):

fao89:
please fix the conflict

PR #274, opened 2024-02-06
Remove manual step to switch ovn-controllers to new db servers

booxter:
This switch (ovn-remote configuration, ovn-controller service re-starts) will happen as part of EDPM deployment of `ovn` Service. So there is no need - and actually there are risks due to point of no return concerns - to execute the switch at this point in adoption procedure.

Since northd should still be down during database backups, added northd stop to the list used to stop other control plane services at the start of adoption procedure.

Closes: https://issues.redhat.com/browse/OSPRH-2303


karelyatin:
recheck adoption job

slawqo (commented):
/lgtm for me

jistr (approved):
LGTM, thanks!

PR #276, opened 2024-02-07
Add os-diff automation into the testsuite roles

matbu:


tests/playbooks/test_minimal.yaml
```diff
@@ -20,6 +20,10 @@
   hosts: local
   gather_facts: false
   force_handlers: true
+  vars:
+    # @matbu: Should overrided by Zuul and be removed from here
```
jistr, line 24: The test suite should be easily executable from a lab machine or a laptop that is connecting to a lab machine. Rather than using absolute paths, we can use paths relative to the playbook being run, like:

https://github.com/openstack-k8s-operators/data-plane-adoption/blob/0ea5b7198572f74c12a6b4695e270c40f6c59f5e/tests/roles/mariadb_copy/defaults/main.yaml

tests/playbooks/test_minimal.yaml
```diff
@@ -20,6 +20,10 @@
...
+    dpa_dir: "/home/zuul/src/github.com/openstack-k8s-operators/data-plane-adoption"
```
jistr, line 25: I'd rename `dpa_dir` to `od_data_dir` or `os_diff_data_dir`?

tests/playbooks/test_minimal.yaml
```diff
@@ -20,6 +20,10 @@
...
+    od_dir: "/home/zuul-worker/src/github.com/openstack-k8s-operators/os-diff"
   module_defaults:
     ansible.builtin.shell:
       executable: /bin/bash
```
jistr, line 29: In the roles list below, 'pull_openstack_configuration' should go right before 'dataplane_adoption'. (Same in the ceph playbook.)

tests/roles/keystone_adoption/tasks/main.yaml
```diff
@@ -68,3 +68,22 @@
     for service in cinderv3 glance neutron nova placement swift; do
       ${BASH_ALIASES[openstack]} service list | awk "/ $service /{ print \$2; }" | xargs ${BASH_ALIASES[openstack]} service delete || true
     done
+
```
jistr, line 71: We should have per-service task files in the pull_openstack_configuration role instead of having OS Diff snippets in the service roles?

To match the documentation flow, we want to execute OS Diff before the EDPM adoption step. Users more familiar with Adoption may want to run it in different order than the docs are written, but i think the test suite should stick with the planned docs order.

tests/roles/pull_openstack_configuration/tasks/main.yaml
```diff
@@ -1,3 +1,52 @@
+# OS Diff automation steps
+- name: Make sure Golang is installed
+  failed_when: false
+  become: true
+  ansible.builtin.shell: |
+    dnf -y install golang
```
jistr, line 6: We should only execute this if `go` binary is not found? IIRC most dev setups install Go differently than from the RPM.

jistr (commented):
Thanks, i wrote a few comments inline.

tests/roles/pull_openstack_configuration/defaults/main.yaml
```diff
@@ -0,0 +1,3 @@
+dpa_dir: "/home/zuul/src/github.com/openstack-k8s-operators/data-plane-adoption"
+os_diff_data_dir: "/home/zuul-worker/src/github.com/openstack-k8s-operators/os-diff"
+standalone_ip: 192.168.122.100
```
jistr, line 3: Sorry i didn't realize this in the original review, but this should probably be called something like `source_os_diff_config_ip` and it should go here in this file among the other similar variables:

https://github.com/openstack-k8s-operators/data-plane-adoption/blob/6be9a9f0888ab5dbc41c4ca90f0ae637663bd129/tests/vars.sample.yaml#L24

We don't want to specifically mention `standalone` in the variables because the test suite needs to work with 3-controller environments as well.

tests/roles/pull_openstack_configuration/defaults/main.yaml
```diff
@@ -0,0 +1,3 @@
+dpa_dir: "/home/zuul/src/github.com/openstack-k8s-operators/data-plane-adoption"
+os_diff_data_dir: "/home/zuul-worker/src/github.com/openstack-k8s-operators/os-diff"
```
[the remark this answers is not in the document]
matbu: Sorry for my late response.
I just updated the PR but since the package are still not ready we are in a mix-spot where the binary and the config file is coming from the github-repo.
With rpm we will have the binary in /usr/bin available and the config file in /etc/os-diff and the github source path could be dropped.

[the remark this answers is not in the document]
matbu: I have put relative path but I'm not sure, in the Zuul environment the relative paths will work.

tests/roles/pull_openstack_configuration/defaults/main.yaml
```diff
@@ -0,0 +1,3 @@
+dpa_dir: "~/data-plane-adoption"
+os_diff_data_dir: "~/os-diff"
```
jistr, line 2: These will pollute user's home directory... Is there a reason not to do what i suggested earlier?

```
# change os_diff_data_dir into:
os_diff_dir: tmp/os-diff

# change dpa_dir into:
os_diff_data_dir: tmp/os-diff/data
```

This also includes descriptive naming of the variable and not creating directories with test data that aren't yet in the `.gitignore` file. The `playbooks/tmp` directory is already in gitignore. So i think the above should "just work" both in context of user environments and CI.

jistr (changes requested):

tests/roles/pull_openstack_configuration/defaults/main.yaml
jistr: Note these paths are not in `/tmp`, they are relative to the playbook, so in effect they are `<data_plane_adoption_repo>/tests/playbooks/tmp`.

jistr (approved):
Thanks, looks good. The PR has many commits so we'll use "squash and merge" functionality.

PR #281, opened 2024-02-09
DP3 amendments part 1

jistr:
(see commit messages)


klgill:
Regarding moving "About node selectors" under "Planning the new deployment":  Can "About node selectors" remain in its own file? For Beta, I'm going to be heavily restructuring these files. That means that nearly all headings 2s (==) will be broken out into their own files anyway. So I was thinking that for Beta, "Planning the new deployment" would become a top-level assembly, and then "About node selectors" would be a module embedded within that assembly.

I hope this makes sense. Let me know what you think.

Otherwise, this PR looks good.

jistr:
To make sure i understand what you mean: "About node selectors" would stay in the file where it is now, but instead of including it from `assemblies/openstack_adoption.adoc` we would include it from `modules/openstack-planning.adoc` to make the content appear in the spot where it is in this PR? IOW, we'd be including a module from inside a module?

klgill:
I meant that "About node selectors" would stay in the file where it is now, and still be included in the assemblies/openstack_adoption.adoc file. Basically, just keep it where it was originally.

jistr:
Ack i'll keep it where it is for now. I removed the commit that moved the "About node selector" part.

fao89 (approved):

jistr:
nova-compute:
Error during ComputeManager.update_available_resource: oslo_messaging.exceptions.MessagingTimeout: Timed out waiting for a reply to message ID d02c449e14744f8087882398c92cc640
 
 https://logserver.rdoproject.org/81/281/31634864864c361d4544526d8713bebe8b98429d/github-check/data-plane-adoption-osp-17-to-extracted-crc/345f04d/standalone/containers/nova/nova-compute.log

jistr:
recheck

klgill (approved):

PR #282, opened 2024-02-12
[DP3] Remove mentions of OS Diff for DP3

jistr:
OS Diff will only come after DP3.


fao89 (approved):

klgill:
Just one question: Do you want to comment out the text related to os-diff rather than delete it, in case we need it after DP3?

jistr:
The content stays untouched on the main branch, this PR only affects the DP3 branch. The further branches for beta/GA should be forked again from main, not from dev-preview3 branch.

For example, after we merge this, all the OS Diff content would still be present in https://openstack-k8s-operators.github.io/data-plane-adoption/user/ .

jistr:
We just need to make sure when we want to take the DP3 docs downstream, we have to use the dev-preview3 branch which was created for this purpose, rather than the main branch.

klgill:
> The content stays untouched on the main branch, this PR only affects the DP3 branch. The further branches for beta/GA should be forked again from main, not from dev-preview3 branch.
> 
> For example, after we merge this, all the OS Diff content would still be present in https://openstack-k8s-operators.github.io/data-plane-adoption/user/ .

Yes, that makes sense.

PR #284, opened 2024-02-13
Pull service configurations before stopping them

yadneshk:
Configuration cannot be pulled once the containers are stopped.


jistr:
I thought OS Diff just uses SSH and looks at the config in the filesystem? That should make it ok to fetch even after services are stopped... Our intention is to move OS Diff usage much later in the process actually, right before EDPM adoption. So if that doesn't work we'll have to change our plans... cc @matbu @davidjpeacock

yadneshk:
If I stop keystone container and try pulling config from wallaby, I see this
```
$ ./os-diff pull --verbose
Error, Podman name not found, skipping ...keystone
```
If keystone container is started, this works

Isn't this how we pull service config https://github.com/openstack-k8s-operators/os-diff/blob/main/pkg/servicecfg/utils.go#L68 ?

matbu:
@yadneshk hm right this is actually a bug while I rewrote the "pull" part, it only looks for "running" containers. I will fixe it right now.

yadneshk:
@matbu I tried this again with your recent patch and seems to work fine now. Thank you! Closing this PR.

PR #288, opened 2024-02-15
Swift backend

fmount:


tests/roles/glance_adoption/tasks/main.yaml
```diff
@@ -76,37 +76,82 @@
...
-            code=0
-        fi
-    done
-    exit $code
-  register: glance_running_result
-  until: glance_running_result is success
-  retries: 60
-  delay: 2
+- name: Check the resulting Glance
+  when: glance_backend != 'swift'
```
fmount, line 122: @jistr when patch https://github.com/openstack-k8s-operators/data-plane-adoption/pull/291 lands, we can improve this part and have an additional task in `swift` to `patch` the glance replicas.

fmount: the above is just an example, I suspect we have to both retrieve the right `osctlplane` name and pass the right path to the `check_glance.yaml` tasks.

fmount:
@cescgina do you know how to trigger a `data-plane-adoption` job where we don't have Ceph (I realized in that case Swift will be enabled by default)  and at the same time trigger the swift adoption and pass `glance_backend: swift`?

tests/roles/glance_adoption/defaults/main.yaml
```diff
@@ -1,2 +1,2 @@
-# glance_backend can be 'local' or 'ceph'
-glance_backend: local
+# glance_backend can be 'local', 'ceph' or 'swift'
+glance_backend: swift
```
fmount, line 2: @jistr so this should trigger `test_minimal` to deploy glance w/ a swift backend. While this would be the default behavior, I'm wondering if you think it's better to pass the var to the playbook itself.

cescgina:
> @cescgina do you know how to trigger a `data-plane-adoption` job where we don't have Ceph (`EDPM_COMPUTE_CEPH_ENABLED=false`: I realized in this case `Swift` is enabled by default) and at the same time trigger the swift adoption and pass `glance_backend: swift`?
@fmount Hi! I was out for a couple of days. So the test-suite is currently hardcoded, but it should be fairly easy to add a parameter to choose the test case to run. Would we also need some change in the standalone deployment?

fmount:
> > @cescgina do you know how to trigger a `data-plane-adoption` job where we don't have Ceph (`EDPM_COMPUTE_CEPH_ENABLED=false`: I realized in this case `Swift` is enabled by default) and at the same time trigger the swift adoption and pass `glance_backend: swift`?
> > @fmount Hi! I was out for a couple of days. So the test-suite is currently hardcoded, but it should be fairly easy to add a parameter to choose the test case to run. Would we also need some change in the standalone deployment?

It should be enough to have `EDPM_COMPUTE_CEPH_ENABLED=false` in the standalone deployment, then we should switch from the `test_with_ceph` to `test_minimal`. The current patch already added `glance_backend: swift` as default, so it will be aligned w/ a TripleO deployment without Ceph.

fmount:
> > > > @cescgina do you know how to trigger a `data-plane-adoption` job where we don't have Ceph (`EDPM_COMPUTE_CEPH_ENABLED=false`: I realized in this case `Swift` is enabled by default) and at the same time trigger the swift adoption and pass `glance_backend: swift`?
> > > > @fmount Hi! I was out for a couple of days. So the test-suite is currently hardcoded, but it should be fairly easy to add a parameter to choose the test case to run. Would we also need some change in the standalone deployment?
> > 
> > 
> > Hi, thanks a lot for the help! It should be enough to have `EDPM_COMPUTE_CEPH_ENABLED=false` in the standalone deployment, then we should switch from the `test_with_ceph` to `test_minimal`. The current patch already added `glance_backend: swift` as default, so it will be aligned w/ a TripleO deployment without Ceph.
> 
> ack, then it should be fairly easy I think, I'm creating the rdo-jobs patch to enable it. I'll also need to edit this PR to add the job variant, do you mind if I add a commit? What I'm not sure about is when to trigger this job, maybe on changes to swift_adoption role?

Np at all, feel free to add the commit we need to try the new job! 
I'm not sure either, I would say that depending on the topology of the source cloud (right now it's with or without Ceph), we can trigger one playbook or the other.

fmount:
> > > > > @cescgina do you know how to trigger a `data-plane-adoption` job where we don't have Ceph (`EDPM_COMPUTE_CEPH_ENABLED=false`: I realized in this case `Swift` is enabled by default) and at the same time trigger the swift adoption and pass `glance_backend: swift`?
> > > > > @fmount Hi! I was out for a couple of days. So the test-suite is currently hardcoded, but it should be fairly easy to add a parameter to choose the test case to run. Would we also need some change in the standalone deployment?
> > > 
> > > 
> > > Hi, thanks a lot for the help! It should be enough to have `EDPM_COMPUTE_CEPH_ENABLED=false` in the standalone deployment, then we should switch from the `test_with_ceph` to `test_minimal`. The current patch already added `glance_backend: swift` as default, so it will be aligned w/ a TripleO deployment without Ceph.
> > 
> > 
> > ack, then it should be fairly easy I think, I'm creating the rdo-jobs patch to enable it. I'll also need to edit this PR to add the job variant, do you mind if I add a commit? What I'm not sure about is when to trigger this job, maybe on changes to swift_adoption role?
I don't think so, we can adopt swift but have Glance set with a RBD backend. I think the main driver should be the topology defined in the src cloud (`EDPM_COMPUTE_CEPH_ENABLED=false` should be enough for now).
> 
> Np at all, feel free to add the commit we need to try the new job! I'm not sure either, I would say that depending on the topology of the source cloud (right now it's with or without Ceph), we can trigger one playbook or the other.

cescgina:
@fmount  the new job is running. To clarify on my last comment, I was thinking about on which PRs should we run test-minimal. Do we want to always test both test-suites? Or only run test-minimal without ceph for some PRs like this one?

cescgina:
recheck

tests/roles/glance_adoption/defaults/main.yaml
jistr: Yea the adoption has to follow the configuration of the environment we start with. I'm actually not sure what the Glance backend is when we deploy the Standalone without Ceph, but based on the defaults it's probably Swift, unless it's overriden somewhere else:

https://github.com/openstack/tripleo-heat-templates/blob/811bf370be3b9a4bb29b7234a6f6c16b8290146b/deployment/glance/glance-api-container-puppet.yaml#L227

So changing the test-minimal default to Swift backend makes sense to me. We could feed the variable to the playbook too, but i'm not sure if we want to test with `local` ever, now that we have Swift... So maybe it's ok just to change the default.

docs_dev/assemblies/development_environment.adoc
```diff
@@ -101,10 +101,15 @@ cat << EOF > /tmp/addtional_nets.json
...
-export NTP_SERVER=pool.ntp.org  # Only neccecary if not on the RedHat network ...
+export NTP_SERVER=pool.ntp.org  # Only necessary if not on the RedHat network ...
 export EDPM_COMPUTE_CEPH_ENABLED=false  # Optional
 ----
 
+[Note]
+===
+If `EDPM_COMPUTE_CEPH_ENABLED=false` is set, TripleO configures `Glance` with
+`Swift` as a backend.
+===
```
jistr, line 112: :+1:

tests/roles/swift_adoption/tasks/main.yaml
```diff
@@ -89,3 +89,24 @@
   register: swift_responding_result
   until: swift_responding_result is success
   retries: 15
+
+- name: Enable Glance is swift is set as backend
```
jistr, line 93: Glance adoption comes after Swift adoption, both in docs and in test suite order. We should remove this and just adopt Glance with `replicas: 1` from the start, as Swift is already adopted at that point?

fmount: aah ok great, I assumed glance was one of the first services being adopted, so I adapted the code for that kind of situation.
For the same reason, do we want to adopt cinder first and then glance just because cinder can be used as a glance backend? I'm fine setting it to `replicas: 0` and patch it when we know that `cinderVolume` is active. 
For this patch I agree we don't need this kind of orchestration. Updating it in my next PS.

tests/roles/glance_adoption/defaults/main.yaml
fmount: +1 to keep `swift` as default (which is aligned w/ what TripleO provides). `local` shouldn't be tested at all, but we can keep that code because w/ a slightly change we can be ready to test the `NFS` backend.

fmount:
@cescgina I think I updated the patch and by mistake I reverted your changes! Can you restore your commit that allowed us to test `minimal`? (Thanks and sorry :/)

jistr:
@fmount Yea i think we should move Cinder before Glance for the same reason.

fmount:
Looks like this is failing on `swift`:
```
FAILED - RETRYING: [localhost]: check that Swift is reachable and its endpoints are defined (2 retries left).
FAILED - RETRYING: [localhost]: check that Swift is reachable and its endpoints are defined (1 retries left).
```
and looking at the logs I see many auth failures [1].

[1] https://logserver.rdoproject.org/88/288/4f964feec2300f92c79eda8c96eb667e2cb4c707/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/e069a20/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/swift-proxy-5cb844d88d-mtf2d/logs/proxy-server.log

cescgina:
recheck trying to hold an env

cescgina:
> Looks like this is failing on `swift`:
> 
> ```
> FAILED - RETRYING: [localhost]: check that Swift is reachable and its endpoints are defined (2 retries left).
> FAILED - RETRYING: [localhost]: check that Swift is reachable and its endpoints are defined (1 retries left).
> ```
> 
> and looking at the logs I see many auth failures [1].
> 
> [1] https://logserver.rdoproject.org/88/288/4f964feec2300f92c79eda8c96eb667e2cb4c707/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/e069a20/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/swift-proxy-5cb844d88d-mtf2d/logs/proxy-server.log
> 
> /cc @cschwede in case you want to look at the logs

@fmount @cschwede I'm holding the environment from the last run in case you need it for debug

danpawlik:
recheck

cschwede:
Hmm, I don't see Glance secrets, configmaps, pods or crs in here - what am I missing?

https://logserver.rdoproject.org/88/288/4f964feec2300f92c79eda8c96eb667e2cb4c707/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/21bf54b/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack

fmount:
> Hmm, I don't see Glance secrets, configmaps, pods or crs in here - what am I missing?
> 
> https://logserver.rdoproject.org/88/288/4f964feec2300f92c79eda8c96eb667e2cb4c707/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/21bf54b/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack

Swift adoption happens before Glance, that's the reason you don't see anything related to glance yet. After swift is adopted and swiftProxy running, a couple of steps later we adopt glance. What seems to be failing here is the Swift adoption itself.

cschwede:
> Swift adoption happens before Glance, that's the reason you don't see anything related to glance yet. After swift is adopted and swiftProxy running, a couple of steps later we adopt glance. What seems to be failing here is the Swift adoption itself.

Oh right, makes sense now.

I'm wondering about the network connectivity. Looks like the proxy-server running on CRC can't connect to the account-server on standalone:

proxy-server: ERROR with Account server 172.20.0.100:6002/d1 re: Trying to HEAD /v1/AUTH_29566cfbd05e4e09939d5880c10c13e3: ConnectionTimeout (0.5s) (txn: tx47fb3e2904524954bea70-0065df0002)
https://logserver.rdoproject.org/88/288/4f964feec2300f92c79eda8c96eb667e2cb4c707/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/21bf54b/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/swift-proxy-69f9dc9654-qlwmb/logs/proxy-server.log

And on the standalone deployment is no incoming request at all: https://logserver.rdoproject.org/88/288/4f964feec2300f92c79eda8c96eb667e2cb4c707/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/21bf54b/standalone/containers/swift/swift.log

I'm running a local test now.

cschwede:
> I'm running a local test now.

Local test finished successfully on main (without this patch), using this patch it failed after Swift deployed successfully with the following error:

```
TASK [glance_adoption : deploy podified Glance with Swift backend] *************
fatal: [localhost]: FAILED! => {"msg": "template error while templating string: unexpected '.'. String: {{ shell_header }}\n{{ oc_header }}\noc patch openstackcontrolplane openstack --type=merge --patch '\nspec:\n  glance:\n    enabled: true\n    template:\n      databaseInstance: openstack\n      customServiceConfig: |\n            [DEFAULT]\n            enabled_backends = default_backend:swift\n            [glance_store]\n            default_backend = default_backend\n            [default_backend]\n            swift_store_create_container_on_put = True\n            swift_store_auth_version = 3\n            swift_store_auth_address = {{ .KeystoneInternalURL }}\n            swift_store_endpoint_type = internalURL\n            swift_store_user = service:glance\n            swift_store_key = {{ .ServicePassword }}\n      storageClass: \"local-storage\"\n      storageRequest: 10G\n      glanceAPIs:\n        default:\n          replicas: 1\n          override:\n            service:\n              internal:\n                metadata:\n                  annotations:\n                    metallb.universe.tf/address-pool: internalapi\n                    metallb.universe.tf/allow-shared-ip: internalapi\n                    metallb.universe.tf/loadBalancerIPs: 172.17.0.80\n              spec:\n                type: LoadBalancer\n          networkAttachments:\n          - storage\n'\n. unexpected '.'"}
```

Regarding this CI issue - is there any reproducer on hold that I can have a look at?

fmount:
> > I'm running a local test now.
> 
> Local test finished successfully on main (without this patch), using this patch it failed after Swift deployed successfully with the following error:
> 
> ```
> TASK [glance_adoption : deploy podified Glance with Swift backend] *************
> fatal: [localhost]: FAILED! => {"msg": "template error while templating string: unexpected '.'. String: {{ shell_header }}\n{{ oc_header }}\noc patch openstackcontrolplane openstack --type=merge --patch '\nspec:\n  glance:\n    enabled: true\n    template:\n      databaseInstance: openstack\n      customServiceConfig: |\n            [DEFAULT]\n            enabled_backends = default_backend:swift\n            [glance_store]\n            default_backend = default_backend\n            [default_backend]\n            swift_store_create_container_on_put = True\n            swift_store_auth_version = 3\n            swift_store_auth_address = {{ .KeystoneInternalURL }}\n            swift_store_endpoint_type = internalURL\n            swift_store_user = service:glance\n            swift_store_key = {{ .ServicePassword }}\n      storageClass: \"local-storage\"\n      storageRequest: 10G\n      glanceAPIs:\n        default:\n          replicas: 1\n          override:\n            service:\n              internal:\n                metadata:\n                  annotations:\n                    metallb.universe.tf/address-pool: internalapi\n                    metallb.universe.tf/allow-shared-ip: internalapi\n                    metallb.universe.tf/loadBalancerIPs: 172.17.0.80\n              spec:\n                type: LoadBalancer\n          networkAttachments:\n          - storage\n'\n. unexpected '.'"}
> ```
> 
> Regarding this CI issue - is there any reproducer on hold that I can have a look at?

Ok this is something I can patch and it's probably related to the fact that we're templating things within ansible, so it gets confused about something like -> `swift_store_auth_address = {{ .KeystoneInternalURL }}`, that shouldn't be expanded by ansible. I can probably try to fix that locally!

cschwede:
The latest openstack-operator-index image is broken, tests are failing because there is no openstack deployment running:

https://logserver.rdoproject.org/88/288/d462d361c90fda1e76b2d1ac6bce73a9483f2ae1/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/30b366a/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/

https://logserver.rdoproject.org/88/288/d462d361c90fda1e76b2d1ac6bce73a9483f2ae1/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/30b366a/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/openstack-operator-index-9fq5m/logs/registry-server.log

tests/roles/glance_adoption/files/glance_swift.yaml
```diff
@@ -0,0 +1,37 @@
...
+      storageRequest: 10G
+      customServiceConfig: |
+        [DEFAULT]
+        enabled_backends = default_backend:swift
+        [glance_store]
+        default_backend = default_backend
+        [default_backend]
+        swift_store_create_container_on_put = True
+        swift_store_auth_version = 3
+        swift_store_auth_address = {{ .KeystoneInternalURL }}
```
cschwede, line 19: AFAICT KeystoneInternalURL and ServicePassword are not defined during the playbook run. ServicePassword might be replaced with glance_password though

tests/roles/glance_adoption/tasks/main.yaml
```diff
@@ -80,42 +80,7 @@
...
-                      annotations:
-                        metallb.universe.tf/address-pool: internalapi
-                        metallb.universe.tf/allow-shared-ip: internalapi
-                        metallb.universe.tf/loadBalancerIPs: 172.17.0.80
-                  spec:
-                    type: LoadBalancer
-              networkAttachments:
-              - storage
-    '
+    oc patch openstackcontrolplane openstack --type=merge --patch-file={{ lookup('ansible.builtin.file', 'glance_swift.yaml') }}
```
cschwede, line 83: --patch-file now has the parsed content of glance_swift.yaml, but not the filename of the parsed content file

cschwede (changes requested):

tests/roles/glance_adoption/tasks/main.yaml
fmount: right, I was wrong about using the `lookup` filter. I'm going to update this line with the new patchset and try to rely on the special, builtin variable `role_path` [1].

[1] https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html#term-role_path

fmount:
@cescgina can we recheck and hold the environment? This failure is weird, I see the [endpoints are properly created](https://logserver.rdoproject.org/88/288/58b7d2f18c2b429f524c0a435b51866f58cfe0c8/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/71b2bea/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/ctlplane/endpoints), [swifProxy is up && running](https://logserver.rdoproject.org/88/288/58b7d2f18c2b429f524c0a435b51866f58cfe0c8/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/71b2bea/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/all_resources.log), but I suspect the `container list` command is failing or not returning the expected result. I'm saying that is failing because I see many errors in the [SwiftProxy](https://logserver.rdoproject.org/88/288/58b7d2f18c2b429f524c0a435b51866f58cfe0c8/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/71b2bea/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/swift-proxy-66bc854bc8-lhhmc/logs/proxy-server.log) log. I see both auth errors against keystone and connection timeout, so I think we should hold an environment and let @cschwede look at it.

cescgina:
> @cescgina can we recheck and hold the environment? This failure is weird, I see the [endpoints are properly created](https://logserver.rdoproject.org/88/288/58b7d2f18c2b429f524c0a435b51866f58cfe0c8/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/71b2bea/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/ctlplane/endpoints), [swifProxy is up && running](https://logserver.rdoproject.org/88/288/58b7d2f18c2b429f524c0a435b51866f58cfe0c8/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/71b2bea/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/all_resources.log), but I suspect the `container list` command is failing or not returning the expected result. I'm saying that is failing because I see many errors in the [SwiftProxy](https://logserver.rdoproject.org/88/288/58b7d2f18c2b429f524c0a435b51866f58cfe0c8/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/71b2bea/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/swift-proxy-66bc854bc8-lhhmc/logs/proxy-server.log) log. I see both auth errors against keystone and connection timeout, so I think we should hold an environment and let @cschwede look at it.

@fmount  created the autohold just now, when the next run finishes we should be able to access the environment

cescgina:
recheck added swift_password to rdo-jobs secret file

cescgina:
recheck adding storage_mgmt network

docs_user/modules/openstack-glance_adoption.adoc
```diff
@@ -73,6 +73,95 @@ spec:
...
+              internal:
+                metadata:
+                  annotations:
+                    metallb.universe.tf/address-pool: internalapi
+                    metallb.universe.tf/allow-shared-ip: internalapi
+                    metallb.universe.tf/loadBalancerIPs: 172.17.0.80
+              spec:
+                type: LoadBalancer
+          networkAttachments:
+          - storage
```
konan-abhi, line 140: I think this should be have additional 2 space indent to be inside of 'networkAttachments' section

fmount: Makes sense, thanks: I already fixed it in the tests, but I didn't update the doc reference! Thank you, did it in my last update.

cescgina:
recheck

zuul.d/jobs.yaml
```diff
@@ -6,3 +6,14 @@
...
       # This var is specific to ci-framework docs directory using sphinx
       doc_available: false
+
+- job:
+    name: data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph
+    parent: data-plane-adoption-osp-17-to-extracted-crc
+    vars:
+      use_ceph: "false"
+      dpa_test_suite: "test-minimal"
+    # TODO jgilaber: decide on when we should run this variant
```
jistr, line 16: I think we'd want this job also for Glance and Nova changes? IIRC Compute team wanted a non-Ceph job too. cc @SeanMooney @gibizer @bogdando

jistr: (Anyway i think that can be added in a subsequent PR.)

SeanMooney: yes i would generally say that we should run both the ceph and non ceph version on all change unless there is a strong reason not too.

i would propose one adoption job using ceph for nova/glance/cinder
and one without ceph with glance using the file or swift backend, cinder using lvm or nfs and nova using qcow or raw.

fmount: +1 I agree to have both jobs (w/ and w/o ceph). If no ceph, `Glance` will be configured (by TripleO) w/ `Swift` as a backend in the src cloud. @SeanMooney with this patch you don't need any other change in the `Glance` config (unless you have special needs or you need a particular backends combination).

jistr (approved):

cescgina:
recheck

fmount:
The change itself is good and the glance Pods are up and running.
The CR is properly configured w/ Swift [1], and the job fails during the dataplane adoption (`There are failed AnsibleEE jobs: nova-compute-extraconfig-openstack`).
I'm not sure about the problem on that front [2] but all the ansible jobs fail running the nova-compute container (edpm_container_manage).
@jistr @bogdando do you have any idea to solve [2] ?

[1] https://logserver.rdoproject.org/88/288/65fc4b0f47d09e3093a81253c4f0b3fda4f38e7f/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/57a064f/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/glances.glance.openstack.org/glance.yaml
[2] https://logserver.rdoproject.org/88/288/65fc4b0f47d09e3093a81253c4f0b3fda4f38e7f/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/57a064f/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-mjmrn/logs/openstackansibleee.log

jistr:
Ah sorry this is only about the no-ceph job, so the other job passing is irrelevant (i see it passed here too). Still i think the rebase is worth trying, perhaps after https://github.com/openstack-k8s-operators/data-plane-adoption/pull/338 too.

fmount:
> Ah sorry this is only about the no-ceph job, so the other job passing is irrelevant (i see it passed here too). Still i think the rebase is worth trying, perhaps after #338 too.

ack, I'll rebases after #338 lands!

cescgina:
recheck

cescgina:
recheck

cescgina:
recheck

cescgina:
recheck

cescgina:
@fmount @cschwede as pointed out in https://github.com/openstack-k8s-operators/install_yamls/pull/759#pullrequestreview-1942855955 we should attach the swift pods to the storagemgmt NAD, shouldn't we?

fmount:
> @fmount @cschwede as pointed out in [openstack-k8s-operators/install_yamls#759 (review)](https://github.com/openstack-k8s-operators/install_yamls/pull/759#pullrequestreview-1942855955) we should attach the swift pods to the storagemgmt NAD, shouldn't we?

Under normal circumstances we shouldn't, because pod2pod communication goes through cluster network (`openshift-sdn`), but in this case, where `proxy` pods are in ocp and `swift-storage` is still in the old controllers, we might need to make sure we reach that network. It would probably be a temporary measure, but from the last run I saw that we were blocked in the `nova-compute` adoption, which means `Swift` was happy.
I'll let @cschwede comment to make sure I'm not overlooking important details that influence the whole adoption process.

fmount:
> > > @fmount @cschwede as pointed out in [openstack-k8s-operators/install_yamls#759 (review)](https://github.com/openstack-k8s-operators/install_yamls/pull/759#pullrequestreview-1942855955) we should attach the swift pods to the storagemgmt NAD, shouldn't we?
> > 
> > 
> > Under normal circumstances we shouldn't, because pod2pod communication goes through cluster network (`openshift-sdn`), but in this case, where `proxy` pods are in ocp and `swift-storage` is still in the old controllers, we might need to make sure we reach that network. It would probably be a temporary measure, but from the last run I saw that we were blocked in the `nova-compute` adoption, which means `Swift` was happy. I'll let @cschwede comment to make sure I'm not overlooking important details that influence the whole adoption process.
> 
> ok, so swift adoption is still not complete? it works as it is, so if the current situation is temporary because there is pending swift work then I think we're ok as is and it's not worth to add it, the job passed on the previous run

Right, it works as it is and I would focus on the last failure if swift is happy w/ the current config.
The last failure described here:

```
echo \"There are failed AnsibleEE jobs: $failed\"\n            exit 1\n        fi\n    fi\n\nsleep $DELAY\ndone\n\necho \"Run out of retries\"\nexit 2\n", "delta": "0:22:54.876673", "end": "2024-03-19 04:59:00.417038", "msg": "non-zero return code", "rc": 1, "start": "2024-03-19 04:36:05.540365", "stderr": "+ DEPLOYMENT_NAME=openstack\n+ TRIES=180\n+ DELAY=10\n+ ALLOWED_JOB_RETRIES=3\n++ seq 180\n+ for i in $(seq $TRIES)\n++ oc get osdpd/openstack -o 'jsonpath={.status.conditions[0].status}'\n+ ready=False\n+ '[' False == True ']'\n++ oc get jobs -l osdpd=openstack 
...
```

and I suppose it's about an `ansible task` that fails to deploy nova-compute containers on the EDPM node.

cescgina:
> > > > @fmount @cschwede as pointed out in [openstack-k8s-operators/install_yamls#759 (review)](https://github.com/openstack-k8s-operators/install_yamls/pull/759#pullrequestreview-1942855955) we should attach the swift pods to the storagemgmt NAD, shouldn't we?
> > > 
> > > 
> > > Under normal circumstances we shouldn't, because pod2pod communication goes through cluster network (`openshift-sdn`), but in this case, where `proxy` pods are in ocp and `swift-storage` is still in the old controllers, we might need to make sure we reach that network. It would probably be a temporary measure, but from the last run I saw that we were blocked in the `nova-compute` adoption, which means `Swift` was happy. I'll let @cschwede comment to make sure I'm not overlooking important details that influence the whole adoption process.
> > 
> > 
> > ok, so swift adoption is still not complete? it works as it is, so if the current situation is temporary because there is pending swift work then I think we're ok as is and it's not worth to add it, the job passed on the previous run
> 
> Right, it works as it is and I would focus on the last failure if swift is happy w/ the current config. The last failure described here:
> 
> ```
> echo \"There are failed AnsibleEE jobs: $failed\"\n            exit 1\n        fi\n    fi\n\nsleep $DELAY\ndone\n\necho \"Run out of retries\"\nexit 2\n", "delta": "0:22:54.876673", "end": "2024-03-19 04:59:00.417038", "msg": "non-zero return code", "rc": 1, "start": "2024-03-19 04:36:05.540365", "stderr": "+ DEPLOYMENT_NAME=openstack\n+ TRIES=180\n+ DELAY=10\n+ ALLOWED_JOB_RETRIES=3\n++ seq 180\n+ for i in $(seq $TRIES)\n++ oc get osdpd/openstack -o 'jsonpath={.status.conditions[0].status}'\n+ ready=False\n+ '[' False == True ']'\n++ oc get jobs -l osdpd=openstack 
> ...
> ```
> 
> and I suppose it's about an `ansible task` that fails to deploy nova-compute containers on the EDPM node [1].
> 
> [1] https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/1f012a3/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-whgm9/logs/openstackansibleee.log

yes, we hit this issue every now and then, not sure what causes it, trying a recheck

cescgina:
recheck

cescgina:
recheck

bogdando:
> > > > > @fmount @cschwede as pointed out in [openstack-k8s-operators/install_yamls#759 (review)](https://github.com/openstack-k8s-operators/install_yamls/pull/759#pullrequestreview-1942855955) we should attach the swift pods to the storagemgmt NAD, shouldn't we?
> > > > 
> > > > 
> > > > Under normal circumstances we shouldn't, because pod2pod communication goes through cluster network (`openshift-sdn`), but in this case, where `proxy` pods are in ocp and `swift-storage` is still in the old controllers, we might need to make sure we reach that network. It would probably be a temporary measure, but from the last run I saw that we were blocked in the `nova-compute` adoption, which means `Swift` was happy. I'll let @cschwede comment to make sure I'm not overlooking important details that influence the whole adoption process.
> > > 
> > > 
> > > ok, so swift adoption is still not complete? it works as it is, so if the current situation is temporary because there is pending swift work then I think we're ok as is and it's not worth to add it, the job passed on the previous run
> > 
> > 
> > Right, it works as it is and I would focus on the last failure if swift is happy w/ the current config. The last failure described here:
> > ```
> > echo \"There are failed AnsibleEE jobs: $failed\"\n            exit 1\n        fi\n    fi\n\nsleep $DELAY\ndone\n\necho \"Run out of retries\"\nexit 2\n", "delta": "0:22:54.876673", "end": "2024-03-19 04:59:00.417038", "msg": "non-zero return code", "rc": 1, "start": "2024-03-19 04:36:05.540365", "stderr": "+ DEPLOYMENT_NAME=openstack\n+ TRIES=180\n+ DELAY=10\n+ ALLOWED_JOB_RETRIES=3\n++ seq 180\n+ for i in $(seq $TRIES)\n++ oc get osdpd/openstack -o 'jsonpath={.status.conditions[0].status}'\n+ ready=False\n+ '[' False == True ']'\n++ oc get jobs -l osdpd=openstack 
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and I suppose it's about an `ansible task` that fails to deploy nova-compute containers on the EDPM node [1].
> > [1] https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/1f012a3/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-whgm9/logs/openstackansibleee.log
> 
> yes, we hit this issue every now and then, not sure what causes it, trying a recheck

so, this is intermittent?

@gibi FYI, see in [logs](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/1f012a3/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-whgm9/logs/openstackansibleee.log)
```

TASK [osp.edpm.edpm_container_manage : Create containers managed by Podman for /var/lib/openstack/config/containers] ***
Tuesday 19 March 2024  08:53:44 +0000 (0:00:00.134)       0:02:02.891 ********* 
[WARNING]: ERROR: Container nova_compute exited with code 125 when started
stderr: Error: unable to start container
"21ebca1207e73109a22eca189a8744e10f44d526db53461e21d567ebb7c6f2d6": lsetxattr
/var/lib/nova/compute_id: operation not permitted
fatal: [standalone]: FAILED! => {"changed": false, "msg": "Failed containers: nova_compute"}
```

cescgina:
> > > > > > @fmount @cschwede as pointed out in [openstack-k8s-operators/install_yamls#759 (review)](https://github.com/openstack-k8s-operators/install_yamls/pull/759#pullrequestreview-1942855955) we should attach the swift pods to the storagemgmt NAD, shouldn't we?
> > > > > 
> > > > > 
> > > > > Under normal circumstances we shouldn't, because pod2pod communication goes through cluster network (`openshift-sdn`), but in this case, where `proxy` pods are in ocp and `swift-storage` is still in the old controllers, we might need to make sure we reach that network. It would probably be a temporary measure, but from the last run I saw that we were blocked in the `nova-compute` adoption, which means `Swift` was happy. I'll let @cschwede comment to make sure I'm not overlooking important details that influence the whole adoption process.
> > > > 
> > > > 
> > > > ok, so swift adoption is still not complete? it works as it is, so if the current situation is temporary because there is pending swift work then I think we're ok as is and it's not worth to add it, the job passed on the previous run
> > > 
> > > 
> > > Right, it works as it is and I would focus on the last failure if swift is happy w/ the current config. The last failure described here:
> > > ```
> > > echo \"There are failed AnsibleEE jobs: $failed\"\n            exit 1\n        fi\n    fi\n\nsleep $DELAY\ndone\n\necho \"Run out of retries\"\nexit 2\n", "delta": "0:22:54.876673", "end": "2024-03-19 04:59:00.417038", "msg": "non-zero return code", "rc": 1, "start": "2024-03-19 04:36:05.540365", "stderr": "+ DEPLOYMENT_NAME=openstack\n+ TRIES=180\n+ DELAY=10\n+ ALLOWED_JOB_RETRIES=3\n++ seq 180\n+ for i in $(seq $TRIES)\n++ oc get osdpd/openstack -o 'jsonpath={.status.conditions[0].status}'\n+ ready=False\n+ '[' False == True ']'\n++ oc get jobs -l osdpd=openstack 
> > > ...
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > and I suppose it's about an `ansible task` that fails to deploy nova-compute containers on the EDPM node [1].
> > > [1] https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/1f012a3/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-whgm9/logs/openstackansibleee.log
> > 
> > 
> > yes, we hit this issue every now and then, not sure what causes it, trying a recheck
> 
> so, this is intermittent?
> 
> @gibi FYI, see in [logs](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/1f012a3/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-whgm9/logs/openstackansibleee.log)
> 
> ```
> 
> TASK [osp.edpm.edpm_container_manage : Create containers managed by Podman for /var/lib/openstack/config/containers] ***
> Tuesday 19 March 2024  08:53:44 +0000 (0:00:00.134)       0:02:02.891 ********* 
> [WARNING]: ERROR: Container nova_compute exited with code 125 when started
> stderr: Error: unable to start container
> "21ebca1207e73109a22eca189a8744e10f44d526db53461e21d567ebb7c6f2d6": lsetxattr
> /var/lib/nova/compute_id: operation not permitted
> fatal: [standalone]: FAILED! => {"changed": false, "msg": "Failed containers: nova_compute"}
> ```

it seems so, see for example the second-to-last run currently where the no-ceph jobs passed without any change: https://review.rdoproject.org/zuul/build/c8fd9e024f2f462baff151eb40bcdb53, but the job with ceph that before was working failed with the same issue https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc/c894eb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-29cls/logs/openstackansibleee.log

bogdando:
> it seems so, see for example the second-to-last run currently where the no-ceph jobs passed without any change: https://review.rdoproject.org/zuul/build/c8fd9e024f2f462baff151eb40bcdb53, but the job with ceph that before was working failed with the same issue https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc/c894eb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-29cls/logs/openstackansibleee.log

this could be that racy code perhaps https://github.com/openstack-k8s-operators/edpm-ansible/blob/3c55d2af85a4b8c2c0ad5fdd46cb89f846a3cca1/roles/edpm_nova/templates/nova_compute.json.j2#L22

bogdando:
> passed without any change: https://review.rdoproject.org/zuul/build/c8fd9e024f2f462baff151eb40bcdb53

note that 1st exec of the job has also [failed](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/c8fd9e0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-rv9rb/logs/openstackansibleee.log) the same way, then it worked a [2 min later](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc/c894eb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-29cls/logs/openstackansibleee.log)

so we can say this logic fails consistently, but sometimes retrying it helps... @gibi

cescgina:
recheck

cescgina:
recheck

cescgina:
recheck

gibizer:
> > passed without any change: https://review.rdoproject.org/zuul/build/c8fd9e024f2f462baff151eb40bcdb53
> 
> note that 1st exec of the job has also [failed](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/c8fd9e0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-rv9rb/logs/openstackansibleee.log) the same way, then it worked a [13 min later](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc/c894eb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-29cls/logs/openstackansibleee.log)
> 
> so we can say this logic fails consistently, but sometimes retrying it helps... @gibi

On a greenfield deployment I see a bit different error that is ignored and the deployment is healthy:
```
INFO:__main__:Setting permission for /var/lib/nova/compute_id
ERROR:__main__:Failed to change ownership of /var/lib/nova/compute_id to 42436:42436
Traceback (most recent call last):
  File "/usr/local/bin/kolla_set_configs", line 343, in set_perms
    os.chown(path, uid, gid)
PermissionError: [Errno 1] Operation not permitted: '/var/lib/nova/compute_id'
INFO:__main__:Setting permission for /var/lib/nova/.ssh/ssh-privatekey
```

@bogdando the link to the successful attempt seems to be wrong as it also points to a log where the container failed to start.

gibizer:
> > passed without any change: https://review.rdoproject.org/zuul/build/c8fd9e024f2f462baff151eb40bcdb53
> 
> note that 1st exec of the job has also [failed](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/c8fd9e0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-rv9rb/logs/openstackansibleee.log) the same way, then it worked a [13 min later](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc/c894eb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-29cls/logs/openstackansibleee.log)
> 
> so we can say this logic fails consistently, but sometimes retrying it helps... @gibi

On a greenfield deployment I see a bit different error that is ignored and the deployment is healthy:
```
INFO:__main__:Setting permission for /var/lib/nova/compute_id
ERROR:__main__:Failed to change ownership of /var/lib/nova/compute_id to 42436:42436
Traceback (most recent call last):
  File "/usr/local/bin/kolla_set_configs", line 343, in set_perms
    os.chown(path, uid, gid)
PermissionError: [Errno 1] Operation not permitted: '/var/lib/nova/compute_id'
INFO:__main__:Setting permission for /var/lib/nova/.ssh/ssh-privatekey
```

@bogdando the link to the successful attempt seems to be wrong as it also points to a log where the container failed to start.

> > > passed without any change: https://review.rdoproject.org/zuul/build/c8fd9e024f2f462baff151eb40bcdb53
> > 
> > 
> > note that 1st exec of the job has also [failed](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph/c8fd9e0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-rv9rb/logs/openstackansibleee.log) the same way, then it worked a [13 min later](https://logserver.rdoproject.org/88/288/d205e0f930b4c006e94ba74d491ad3b527616b09/github-check/data-plane-adoption-osp-17-to-extracted-crc/c894eb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pods/nova-compute-extraconfig-openstack-29cls/logs/openstackansibleee.log)
> > so we can say this logic fails consistently, but sometimes retrying it helps... @gibi
> 
> On a greenfield deployment I see a bit different error that is ignored and the deployment is healthy:
> 
> ```
> INFO:__main__:Setting permission for /var/lib/nova/compute_id
> ERROR:__main__:Failed to change ownership of /var/lib/nova/compute_id to 42436:42436
> Traceback (most recent call last):
>   File "/usr/local/bin/kolla_set_configs", line 343, in set_perms
>     os.chown(path, uid, gid)
> PermissionError: [Errno 1] Operation not permitted: '/var/lib/nova/compute_id'
> INFO:__main__:Setting permission for /var/lib/nova/.ssh/ssh-privatekey
> ```
> 
> @bogdando the link to the successful attempt seems to be wrong as it also points to a log where the container failed to start.

I will try to troubleshoot the issue separately in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/370

cescgina:
recheck

cescgina:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

cescgina:
recheck

fmount:
recheck

fmount:
@cescgina an interesting number of rechecks to get CI green: is there a particular known issue?

cescgina:
> @cescgina an interesting number of rechecks to get CI green: is there a particular known issue?

yes, sorry, there are a number of problems going on at once. I was using this to test changes, but I moved now to use a testproject to reduce the noise

fmount:
> > @cescgina an interesting number of rechecks to get CI green: is there a particular known issue?
> 
> yes, sorry, there are a number of problems going on at once. I was using this to test changes, but I moved now to use a testproject to reduce the noise

np, looks like all the dependencies merged, I think this change is good to go as long as the CI is green, or do you have concerns except a green CI?

cescgina:
> > > @cescgina an interesting number of rechecks to get CI green: is there a particular known issue?
> > 
> > 
> > yes, sorry, there are a number of problems going on at once. I was using this to test changes, but I moved now to use a testproject to reduce the noise
> 
> np, looks like all the dependencies merged, I think this change is good to go as long as the CI is green, or do you have concerns except a green CI?

I agree, I think it's good to go, we even had some green runs earlier this week, hopefully we can get CI back on track soon and this can be merged

fmount:
recheck

fmount:
@cescgina @jistr finally green CI. All the dependencies are merged, is it ok to approve and land this patch?

cescgina (approved):

marios (approved):

PR #292, opened 2024-02-19
ospdo adoption

pinikomarov:
Add Director Operator scripts for adoption
    
    Jiras:


pinikomarov:
need to encript ssh key to odpdo BM

tests/playbooks/test_minimal_ospdo.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+
+- name: Adoption
+  hosts: local
+  gather_facts: false
+  module_defaults:
+    ansible.builtin.shell:
+      executable: /bin/bash
+  tasks:
+    - name: set ospdo_src var
```
archana203, line 27: If all the ospdo specific task you have added in service roles are conditional, then the same test-minimal test suite could be used by defining conditional variable ospdo_src in vars.yaml.

pinikomarov: that's a good point , I'll change accordingly

Makefile
```diff
@@ -9,7 +9,12 @@ TEST_CONFIG ?= tests/ansible.cfg
 test-minimal: TEST_OUTFILE := tests/logs/test_minimal_out_$(shell date +%FT%T%Z).log
 test-minimal:
 	mkdir -p tests/logs
-	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
+	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e ospdo_src=True -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
```
ciecierski, line 12: Should it be ospdo_src=False as it is under minimal?

docs/openstack/ospdo_adoption.md
```diff
@@ -0,0 +1,5 @@
+# OSPdO adoptinon
+
+Openstack Director Operator adoption, branches of the tripleo adoption
```
ciecierski, line 3: s/tripleo/TripleO

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -11,6 +11,20 @@
     cd {{ install_yamls_path }}
     make input
 
+- name: get service passwords
+  delegate_to: "{{ adoption_ospdo_host }}"
+  ansible.builtin.shell: |
+     oc get secret tripleo-passwords -o jsonpath='{.data.*}' | base64 -d>~/tripleo-standalone-passwords.yaml
+  when: ospdo_src is defined
```
ciecierski, line 18: It is defined always. Both in minimal make target(false) and in ospdo target. It would be better to use :
when: ospdo_src| default(false) | bool

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -11,6 +11,20 @@
...
+
+- name: Fetch file from delegate host
+  delegate_to: "{{ adoption_ospdo_host }}"
+  fetch:
+    src: ~/tripleo-standalone-passwords.yaml
+    dest: ~/tripleo-standalone-passwords.yaml
+    flat: yes
+  when: ospdo_src is defined
```
ciecierski, line 26: ditto ^

tests/roles/ceph_backend_configuration/tasks/main.yaml
ciecierski, line 1: same comment regarding ospdo_src as in tests/roles/backend_services/tasks/main.yaml

tests/roles/stop_openstack_services/templates/systemctl_disable.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Try to stop a service , fail only when the service is present and doesn't stop correctly.
```
ciecierski, line 2: This comment seems to be copy-paste for systemctl_stop.

ciecierski (commented):
Please check comments about when: ospdo_src is defined

Makefile
pinikomarov: that should be removed , no need to add the variable as false as the tag is checking if it's defined , the bool check is not a must, and I didnt't want to add unnecessary code to the regular test-minimal.

docs/openstack/ospdo_adoption.md
pinikomarov: ack

tests/roles/backend_services/tasks/main.yaml
pinikomarov: I removed the definition from tes-minimal , so this would work as is

tests/roles/backend_services/tasks/main.yaml
pinikomarov: same comment as above

tests/roles/stop_openstack_services/templates/systemctl_disable.sh
pinikomarov: ack , updated he comment

tests/roles/ceph_backend_configuration/tasks/main.yaml
pinikomarov: I removed the definition from tes-minimal , so this would work as is

tests/roles/stop_openstack_services/tasks/main.yaml
```diff
@@ -1,46 +1,56 @@
...
-                    "tripleo_nova_api.service"
-                    "tripleo_nova_api_cron.service"
-                    "tripleo_nova_api.service"
-                    "tripleo_nova_conductor.service"
-                    "tripleo_nova_metadata.service"
-                    "tripleo_nova_scheduler.service"
-                    "tripleo_nova_vnc_proxy.service"
-                    "tripleo_ovn_cluster_northd.service"
-                    "tripleo_placement_api.service")
+    ServicesToStop=({{ services_to_stop | join(' ') }})
```
jistr, line 48: This should use the shell vars as it was, the tests have to run what's written in the documentation. We're not aiming for "prettiest Ansible", we're aiming for a test suite that verifies what's written in the docs and will be executed by admins, which often means having to use shell constructs instead of Ansible constructs in the test suite.

tests/roles/stop_openstack_services/templates/systemctl_disable.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Disable and mask the service on the Computes.
+service=$1
+
+cat << EOF > ~/disable_service.yaml
```
jistr, line 5: This seems to be getting quite far from what's written in the docs... If there are OSPDO specifics that make us go through extra steps / indirection, then that's fine but they should probably be added to the docs too.

Can we just use a SSH command to stop the services like we're doing for the other environment? We don't assume that the end user will have Ansible at their disposal.

tests/roles/stop_openstack_services/templates/systemctl_stop.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# Try to stop a service , fail only when the service is present and doesn't stop correctly.
+service=$1
+
+cat << EOF > ~/stop_service.yaml
```
jistr, line 5: Ditto re getting far from the docs.

tests/roles/stop_remaining_services/tasks/main.yaml
```diff
@@ -11,32 +11,40 @@
...
+      - "haproxy-bundle"
+      - "rabbitmq-bundle"
+
 - name: stop compute services
   no_log: "{{ use_no_log }}"
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ stop_other_services_shell_vars }}
 
-    ComputeServicesToStop=(
```
jistr, line 20: Same comment here as in the controller services stopping -- we need to execute what's written in the docs.

tests/secrets_ospdo.yaml
```diff
@@ -0,0 +1,61 @@
+# Command to login to OpenShift
+oc_login_command: |
+  oc login -u kubeadmin -p {{ admin_password }}
+
+tripleo_passwords: ~/tripleo-standalone-passwords.yaml #CUSTOMIZE_THIS
+
+# Adopted OSPD director operator env Baremetal fqdn
+adoption_ospdo_host: titan_xx.lab.eng.tlv2.redhat.com
```
jistr, line 8: I wouldn't put values in here that are as env-specific as a hostname. But keeping the OSPDO example file is good i think.

jistr (changes requested):

jistr:
There's another option that we could do -- given how many differences there are in some of the roles between OSPD vs. OSPDO, maybe we could have whole task files for the OSPDO separate from the task files for OSPD? This would certainly bring duplication, but it would make it easier to read for a human.

ciecierski:
> There's another option that we could do -- given how many differences there are in some of the roles between OSPD vs. OSPDO, maybe we could have whole task files for the OSPDO separate from the task files for OSPD? This would certainly bring duplication, but it would make it easier to read for a human.

+1 on separate task file for OSPD. Better readability of code and logs.

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -11,19 +11,9 @@
...
-- name: Fetch file from delegate host
-  delegate_to: "{{ adoption_ospdo_host }}"
-  fetch:
-    src: ~/tripleo-standalone-passwords.yaml
-    dest: ~/tripleo-standalone-passwords.yaml
-    flat: yes
-  when: ospdo_src is defined
+- name: execute alternative tasks when source env is ODPdO
+  ansible.builtin.include_tasks: ospdo_backend_services.yaml
+  when: ospdo_src| default(false) | bool
```
marios, line 16: might be nice to have this in the vars.sample.yaml with a note about what it does?
I you wanted to you could default to false there and not have to worry about using the | default(false) in all the places too.

tests/roles/ovn_adoption/tasks/ospdo_ovn_adoption.yaml
```diff
@@ -0,0 +1,37 @@
+- name: get OSPdO OVN ovsdb-server IP
+  delegate_to: "{{ adoption_ospdo_host }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    oc get cm -o yaml|grep ovn_dbs_node_ips|tr -d ' '|sed 's/.*ovn_dbs_node_ips:\\n-//g'|sed 's/\\n.*//g'
+  register: source_ovndb_ip
+  when: get_source_ovndb_ip| default(false) | bool
+
+- name: Fetch OVN databases, cp into helper pod
+  block:
```
marios, line 10: you guard the previous task with "get_source_ovndb_ip" on line 7

so if that is set to false, then this following block will fail becuase source_ovndb_ip will be undefined.

so maybe a conditional on this block?

when: source_ovndb_ip is defined?
or even simpler put the previous task inside the block and just have the same condition as above but for the block:

- name: Fetch OVN databases, cp into helper pod
   when: get_source_ovndb_ip| default(false) | bool
   block:

tests/roles/keystone_adoption/tasks/ospdo_keystone_adoption.yaml
```diff
@@ -0,0 +1,31 @@
...
+  fetch:
+    src: ~/oc_apply_fer.yaml
+    dest: ~/oc_apply_fer.yaml
+    flat: yes
+
+- name: OC apply keystone fernet keys secret file
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc apply -f ~/oc_apply_fer.yaml
```
marios, line 31: i think this will produce a linting error when you put this into ready for review (currently in draft state)
i.e. you'll need to add a newline here

tests/roles/ovn_adoption/tasks/ospdo_ovn_adoption.yaml
pinikomarov: the   register: source_ovndb_ip task is used in the next task:
- name: set OVN copy shell vars , and that creates a variable that is used in the 
- name: start an adoption helper pod task that later is used lastly by my the - name: Stop Fetch OVN databases, cp into helper pod
block . So I to not interrupt the original flow it seems to me the best approach of safeguarding this is by: adding an and source_ovndb_ip is defined in the block's when list.

tests/roles/keystone_adoption/tasks/ospdo_keystone_adoption.yaml
pinikomarov: done

tests/roles/backend_services/tasks/main.yaml
pinikomarov: done

tests/roles/backend_services/tasks/ospdo_backend_services.yaml
```diff
@@ -0,0 +1,11 @@
+- name: get service passwords
+  delegate_to: "{{ adoption_ospdo_host }}"
+  ansible.builtin.shell: |
+     oc get secret tripleo-passwords -o jsonpath='{.data.*}' | base64 -d>~/tripleo-standalone-passwords.yaml
```
holser, line 4: I guess we can flat yaml here, though I don't insist

tests/roles/backend_services/tasks/ospdo_backend_services.yaml
```diff
@@ -0,0 +1,11 @@
+- name: get service passwords
+  delegate_to: "{{ adoption_ospdo_host }}"
```
holser, line 2: I guess we can avoid most of delegate_to by creating tasks for specific hosts.
```
---
- name: get service passwords
  hosts: adoption_ospdo_host
  tasks:
  ***
  roles:
  ***
```

tests/roles/mariadb_copy/templates/get_dbs_ospdo.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -x
+galera_con=$(sudo podman ps --format="{{.Names}}"|grep galera)
+mariadb_templates='/home/cloud-admin/data-plane-adoption_pkomarov/tests/roles/mariadb_copy/templates/'
+
+sudo podman cp ${mariadb_templates}/dump_dbs_ospdo.sh $galera_con:/
+sudo podman exec -it $galera_con bash -c "/dump_dbs_ospdo.sh"
+sudo podman cp  ${galera_con}:/tmp/mariadb /tmp
```
holser, line 8: we do not run podman on operator's machine anymore. All tasks are executed on destionation. Please see dump_dbs.bash of mariadb_copy role

holser: Why not to reuse  mariadb_copy role?

tests/roles/ovn_adoption/tasks/ospdo_ovn_adoption.yaml
```diff
@@ -0,0 +1,51 @@
...
+      $CONTROLLER2_SSH sudo systemctl stop tripleo_ovn_cluster_northd.service
+      $CONTROLLER3_SSH sudo systemctl stop tripleo_ovn_cluster_northd.service
+
+  - name: dump OVN databases
+    delegate_to: "{{ adoption_ospdo_host }}"
+    ansible.builtin.shell: |
+      {{ exec_ospdoclnt_pod }} ovsdb-client  backup tcp:{{ source_ovndb_ip.stdout }}:{{ item }}
+    with_items:
+    - 6641 >/tmp/ovs-nb.db
+    - 6642 >/tmp/ovs-sb.db
```
holser, line 29: That was refactored to do dump to destination cloud

tests/roles/ceph_backend_configuration/tasks/ospdo_ceph_backend_configuration.yaml
```diff
@@ -0,0 +1,28 @@
...
+- name: update the openstack keyring caps for Manila
+  delegate_to: "{{ adoption_ospdo_host }}"
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    CEPH_SSH="{{ controller1_ssh }}"
+    CEPH_CAPS="mgr 'allow *' mon 'allow r, profile rbd' osd 'profile rbd pool=vms, profile rbd pool=volumes, profile rbd pool=images, allow rw pool manila_data'"
+    OSP_KEYRING="client.openstack"
+    CEPH_ADM=$($CEPH_SSH "cephadm shell -- ceph auth caps $OSP_KEYRING $CEPH_CAPS")
```
jistr, line 28: Why can't we use the existing code for this?

tests/roles/keystone_adoption/tasks/ospdo_keystone_adoption.yaml
jistr, line 31: Why can't we use the existing code for this?

tests/roles/mariadb_copy/templates/get_dbs_ospdo.sh
jistr: +1, the shell scripts in the tests have no representation in the docs.

jistr (commented):
Thanks for the refactor to separate out the special tasks. I wonder if we can bring it closer to the docs in some places, so that we can insert small bits of "if your env is OSPDO, execute this snippet instead to get the shell variables, but the rest of the procedure is the same".

My overarching concern is that if we merge this, we're enabling testing for OSPDO but it's all specific to the testing environment. The admins who have OSDPDO are still left in the dark, there is nothing in the documentation that reflects the forks in logic. And ideally, rather than start forking docs too, i wonder if we can bring the OSPDO testing closer to the existing docs. At least from my unserstanding, it looks like in a bunch of places the departure from the existing procedure is greater than it needs to be?

pinikomarov:
Thanks guys , I'll start refactoring my changes to be as close to the existing methods as possible

tests/vars.sample.yaml
```diff
@@ -29,3 +29,7 @@ timesync_ntp_servers:
 
 # Auth URL to use for adopted Keystone.
 auth_url: http://keystone-public-openstack.apps-crc.testing
+
+# Whether source env is OSPD Director Operator
+# i.e. https://github.com/openstack-k8s-operators/osp-director-operator
+ospdo_src: false
```
marios, line 35: theres a couple of those it means you need to add a newline

marios (changes requested):

tests/vars.sample.yaml
pinikomarov: ack

pinikomarov:
Hi guys , can you take a look at the mariadb_copy refactor I did , 
please share if this pattern is ok , 
it's using the mariadb-copy-data pod locally and remotely (on src env) in the same way. 
so that the only difference in action is where the pod runs, then data is transferred.
meaning : run remote mariadb-copy-data pod on the ospdo BM env, copy data to local standalone  mariadb-copy-data pod (crc node) then continue as usual.

https://github.com/openstack-k8s-operators/data-plane-adoption/pull/292/files#diff-d5dba85fef50c60b9a5a0ead4321eea3518b41e38cb65892376198a8cd23c1c8
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/292/files#diff-0250193aeb7a82237a8daafe84e96fbfd0cac9b57885e4a06c9fddfbf6faab5a

jistr (changes requested):
Thank you for the amendments, i think indeed the MariaDB parts are closer to what we should be doing by using the existing bash scripts for dump & restore. I still think there are some parts which need amending, but mainly, i think this is difficult to review because the PR is based on an old commit, it seems to miss more than a month of `main` branch progress. So some things that show up in the diff as added by this PR are already in the main branch. Can you please rebase so that we can see the actual extent of the changes?

tests/roles/keystone_adoption/tasks/ospdo_keystone_adoption.yaml
pinikomarov: I had some issues with using "-" since I'm passing commands through oc exec and ssh to controller, I'll check again though if the original way can be done here

tests/roles/mariadb_copy/tasks/ospdo_mariadb_copy.yaml
```diff
@@ -0,0 +1,144 @@
...
+        MARIADB_IMAGE=quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
+        STORAGE_CLASS=host-nfs-storageclass
+        SOURCE_MARIADB_IP={{ source_mariadb_ip.stdout }}
+        declare -A SOURCE_GALERA_MEMBERS
+        SOURCE_GALERA_MEMBERS=(
+        ["standalone.localdomain"]={{ source_mariadb_ip.stdout }}
+        )
+        SOURCE_DB_ROOT_PASSWORD={{ source_db_root_pass.stdout }}
+
+  - name: start an adoption mariadb helper pod on ospdo src env
```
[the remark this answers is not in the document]
pinikomarov: this pod is on the ospdo env , startred there also to get access to mariadb DBs, because the local maria_helper can't reach the mariadb ip inside the ospdo env

tests/roles/mariadb_copy/tasks/main.yaml
```diff
@@ -6,6 +6,67 @@
   ansible.builtin.include_tasks:
     file: env_vars_dst.yaml
 
+- name: start an adoption mariadb helper pod
```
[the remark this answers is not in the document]
pinikomarov: done , new patch is rebased

tests/roles/mariadb_copy/tasks/main.yaml
```diff
@@ -38,6 +106,13 @@
   no_log: "{{ use_no_log }}"
   ansible.builtin.shell:
     cmd: "{{ lookup('ansible.builtin.template', 'dump_dbs.bash') }}"
+  when: not ospdo_src| bool
+
+- name: execute alternative tasks when source env is ODPdO
+  ansible.builtin.include_tasks: ospdo_mariadb_copy.yaml
+  when: ospdo_src| bool
+  vars:
+    ospdo_get_dbs: True
```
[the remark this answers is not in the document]
pinikomarov: done

tests/roles/mariadb_copy/templates/exec_mysql_ospdo.sh
```diff
@@ -0,0 +1,61 @@
...
+  "nova_host_names")
+    nova_host_names
+    ;;
+  "list_nova_cells")
+    list_nova_cells
+    ;;
+  "check_wsrep")
+    check_wsrep
+    ;;
+esac
```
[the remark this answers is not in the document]
pinikomarov: done

pinikomarov:
continued here : https://github.com/openstack-k8s-operators/data-plane-adoption/pull/444

PR #294, opened 2024-02-19
[DP3] backport downstream docs variant rendering

jistr:


jistr:
Ah wrong target branch :)

PR #296, opened 2024-02-20
[ovn_adoption] start northd/ovn-controller after ovndbs adoption

karelyatin:
ovn-controller and northd were started along with ovn dbs, instead these should be started after ovn dbs are adopted.

Currently enable ovnnorthd task was noop as it was already started.

Also remove references for ovs CR which no longer exists since [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/315

Depends-On: https://github.com/openstack-k8s-operators/ovn-operator/pull/232
Related-Issue: OSPRH-2437

### DIFF
--- a/docs_user/modules/openstack-backend_services_deployment.adoc
+++ b/docs_user/modules/openstack-backend_services_deployment.adoc
@@ -232,9 +232,11 @@ spec:
           networkAttachment: internalapi
       ovnNorthd:
         networkAttachment: internalapi
-        replicas: 1
+        replicas: 0
       ovnController:
         networkAttachment: tenant
+        nodeSelector:
+          node: non-existing-node-name
 
   placement:
     enabled: false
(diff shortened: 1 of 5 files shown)


karelyatin:
recheck Perform container registry login(s) with podman

slawqo (approved):

karelyatin:
we would also needs the dependent patch in https://github.com/openstack-k8s-operators/ovn-operator/pull/232 , without that i think dataplane tests will fail

karelyatin:
> we would also needs the dependent patch in [openstack-k8s-operators/ovn-operator#232](https://github.com/openstack-k8s-operators/ovn-operator/pull/232) , without that i think dataplane tests will fail

or may be not as doesn't seem check ovn, oscp CR status in tests. The other PR already merged

PR #298, opened 2024-02-21
Update docs about how to render docs locally

jistr:


jistr:
I'll go ahead and merge it so that i can link the rendered version in the upcoming call :)

PR #299, opened 2024-02-22
Adding a navbar to the docs

fao89:
image

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -49,15 +49,61 @@ jobs:
       - name: Create index.html
         run: |
           cat > index.html <<EOF
+          <!DOCTYPE html>
           <html>
+          <head>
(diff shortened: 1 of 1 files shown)


jistr:
Hmm now i wonder if rendering the docs in the iframe would allow us to pass around direct links to particular headings inside the docs. I guess it would, they just wouldn't open with the top navbar... Which is perhaps ok :)

.github/workflows/docs.yaml
```diff
@@ -49,15 +49,61 @@ jobs:
...
+                  <a href="dev/" target="content">Contributor</a>
+              </div>
+
+              <div class="asciidoc">
+                  <iframe name="content" width="100%" height="100%" src="user/index.html"></iframe>
+              </div>
+
           </body>
-          <html>
+          </html>
```
jistr, line 106: :smile_cat:

jistr (approved):

PR #302, opened 2024-02-23
Introduce MDS migration guide

fmount:
This patch represents the third chapter of the Ceph cluster daemons migration. It describes how to properly move `MDS` daemons on the target nodes (either `CephStorage` or `Computes` in case of `HCI`).

Implements: [OSPRH-4850](https://issues.redhat.com/browse/OSPRH-4850)


jistr (approved):
LGTM, thanks :+1:

jistr (approved):

PR #304, opened 2024-02-26
sudo rights required to access swift.conf file

archana203:
When ssh command is using user other than root, sudo rights required to access the swift.conf and feeding data in swift secret.

### DIFF
--- a/docs_user/modules/openstack-swift_adoption.adoc
+++ b/docs_user/modules/openstack-swift_adoption.adoc
@@ -37,7 +37,7 @@ metadata:
   namespace: openstack
 type: Opaque
 data:
-  swift.conf: $($CONTROLLER1_SSH cat /var/lib/config-data/puppet-generated/swift/etc/swift/swift.conf | base64 -w0)
+  swift.conf: $($CONTROLLER1_SSH sudo cat /var/lib/config-data/puppet-generated/swift/etc/swift/swift.conf | base64 -w0)
 EOF
 ----
 
--- a/tests/roles/swift_adoption/tasks/main.yaml
+++ b/tests/roles/swift_adoption/tasks/main.yaml
@@ -11,7 +11,7 @@
       namespace: openstack
     type: Opaque
     data:
-      swift.conf: $($CONTROLLER1_SSH cat /var/lib/config-data/puppet-generated/swift/etc/swift/swift.conf | base64 -w0)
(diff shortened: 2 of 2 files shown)


cschwede (changes requested):
The doc itself should be updated as well here: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/openstack-swift_adoption.adoc?plain=1#L40

And we might need to add sudo to https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/swift_adoption/tasks/main.yaml#L28C1-L28C156 and https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/openstack-swift_adoption.adoc?plain=1#L40 respectively

archana203:
> The doc itself should be updated as well here: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/openstack-swift_adoption.adoc?plain=1#L40
Done
> And we might need to add sudo to https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/swift_adoption/tasks/main.yaml#L28C1-L28C156 
- We do not need it here as the command executes with no permission issue using stack user as well.

cschwede (approved):
/lgtm

jistr (approved):

fmount (approved):
/lgtm

archana203:
@jistr Do we also need to add the swift service with default: false in base openstackcontrolplane template https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/config/base/openstack_control_plane.yaml

fmount:
also @archana203 do we have a way to test https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288 on top of the current change? When we don't deploy Ceph and don't pass any backend to Glance in a TripleO based environment, swift is the default value used. As long as we perform this first phase of `Swift` adoption, we might want to increase glance replicas and adopt that service using `glance_backend: swift`.
Local storage is not a real use case, and I'm wondering how the environment you test is built.

jistr:
@fmount @archana203 This is OT from the pull request perspective but perhaps we could amend the test_minimal.yaml scenario to deploy Glance with Swift backend rather than local backend?

jistr:
We don't run it anywhere in pre-merge CI currently, but if we do have the resources for it, perhaps we could...

fmount:
> @fmount @archana203 This is OT from the pull request perspective but perhaps we could amend the test_minimal.yaml scenario to deploy Glance with Swift backend rather than local backend?

Yeah that makes sense to me.

PR #312, opened 2024-02-28
Patch glanceAPI using per-backend files

fmount:
Instead of passing an inline string to patch the controlplane, rely on
per-backend files (because the config is pretty much static) that can
be called from the role. This allows to improve readability and it helps
in terms of maintenance, and it also avoid conflicts in case we use
golang templating within the customServiceConfig option.

Related: [OSPRH-784](https://issues.redhat.com/browse/OSPRH-784)

### DIFF
--- a/tests/roles/glance_adoption/files/glance_ceph.yaml
+++ b/tests/roles/glance_adoption/files/glance_ceph.yaml
@@ -1,0 +1,32 @@
+spec:
+  glance:
+    enabled: true
+    template:
+      databaseInstance: openstack
+      customServiceConfig: |
+        [DEFAULT]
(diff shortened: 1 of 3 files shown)


fmount:
Nice, the ceph job basically validated this patch and we can rely on this new mechanism when we patch the ctlplane glance files.
@jistr I'm going to propose the same thing for the existing backends, then the swift patch https://github.com/openstack-k8s-operators/data-plane-adoption/pull/288 can be rebased on top of this (and we can avoid the golang vs jinja templating conflict).

fao89 (approved):

jistr (approved):

PR #286, opened 2024-02-14
Downstream docs variant

jistr:
Downstream docs preview can be built with:

```
BUILD=downstream make docs-user
```

(and analogously for the `docs-user-open` and `docs-user-watch` make targets).

PR #315, opened 2024-03-01
Cinder adoption fixes

Akrog:
This PR does 2 things related to the cinder adoption:

- Re-adds cinder-cfg.py script: On commit fba8ae0129655795d5481a1434265a2bc29d8342 the cinder-cfg.py script was removed when cleaning up old Markdown/MkDocs files, but this file is still being referenced in the documentation, so we'll get a 404 error when trying to follow the link in the rendered docs.

- Explicitly mentions what needs to happen with the `[barbican]` section of the old `cinder.conf`.

### DIFF
--- a/docs_user/modules/cinder-cfg.py
+++ b/docs_user/modules/cinder-cfg.py
@@ -1,0 +1,765 @@
+#!/usr/bin/env python
+# Helper tool for adoption of a Director deployed OpenStack.
+# Conde is anything but good, though it should be somewhat useful.
+# It helps create a draft patch file (cinder.patch) with the cinder
+# configuration.
+# It may also create a file (cinder-prereq.yaml) with manifests for secrets and
+# MachineConfigs depending on the provider cinder configuruation file.
+#
+import argparse
+import base64
+import collections
+import copy
+import logging
+import os
+
+import yaml
+
+
+LOG = logging
+PATCH_FILE = 'cinder.patch'
+PREREQ_FILE = 'cinder-prereq.yaml'
+
+CINDER_TEMPLATE = """
+spec:
+  cinder:
+    enabled: true
+    apiOverride:
+      route: {}
+    template:
+      databaseInstance: openstack
+      secret: osp-secret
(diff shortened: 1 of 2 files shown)


docs_user/modules/cinder-cfg.py
```diff
@@ -0,0 +1,765 @@
+#!/usr/bin/env python
```
jistr, line 1: How will this script be used in downstream context? I remember at some point we discussed whether we should be using just OS Diff for everything and phase this script out but i can't find what the result of the discussion was.

Akrog: I don't know if it will be used or not.
I know we will not be providing an RPM for it or anything like that.

I wasn't part of the OS Diff discussion, and I haven't tested that tool against the various cinder configuration scenarios I tried this script with back then.

jistr: If there's no intention to ship the tool, we should remove it and reformulate the docs to not depend on it?

jistr: I'm not sure if we can just point to a Github script from downstream docs...

Akrog: I don't know if we can or cannot, but what we have in this upstream repo is broken, so I was just looking for a quick fix right now.

jistr: Ack i think we can merge this for now (the PR also contains other tweaks).

We also discussed today in UA/TC/PM sync for Upgrades -- we could keep this for upstream only in theory, but we need to get rid of mentioning anything in the downstream docs render that we do not intend to ship. The downstream/upstream rendering can be differentiated via if/else branching as seen here for example:

https://github.com/openstack-k8s-operators/data-plane-adoption/pull/286

But my 2 cents would be to keep the upstream/downstream instructions as close as possible, which probably means not relying on this script in documentation. IMO there is limited utility to anything we don't intend to ship and support...

I'll create a bug for tracking.

PR #318, opened 2024-03-01
docs: first pass at ironic adoption text

juliakreger:


docs_user/modules/openstack-ironic_adoption.adoc
```diff
@@ -8,20 +8,181 @@
...
+      ironicInspector:
+        replicas: 1
+      ironicNeutronAgent:
+        replicas: 1
+      secret: osp-secret
+'
+----
+
+// TODO ? Do we want to include something like nova has
+// oc wait --for condition=Ready --timeout=300s Nova/nova
```
jistr, line 115: Perhaps if we want to make sure everything has come up before we go for the RBAC steps below. But for majority of services we don't include such command, we've just been relying on the "post checks" to verify that everything has come up fine (and in tests we have a retry loop on the post checks with some timeout).

jistr (approved):

docs_user/modules/openstack-ironic_adoption.adoc
juliakreger: That was kind of my impression and quandary as well. The API/Conductor shouldn't take long to really start in the grand scheme of the universe. Likely something we just need to get some feedback upon to really make a solid decision upon, and should thus drive towards consistency.

steveb (approved):
/lgtm

PR #319, opened 2024-03-01
Neutron SR-IOV agent adoption

slawqo:
This patch adds adoption of the neutron-sriov-nic-agent on the edpm nodes.
As we run our tests on the virtualized environment, there is no really way to spawn VM which uses SR-IOV port and test if this VM works after adoption. This will need to be tested in some other layer of our CI. This patch adds documentation about adoption of the SR-IOV agent and simple test which will make sure that neutron-sriov-nic-agent is ALIVE after adoption.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/755

Closes: OSPRH-2433

### DIFF
--- a/docs_user/modules/openstack-edpm_adoption.adoc
+++ b/docs_user/modules/openstack-edpm_adoption.adoc
@@ -108,6 +108,12 @@ spec:
 EOF
 ----
 
+* In case when neutron-sriov-nic-agent is running on the existing compute nodes,
+  physical device mappings needs to be checked and set the same in the
+  OpenStackDataPlaneNodeSet CR. Those options will need to be set in the
+  OpenStackDataPlaneNodeSet CR.
+  For more information, see xref:pulling-the-openstack-configuration_{context}[Pulling the OpenStack configuration].
+
 == Procedure - EDPM adoption
 
 * _Temporary fix_ until the OSP 17 https://code.engineering.redhat.com/gerrit/q/topic:stable-compute-uuid[backport of the stable compute UUID feature]
@@ -394,6 +400,31 @@ endif::[]
         edpm_ovs_packages:
         - openvswitch3.1
 EOF
+----
+
+* Optionally enable neutron-sriov-nic-agent in the OpenStackDataPlaneNodeSet CR
++
+[source,yaml]
+----
+oc patch openstackdataplanenodeset openstack --type='json' --patch='[
+  {
+    "op": "add",
+    "path": "/spec/services/-",
+    "value": "neutron-sriov"
+  }, {
+    "op": "add",
+    "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_physical_device_mappings",
+    "value": "dummy_sriov_net:dummy-dev"
+  }, {
+    "op": "add",
+    "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_resource_provider_bandwidths",
+    "value": "dummy-dev:40000000:40000000"
+  }, {
+    "op": "add",
+    "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_resource_provider_hypervisors",
+    "value": "dummy-dev:standalone.localdomain"
+  }
+]'
 ----
 
 * Deploy OpenStackDataPlaneDeployment:
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -256,6 +256,31 @@
             edpm_ovn_ofctrl_wait_before_clear: 8000
     EOF
 
+- name: set neutron-sriov-nic-agent configuration in the OpenStackDataPlaneNodeSet CR
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -16,6 +16,10 @@ test-with-ceph:
 	mkdir -p tests/logs
 	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_with_ceph.yaml 2>&1 | tee $(TEST_OUTFILE)
 
+test: TEST_OUTFILE := tests/logs/test_with_ceph_out_$(shell date +%FT%T%Z).log
+test:
+	mkdir -p tests/logs
+	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test.yaml 2>&1 | tee $(TEST_OUTFILE)
```
jistr, line 22: I guess this is your development tweak? The test.yaml playbook doesn't exist in the repo. Otherwise LGTM.

slawqo: yes, sorry for that :)

docs_user/modules/openstack-edpm_adoption.adoc
```diff
@@ -108,6 +108,44 @@ spec:
 EOF
 ----
 
+* In case when neutron-sriov-nic-agent is running on the existing compute nodes,
+  physical device mappings needs to be checked and set the same in the
+  OpenStackDataPlaneNodeSet CR
+
+----
+openstack network agent list
```
booxter, line 116: This is great but isn't it too late (assuming the reader goes from start to the end of the guide)? I believe when they get to EDPM adoption, their API services are already down and the command below won't work. Should we instruct them to pull the data earlier, somewhere in Network section perhaps?

docs_user/modules/openstack-edpm_adoption.adoc
```diff
@@ -392,6 +430,31 @@ endif::[]
...
+    "path": "/spec/services/-",
+    "value": "neutron-sriov"
+  }, {
+    "op": "add",
+    "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_physical_device_mappings",
+    "value": "dummy_sriov_net:dummy-dev"
+  }, {
+    "op": "add",
+    "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_resource_provider_bandwidths",
+    "value": ""
```
booxter, line 449: it would be helpful to see an actual non-empty value to copy (above and here). (Also below.)

slawqo: done

docs_user/modules/openstack-edpm_adoption.adoc
slawqo: You are totally right. I checked it once again and just added link to the "Pulling openstack configuration section" for more details. I think this should be enough.

booxter: I think the newer version is still too late. We stop API services very early in the process, that's why it's important to pull the sriov configuration data prior to any actual adoption steps. At the very least, we shoud inform the reader in the Pulling the configuration section somehow that they have to take care of SR-IOV (either as a link from that earlier section to here, or by moving the text about pulling the sriov configuration there directly; I prefer the latter.)

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -256,6 +256,31 @@
...
+      }, {
+        "op": "add",
+        "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_resource_provider_bandwidths",
+        "value": ""
+      }, {
+        "op": "add",
+        "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_neutron_sriov_agent_SRIOV_NIC_resource_provider_hypervisors",
+        "value": ""
+      }]'
+  when: edpm_neutron_sriov_agent_enabled is true
```
karelyatin, line 282: edpm_neutron_sriov_agent_enabled|bool

tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
```diff
@@ -18,6 +18,20 @@
   retries: 2
   delay: 1
 
+- name: get neutron-sriov-nic-agent alive state
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ neutron_header }}
+    ${BASH_ALIASES[openstack]} network agent list | grep -F 'neutron-ovn-metadata-agent' | grep -qF 'XXX' || echo ALIVE
```
karelyatin, line 25: should be sriov agent instead of ovn metadata agent

docs_user/modules/openstack-edpm_adoption.adoc
slawqo: I changed it once again, adding info about fetching sriov agent's configuration in the "pull openstack configuration" section and linking to the edpm adoption. Instead of using neutron API I wrote it in the way that data is fetched directly from the neutron DB which is consistent with how other things are fetched there (like e.g. hostnames of the nova hypervisor names, etc.)

fao89:
should we merge it now?

jistr:
Yup the dependency is merged. I created the "check-before-merge/depends-on" label mainly as a reminder to myself, since i sometimes tend to miss the depends-on and merge things out of order :).

PR #327, opened 2024-03-05
Fix keystone service cleanup list

jistr:
Fixes https://issues.redhat.com/browse/OSPRH-5372


jistr:
recheck data plane timeout

fao89 (approved):

PR #329, opened 2024-03-05
Show upstream version in github preview

gibizer:
The github doc preview is missing sections of the doc as the build
variable is not defined in that env and therefore all our conditional
upstream/downstream ifeval blocks are hidden.

This patch changes the upstream conditions from:
```    ifeval::["{build}" == "upstream"]```
to
```    ifeval::["{build}" != "downstream"]```
so the upsteam blocks are visible even if the build variable is not
defined.

### DIFF
--- a/docs_user/modules/ceph-monitoring_migration.adoc
+++ b/docs_user/modules/ceph-monitoring_migration.adoc
@@ -90,7 +90,7 @@ shipped by cephadm, remove the following config options from the Ceph cluster.
 $ ceph config dump
 ...
 ...
-ifeval::["{build}" == "upstream"]
+ifeval::["{build}" != "downstream"]
 mgr   advanced  mgr/cephadm/container_image_alertmanager    undercloud-0.ctlplane.redhat.local:8787/ceph/alertmanager:v0.25.0
 mgr   advanced  mgr/cephadm/container_image_base            undercloud-0.ctlplane.redhat.local:8787/ceph/ceph:v18
 mgr   advanced  mgr/cephadm/container_image_grafana         undercloud-0.ctlplane.redhat.local:8787/ceph/ceph-grafana:9.4.7
@@ -173,7 +173,7 @@ is not supported for an external Ceph cluster.
 [source,bash]
 ----
 for item in $(sudo cephadm shell --  ceph orch host ls --format json | jq -r '.[].hostname'); do
-    sudo cephadm shell -- ceph orch host label add  $item monitoring; 
+    sudo cephadm shell -- ceph orch host label add  $item monitoring;
 done
 ----
 
@@ -373,7 +373,7 @@ prometheus.cephstorage-2    cephstorage-2.redhat.local  172.17.3.144:9092
 
 [source,bash]
 ----
-[ceph: root@controller-0 /]# ceph config dump 
+[ceph: root@controller-0 /]# ceph config dump
 ...
 ...
 mgr  advanced  mgr/dashboard/ALERTMANAGER_API_HOST   http://172.17.3.83:9093
--- a/docs_user/modules/ceph-rbd_migration.adoc
+++ b/docs_user/modules/ceph-rbd_migration.adoc
@@ -216,7 +216,7 @@ force-purge the containers in the node:
 ----
 [root@oc0-controller-1 ~]# sudo podman ps
 CONTAINER ID  IMAGE                                                                                        COMMAND               CREATED         STATUS             PORTS       NAMES
-ifeval::["{build}" == "upstream"]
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by klgill** [comment]
The changes made in this PR fulfill the need described in Gibi's comment. The upstream doc preview is building with the sections that include the ifeval conditions.
However, the use of ifevals will likely be changing because we found that the ifevals do not render downstream. I'm not sure whether to approve this PR for now, and in the meantime, the docs team will find a solution.

**Comment by gibizer** [comment]
> The changes made in this PR fulfill the need described in Gibi's comment. The upstream doc preview is building with the sections that include the ifeval conditions. However, the use of ifevals will likely be changing because we found that the ifevals do not render downstream. I'm not sure whether to approve this PR for now, and in the meantime, the docs team will find a solution.

Do you use the Makefile from this repo downstream? The `BUILD` env variable defines which version of the doc is built. 
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/Makefile#L1

Or if you use the CLI directly the you need to pass `-a build=downstream` to the asciidoctor command to define the variable checked by the ifeval blocks in the doc
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/Makefile#L59

**Comment by jistr** [comment]
Rendering upstream by default so that it shows correctly in Github preview makes sense to me. Let's pause until we figure out the downstream rendering though. I think we should be able to solve both without conflicting (as long as we don't default to downstream). I mentioned it in https://issues.redhat.com/browse/OSPRH-5429 .

@gibi once we know how to proceed on the Jira bug linked above, would you be able to add your snippet to all the modules, to keep consistency? I think your solution here would likely still apply as-is even after we have the downstream rendering figured out, but IMO let's pause work on this for now, just in case...

**Comment by gibizer** [comment]
> Rendering upstream by default so that it shows correctly in Github preview makes sense to me. Let's pause until we figure out the downstream rendering though. I think we should be able to solve both without conflicting (as long as we don't default to downstream). I mentioned it in https://issues.redhat.com/browse/OSPRH-5429 .

OK, lets wait for the downstream solution and then reevaluate.

> 
> @gibi once we know how to proceed on the Jira bug linked above, would you be able to add your snippet to all the modules, to keep consistency? I think your solution here would likely still apply as-is even after we have the downstream rendering figured out, but IMO let's pause work on this for now, just in case...

I added it already to all the modules where we have ifeval blocks, but you have a good point adding it to all modules for forward correctness. Sure I can do that.****

**Comment by jistr** [comment]
> I added it already to all the modules where we have ifeval blocks, but you have a good point adding it to all modules for forward correctness. Sure I can do that.

Ah i didn't realize you covered all that have the if blocks. Then adding them elsewhere i think is nice (i think it's likely we'll miss that part if adding if blocks elsewhere) but not a blocker.

**Comment by fao89** [comment]
Maybe the pattern should be:
```
```
instead of
```
ifeval::["{build}" == "upstream"]
```

**Comment by jistr** [comment]
We have a downstream solution, and the condition can continue to be `ifeval::["{build}" == "downstream"]`.

I think we should do what Fabricio suggested, which seems to be the simplest way:

```
< upstream content goes here >
endif::[]

< downstream content goes here >
endif::[]
```

gibizer:
> Maybe the pattern should be:
> 
> ```
> ifeval::["{build}" != "downstream"]
> ```
> 
> instead of
> 
> ```
> ifeval::["{build}" == "upstream"]
> ```

Do you also suggest to drop BUILD=upstream logic from the Makefile:
* https://github.com/openstack-k8s-operators/data-plane-adoption/blob/eba492929601abb964735b6325481dab9a2b713a/Makefile#L35
* https://github.com/openstack-k8s-operators/data-plane-adoption/blob/eba492929601abb964735b6325481dab9a2b713a/docs_user/Makefile#L1
* https://github.com/openstack-k8s-operators/data-plane-adoption/blob/eba492929601abb964735b6325481dab9a2b713a/docs_user/Makefile#L1

If none of our docs condition refers to the "upstream" value of BUILD then the Makefile should not do it either? Or do we foresee a case where we will have a 3rd value for BUILD ?

jistr:
> If none of our docs condition refers to the "upstream" value of BUILD then the Makefile should not do it either? Or do we foresee a case where we will have a 3rd value for BUILD ?

I don't see a 3rd value currently... We could drop the "upstream" value from the Makefile but we do need to still support "BUILD=downstream" there so we can't fully drop the BUILD variable. So IMO we may keep the "upstream" value in the Makefiles too... I don't have a strong opinion on that one...

fao89 (approved):
thank you!

jistr (approved):
:+1: thanks!

PR #333, opened 2024-03-06
Write uuid to /var/lib/nova/compute_id only if needed

bogdando:
File /var/lib/nova/compute_id may be existing and locked with chattr -x. Fix idempotency of writing uuid into it, only when its stored uuid contents doesn't match the one needs to be written there.

### DIFF
--- a/docs_user/modules/openstack-edpm_adoption.adoc
+++ b/docs_user/modules/openstack-edpm_adoption.adoc
@@ -127,7 +127,7 @@ do
   ssh \
     -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa \
     root@"${computes[$name]}" \
-                "echo $uuid | sudo tee /var/lib/nova/compute_id && sudo chown 42436:42436 /var/lib/nova/compute_id && sudo chcon -t container_file_t /var/lib/nova/compute_id"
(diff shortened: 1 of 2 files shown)


docs_user/modules/openstack-edpm_adoption.adoc
```diff
@@ -127,7 +127,7 @@ do
   ssh \
     -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa \
     root@"${computes[$name]}" \
-                "echo $uuid | sudo tee /var/lib/nova/compute_id && sudo chown 42436:42436 /var/lib/nova/compute_id && sudo chcon -t container_file_t /var/lib/nova/compute_id"
+      "grep -qF $uuid /var/lib/nova/compute_id || (echo $uuid | sudo tee /var/lib/nova/compute_id && sudo chown 42436:42436 /var/lib/nova/compute_id && sudo chcon -t container_file_t /var/lib/nova/compute_id)"
```
jistr, line 130: I guess we can merge but this will go away with https://issues.redhat.com/browse/OSPRH-2302 anyway, right?

gibizer: It will be deleted altogether when 17.3 will be on the CDN as in 17.3 the uuid file is generated by nova already.

gibizer (approved):
@bogdando good point. I did not considered idempotency of the adoption steps when I wrote this. But make sense to only generate the file if it is not there yet

PR #337, opened 2024-03-08
Update dataplane

fao89:
remove default variables and add a new service

### DIFF
--- a/tests/roles/dataplane_adoption/vars/rhev.yaml
+++ b/tests/roles/dataplane_adoption/vars/rhev.yaml
@@ -129,11 +129,4 @@ edpm_network_config_template: |
   {% endraw %}
 neutron_physical_bridge_name: br-ex
 neutron_public_interface_name: eth4
-role_networks:
-  - InternalApi
-  - Tenant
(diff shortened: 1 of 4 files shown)


tests/roles/dataplane_adoption/vars/rhev.yaml
```diff
@@ -129,11 +129,4 @@ edpm_network_config_template: |
   {% endraw %}
 neutron_physical_bridge_name: br-ex
 neutron_public_interface_name: eth4
-role_networks:
-  - InternalApi
-  - Tenant
```
jistr, line 134: Do we need to keep the role_networks variable? AFAICT the default includes the Storage network but we don't have it here in the RHEV config. cc @archana203

fao89: it is generated here: https://github.com/openstack-k8s-operators/dataplane-operator/blob/04387adc47562b7db9be3d2d6e1ab7370e0748c2/pkg/deployment/inventory.go#L286

jistr (approved):

PR #340, opened 2024-03-11
Enable edpm_nodes_validation_validate_fqdn

fao89:


jistr:
@gibizer @SeanMooney IIRC we only wanted this for Dev Preview 3 but not for GA, right?

fao89:
recheck

jistr (approved):
I'm not sure whether we wanted this in main as well, or just DP3... But it's not a blocker, we can remove the var in main later if we discover the validation fails when it shouldn't.

gibizer:
> @gibizer @SeanMooney IIRC we only wanted this for Dev Preview 3 but not for GA, right?

Sorry for the late reply. We can enable this in DP3 as there we know we cannot adopt a cloud where this validation fails. 
In GA we would like to support adoption of a cloud where the /etc/hostname does not contain an FQDN. The is being slowly fixed in https://github.com/openstack-k8s-operators/edpm-ansible/pull/587

jistr:
@fao89 So i think we should only target this to dev-preview3 branch? And i don't think we'll have the opportunity to backport https://github.com/openstack-k8s-operators/edpm-ansible/pull/595 into DP3. Do you think the validation still makes sense without it or would it be broken always?

fao89:
@jistr I don't know, I would need to find the piece of code that adds `START_HOST_ENTRIES_FOR_STACK` to the `/etc/hosts`,
our code is catching the comments because `standalone` is there: `# START_HOST_ENTRIES_FOR_STACK: standalone`. I wonder if the comment always adds "standalone" or if it is indeed short of the hostname

jistr:
Ah i see, i think it wouldn't be the short of hostname but the TripleO Heat stack name. In production envs, if i recall correctly, the default would be "overcloud", and i think that would likely feature in the host FQDNs but not in their short names. So we may be ok in majority of cases. I think it worked for Pedro with the validation.

So IMO let's post this against DP3 branch, and we can drop the dependency? That's our last thing to backport into DP3 i think and we're ready :rocket: .

fao89:
recheck

jistr (approved):

PR #352, opened 2024-03-13
Change docs layout

fao89:


jistr:
Looks unchanged? https://logserver.rdoproject.org/52/352/2c4f1b1634f26433630f810bcadede361e9f147a/github-check/adoption-docs-preview/52d9979/docs_build/docs_build/adoption-user/index-upstream.html

fao89:
Actually, it is minor things such as icons and callouts, I have an example here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/762

fao89:
current:
image
image


PR:
image
image

jistr:
Ah cool, that looks good then. @klgill is this ok from downstream perspective? I think we don't use the upstream `main.adoc` file downstream?

klgill:
> Ah cool, that looks good then. @klgill is this ok from downstream perspective? I think we don't use the upstream `main.adoc` file downstream?

You're right. We don't use the main.adoc file downstream. These notes are going to look a bit different when rendered downstream. For example, the asciidoc formatting is [IMPORTANT] and [WARNING]. I think this is fine for upstream, and then I'll change the formatting downstream eventually.

klgill (approved):

PR #355, opened 2024-03-14
Remove reboot-os role during edpm adoption

bogdando:
Rebooting of dataplane nods shuold be avoided at all costs during adoption.

Remove reboot-os role to recover the basic VMs adoption case with local storage.

JIRA [OSPRH-3123](https://issues.redhat.com/browse/OSPRH-3123)

### DIFF
--- a/docs_user/modules/openstack-edpm_adoption.adoc
+++ b/docs_user/modules/openstack-edpm_adoption.adoc
@@ -260,7 +260,6 @@ spec:
     - configure-os
     - ssh-known-hosts
     - run-os
-    - reboot-os
     - install-certs
     - libvirt
     - nova-compute-extraconfig
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -161,7 +161,6 @@
         - configure-os
         - ssh-known-hosts
         - run-os
-        - reboot-os
         - install-certs
         - libvirt
         - nova-compute-extraconfig


fao89 (approved):
Maybe it is worth adding a warning in the docs mentioning the nodes shouldn't reboot during the adoption

bogdando:
@jistr could you please merge ?

bogdando:
> Maybe it is worth adding a warning in the docs mentioning the nodes shouldn't reboot during the adoption

this is a good point, we can address the docs updates in follow ups, to avoid loading CI systems

fmount (approved):
/lgtm

docs_user/modules/openstack-edpm_adoption.adoc
```diff
@@ -260,7 +260,6 @@ spec:
     - configure-os
     - ssh-known-hosts
     - run-os
-    - reboot-os
```
gibizer, line 263: We need to note somewhere that this is just a WA. The solution should also prevent any reconfiguration by edpm_ansible on the EDPM node that would require the reboot-os role to be run. If we don't catch those reconfigurations the we create a timebomb on the EDPM node that will activate at the next unrelated reboot of the node.

gibizer: I think we need to file a bug describing the case when such reconfiguration happens during adoption (e.g. different kernelargs).

jistr (approved):

docs_user/modules/openstack-edpm_adoption.adoc
jistr: I wonder if the "reboot os" role should be made somehow safe by default? If this role is in the list, and some of the sensitive parameters change, would it just reboot the whole dataplane? (Regardless if it was adopted or not...) Maybe the role should have some safety param that needs to be set for the reboots to be enabled, and if it detects it should reboot but this parameter isn't set, it should fail...

gibizer: Would that failure happen too late after some changes(e.g. another dataplane service run to completion) on the node that prevents the rollback to the old software version?

PR #357, opened 2024-03-14
Document details about network adoption when using a new subnet for CP

booxter:
Using a new subnet for CP requires that we set up link local routes on both OCP workers and tripleo nodes. For OCP, it's done in `NetworkNodeConfigurationPolicy` CRs. For tripleo nodes, it's done via `net_config.yaml` files, listing additional `routes` with `next_hop` set to {{ ctlplane_ip }}.

Once this is done, the next step is to set EDP NodeSet to retain the routes in net_config; as well as configure ssh range to allow access from ansibleee pods.

This procedure was validated using
https://github.com/openstack-k8s-operators/install_yamls/pull/774

### DIFF
--- a/docs_user/modules/openstack-planning.adoc
+++ b/docs_user/modules/openstack-planning.adoc
@@ -320,9 +320,6 @@ At the end of this process, you should have the following information:
 
 === IPAM planning
 
-// TODO: explain which IP addresses will change during adoption, and which will
-// stay the same.
-
 The new deployment model puts additional burden on the size of IP allocation
 pools available for OpenStack services. This is because each service deployed
 on OpenShift worker nodes will now require an IP address from the IPAM pool (in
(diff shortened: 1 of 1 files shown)


docs_user/modules/openstack-planning.adoc
```diff
@@ -386,16 +386,147 @@ addresses for the new control plane services.
...
+
+```yaml
+network_config:
+  - type: ovs_bridge
+    name: br-ctlplane
+    routes:
+    - ip_netmask: 0.0.0.0/0
+      next_hop: 192.168.1.1
+    - ip_netmask: 172.31.0.0/24  # <- new ctlplane subnet
+      next_hop: 192.168.1.100    # <- adopted node ctlplane IP address
```
jistr, line 429: So e.g. for setting up a route for the new storage subnet, the `next_hop` used would be the node's own IP on the old storage subnet?

jistr (approved):
LGTM, thank you! :+1:

docs_user/modules/openstack-planning.adoc
booxter: The route is slightly different (dev <bridge> in OCP; via <IP> in OSP), but the effect seems identical. The reason why this works in OSP is because Linux, when sending a reply, will pick the route with the longest matching prefix with the destination. (Without this route, the frame was sent to default gateway, which then redirected it back to destination. This of course assumed that the default gateway would redirect and should not be relied on.)

jistr: Ack thanks for the explanation. Merging :)

PR #359, opened 2024-03-15
Add Swift data migration docs and test

cschwede:
Describes and tests migrating Swift data from existing nodes to new nodes using PVCs as backends.

Due to the length of the section and the possibility to migrate data without downtime after all other services, this has been put into its own assembly.

Tests are put into a new role and not enabled by default, as these take quite some time (~20-25 min) to finish. It also requires a standalone deployment with at least two replicas for Swift. Standalone with a single Swift replica does not enable replication processes, but these are required for the migration.

If using install_yamls this can be enabled with the following environment variable:

  SWIFT_REPLICATED=true make standalone

And tests can then be run after adoption finished:

  make test-minimal
  make test-swift-migration

Tests are using "forced_rebalance" instead of "rebalance"; this is to avoid a minimum waiting time of one hour between rebalances in the tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,10 @@ test-with-ceph:
 	mkdir -p tests/logs
 	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) tests/playbooks/test_with_ceph.yaml 2>&1 | tee $(TEST_OUTFILE)
 
+test-swift-migration: TEST_OUTFILE := tests/logs/test_swift_migration_out_$(shell date +%FT%T%Z).log
+test-swift-migration:
+	mkdir -p tests/logs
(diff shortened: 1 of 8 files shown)


cschwede:
I used the following script for testing. The openstack-operator-index is currently required due to recently merged patches to swift-operator, should be no longer needed after openstac-operator is updated. install_yamls should also be updated.

```
#!/bin/sh
export OPENSTACK_IMG=quay.io/cschwede/openstack-operator-index:v0.0.1
export CPUS=12 MEMORY=25600 DISK=100
export SWIFT_REPLICATED=true

cd ~/install_yamls/devsetup/
make crc_cleanup
make edpm_compute_cleanup
make crc crc_attach_default_interface 

cd ~/install_yamls
make nmstate namespace

cd ~/install_yamls/devsetup/
make bmaas standalone

cd ~/install_yamls
make crc_storage input openstack_wait

rm ~/.ssh/known_hosts
ssh-keyscan 192.168.122.100 >> ~/.ssh/known_hosts

scp -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100:/root/tripleo-standalone-passwords.yaml ~/

cd ~/data-plane-adoption
make test-minimal
make test-swift-migration
```

jistr (approved):
LGTM, thanks!

PR #363, opened 2024-03-15
Fix ceph mds ascidocs warnings

bogdando:

PR #338, opened 2024-03-08
Fix Nova server diagnostic tests

bogdando:

PR #367, opened 2024-03-18
Revert Ceph docs edit -- the problem was elsewhere

jistr:
PR https://github.com/openstack-k8s-operators/data-plane-adoption/pull/363 made the Ceph docs slightly inconsistent and didn't seem to fix the docs issue we saw in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/338 -- the problem is likely in that PR.

Let's revert https://github.com/openstack-k8s-operators/data-plane-adoption/pull/363 to make Ceph docs consistent again. I created a separate commit to fix the quoting issue in one of the snippets, i think we want to keep that.

### DIFF
--- a/docs_user/modules/ceph-mds_migration.adoc
+++ b/docs_user/modules/ceph-mds_migration.adoc
@@ -122,7 +122,7 @@ Check the current OSD blocklist and clean up the client list:
 [ceph: root@controller-0 /]# ceph osd blocklist ls
 ..
 ..
-for item in $(ceph osd blocklist ls | awk ‘{print $0}’); do
+for item in $(ceph osd blocklist ls | awk '{print $0}'); do
      ceph osd blocklist rm $item;
 done
 ----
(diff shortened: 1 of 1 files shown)


fmount (approved):
/lgtm

fmount:
ack and thanks for confirming that syntax was correct.

jistr:
recheck error fetching from rubygems.org

jistr:
@klgill No, the issue was unrelated, it was missing end of the code block `----` in PR 338. Regarding `[source,bash]` i only put the state back into how it was before. I saw at least one other Ceph module has `[source,bash]`, so it seemed random to remove it from that one doc, and i put it back to the original state before #363 was merged. #363 was merged only because we thought the CI was broken, otherwise we'd be looking at that patch more carefully i think.

So in conclusion, nothing of importance happened here i think :grin: . PR 363 landed and here we immediately reverted it after we discovered it wasn't the fix we were looking for.

jistr:
We kept just this bit from PR 363 -- https://github.com/openstack-k8s-operators/data-plane-adoption/pull/367/commits/cbcf2dbf686bd64b72d2190a8e0b3b095cbd0a74

PR #368, opened 2024-03-19
Remove install_yamls reference from user facing docs

fao89:


jistr:
I think it makes sense for downstream docs especially. For upstream docs used by developers, i think it has been kinda useful to have `ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa ...` in the examples. So maybe the lint could target the rendered index-downstream.html? This one: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/247eaaf96798556622bb54a539f1b6e09ee7e211/.github/workflows/docs.yaml#L115

jistr (approved):
Thanks, LGTM

PR #369, opened 2024-03-19
copying downstream DP3 files back upstream

klgill:
This PR copies the files that we used downstream for DP3 back upstream. The DP3 files were peer reviewed downstream, so this PR is intended to ensure that downstream and upstream files are in sync before I begin restructuring the docs for Beta.
I have a couple of questions that must be addressed before we merge this PR:

1) Should this PR be merged into main only, or both main and the dev-preview3 branch?
2) The files I copied contain some ifeval conditions that are specific to DP3. Should those conditions be removed?
3) How can I be sure I'm not copying over anything new that was added in main? For example, the "Barbican" heading is not included in the "Planning" chapter downstream.


jistr:
Majority of the changes proposed in this PR are undoing the progress that's been done on main branch since the fork of Dev Preview 3. So i don't think we want to get those into the main branch. We may need a different approach than a blanket copy of downstream DP3 over the main branch. Probably just picking the individual changes that you've done to downstream?

Or you could do a blanket copy locally in your repo checkout, and then use `git add --patch` to only stage for commiting those individual bits that are the edits we want to forward-port into `main`.

klgill:
Closing this PR because it copied over progress that had been made in the main branch. I will manually add the peer review comments from downstream in another PR.

PR #371, opened 2024-03-20
Do not mention local backend in the Glance adoption procedure

fmount:
Local backend, aka File, is not something supported in TripleO. This patch removes the local backend section to make sure we do not generate confusion about the target/adopted glance API.


jistr (approved):

konan-abhi (approved):
++ Thank you!

PR #372, opened 2024-03-20
Tests for the control plane rollback

jistr:


tests/roles/control_plane_rollback/tasks/main.yaml
```diff
@@ -0,0 +1,149 @@
...
+    oc delete --ignore-not-found=true --wait=false pod mariadb-copy-data
+    oc delete --ignore-not-found=true --wait=false pvc mariadb-data
+    oc delete --ignore-not-found=true --wait=false pod ovn-copy-data
+    oc delete --ignore-not-found=true secret osp-secret
+
+- name: verify that the source control plane is responding
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    ssh -i {{ edpm_privatekey_path }} root@{{ edpm_node_ip }} OS_CLOUD=standalone openstack user list
```
archana203, line 145: Can we make this command more generic. So that, it will verify the controlplane in case of HA OSP cloud deployed on RHV or other platform instead of standalone specific.

jistr: And i tweaked it a bit more in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/372/commits/10522966d5297e2ee4dca9302a763014fdcde3dc to make it use `$CONTROLLER1_SSH`. But still there's the `OS_CLOUD=standalone` bit so the variable may need overriding in non-standalone environments anyway.

fao89 (approved):

tests/roles/control_plane_rollback/tasks/main.yaml
archana203: The override variable can be put in standard vars.yaml?

tests/playbooks/test_rollback_minimal.yaml
archana203, line 1: In the Makefile, are we going to add the calling for test_rollback_minimal playbook?

tests/roles/control_plane_rollback/tasks/main.yaml
jistr: Yes i think that should work.

tests/playbooks/test_rollback_minimal.yaml
jistr: Ah i forgot to add the Makefile part because i simultaneously did another edit there to make the testing easier (using TEST_ARGS instead of TESTS_TO_SKIP). I'll merge this PR to make use of the green CI and post another PR right away with the Makefile changes.

jistr: Here it is https://github.com/openstack-k8s-operators/data-plane-adoption/pull/394

PR #376, opened 2024-03-21
do not create a seperate nova service for post ffu

SeanMooney:
the previous workflow created a new configmap and
dataplane service to deploy nova in a second deployment
post ffu. this was inherintly buggy as the node set still
refernced the orginal dataplane service and the old configmap.

That means if you did a deployment in the futrue it would revert
to the ffu config.

This change correct that by updating the workflow to modify the existing
configmap and redeploy the orginal service. This is more inline with
how adoption is intended to work as we dont intend multiple datapalne
services for the same openstack service to be used against the
same nodeset.

### DIFF
--- a/docs_user/modules/openstack-edpm_adoption.adoc
+++ b/docs_user/modules/openstack-edpm_adoption.adoc
@@ -584,26 +584,12 @@ oc apply -f - <<EOF
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: nova-compute-ffu
+  name: nova-compute-extraconfig
   namespace: openstack
 data:
-  20-nova-compute-cell1-ffu-cleanup.conf: |
+  20-nova-compute-cell1-workarounds.conf: |
     [workarounds]
     disable_compute_service_check_for_ffu=false
----
-apiVersion: dataplane.openstack.org/v1beta1
-kind: OpenStackDataPlaneService
-metadata:
-  name: nova-compute-ffu
-  namespace: openstack
-spec:
-  label: nova.compute.ffu
-  configMaps:
-    - nova-compute-ffu
-  secrets:
-    - nova-cell1-compute-config
-    - nova-migration-ssh-key
-  playbook: osp.edpm.nova
 ---
 apiVersion: dataplane.openstack.org/v1beta1
 kind: OpenStackDataPlaneDeployment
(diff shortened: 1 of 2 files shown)


fao89 (approved):

jistr (approved):
LGTM but tagging @gibizer and @bogdando to give time for Compute peer review

tests/roles/dataplane_adoption/tasks/nova_ffu.yaml
```diff
@@ -95,18 +81,38 @@
...
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc wait --for condition=Ready osdpd/openstack-nova-compute-ffu --timeout=5m
-  register: nova_ffu_edpm_result
-  until: nova_ffu_edpm_result is success
-  retries: 10
-  delay: 6
+
+    DEPLOYMENT_NAME=openstack-nova-compute-ffu
```
bogdando, line 92: I am not sure why just waiting for the ready status condition is no longer  sufficient.
If we really have to, please adjust the docs to follow this change

docs_user/modules/openstack-edpm_adoption.adoc
```diff
@@ -584,28 +584,14 @@ oc apply -f - <<EOF
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: nova-compute-ffu
+  name: nova-compute-extraconfig
```
bogdando, line 587: this passes in CI, but I'm not sure how updating this cm works in-place as we've created it already earlier https://github.com/openstack-k8s-operators/data-plane-adoption/pull/376/files ?
I though k8s apisesrver reject updating configmaps, and those need to be recreated?

tests/roles/dataplane_adoption/tasks/nova_ffu.yaml
SeanMooney: the reason was at least for @vakwetu the wait can happen and fail before the deployment is created


so if we use the wait we need to extend the time interval

i.e. it took time for the nodeset to be reconsolied for the changed config map before the deployment was created so all the reties were used before the resource was available to wait on

docs_user/modules/openstack-edpm_adoption.adoc
SeanMooney: no you can update them as often as you like.

we are not using imutable config maps.
in this specific instance it would be incorrect too as users are expected to create there services once but modify the content of the reference config maps whenever they need to modify the deployed config.

bogdando:
LGTM

mrkisaolamb (approved):
lgtm

PR #377, opened 2024-03-22
added a rhoso attributes file and tested some variables

klgill:
I copied the downstream "rhosp-attributes" file to the upstream repo and renamed it "rhoso attributes." I added a couple of variables that apply upstream and then tested the variables in 2 files. The variables are rendering correctly in both the upstream preview and downstream preview.

I will likely add more variables to the attributes file as I review the docs more thoroughly, but this is a start.

### DIFF
--- a/docs_user/main.adoc
+++ b/docs_user/main.adoc
@@ -6,6 +6,8 @@
 :doctype: book
 :context: osp
 
+include::../docs_user/rhoso_attributes.adoc[]
+
 include::assemblies/openstack_adoption.adoc[leveloffset=+1]
 
 include::assemblies/ceph_migration.adoc[leveloffset=+1]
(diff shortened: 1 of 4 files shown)


fao89 (approved):

docs_user/main.adoc
```diff
@@ -6,6 +6,8 @@
 :doctype: book
 :context: osp
 
+include::../docs_user/rhoso_attributes.adoc[]
```
jistr, line 9: We're navigating to the parent directory and then back to `docs_user`. I think this could be just `include::rhoso_attributes.adoc[]`? Unless there's a reason to have it like this because of some downstream/upstream compatibility/likeness.

klgill: I changed it to what you suggested and it worked without errors (both in the file and the local preview). I'll commit and push my changes.

jistr (approved):
:+1: thanks

PR #380, opened 2024-03-25
Log console of the test VM if ping fails

slawqo:
After adoption, if ping to the test VM fails, we will now log console log of the test VM - that can help in some cases what was wrong and why VM is not responding anymore.

Related: OSPCIX-226

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
+++ b/tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
@@ -1,6 +1,5 @@
 ---
 - name: set Neutron services shell vars
-  no_log: "{{ use_no_log }}"
   ansible.builtin.set_fact:
     neutron_header: |
       alias openstack="oc exec -t openstackclient -- openstack"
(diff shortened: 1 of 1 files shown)


jistr (approved):
:+1:

tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
```diff
@@ -33,5 +32,13 @@
...
 - name: verify connectivity to the existing test VM instance using Floating IP
-  ansible.builtin.shell: |
-      ping -c4 {{ lookup('env', 'FIP') | default('192.168.122.20', True) }}
+  block:
+    - name: Ping VM using Floating IP
+      ansible.builtin.shell: |
+        ping -c4 {{ lookup('env', 'FIP') | default('192.168.122.20', True) }}
+  rescue:
+    - name: Get console log of the test VM
+      ansible.builtin.shell: |
```
bogdando, line 41: maybe use failed_when = false for the above task, and register outputs, then the console log should use the fail module yielding the console log as an error message? See an example https://github.com/openstack-k8s-operators/edpm-ansible/pull/667/files#diff-14d838f079a0c024754d82e20d9e153b058aef1491302e4468d39584481be706R55-R63

bogdando (commented):
lgtm just a note about logging failure as a failure

PR #331, opened 2024-03-06
TLS Everywhere adoption

xek:

PR #383, opened 2024-03-27
TLS-e Updating documentation

mauricioharley:


docs_user/modules/openstack-backend_services_deployment.adoc
```diff
@@ -136,6 +138,11 @@ oc set data secret/osp-secret "SwiftPassword=$SWIFT_PASSWORD"
 * Deploy OpenStackControlPlane. *Make sure to only enable DNS,
 MariaDB, Memcached, and RabbitMQ services. All other services must
 be disabled.*
+
+* If the source environment does not use TLS Everywhere, modify
+spec:tls:endpoint:internal:enabled to false. Do the same for
+spec:tls:endpoint:public:enabled.
```
vakwetu, line 144: Replace with:

On RHOSO 18, TLS Everywhere is enabled by default.  If the source environment does not use TLS Everywhere
and TLS Everywhere is not desired for the new environment, modify spec:tls:endpoint:internal:enabled to false.

docs_user/modules/openstack-planning.adoc
```diff
@@ -204,6 +204,23 @@ are eligible to run the different OpenStack services.  Refer to the xref:node-se
...
+
+If the Director deployment was deployed with TLS Everywhere, FreeIPA (IdM) is used
+to issue certificates for the OpenStack services. Certmonger, a client process which
+is installed on all hosts, interacts with FreeIPA (IdM) to request, install, track
+and renew these certificates.
+
+The new Operator based deployment uses the cert-manager operator to issue, track
+and renew the certificates.
+
+Because the same root CA is used to generate new certificates, the currently used chain
```
vakwetu, line 221: Replace this last sentence with:

The TLS everywhere document (link) describes how to extract the root CA from IdM in the source environment and import it for use by certmanager in the new environment.  Because the same root CA is used to issue new certificates, we can minimize disruption on the compute nodes because the currently used chain of trust does not need to be modified.

docs_user/modules/openstack-tls_everywhere.adoc
```diff
@@ -0,0 +1,211 @@
...
+== Prerequisites
+
+* Check that the source deployment is using TLS Everywhere, skip this step if it's not.
+* Make sure the previous Adoption steps have been performed successfully.
+* Make sure the backend services on the new deployment are not started yet.
+
+== Variables
+
+Define the shell variables used in the following steps. The values are illustrative and refer to a single node standalone director deployment. Use values that are correct for your environment:
+
```
vakwetu, line 20: This needs to be conditional or it will fail the build.

----
IPA_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100 podman exec -ti freeipa-server-container"
endif::[]
IPA_SSH="ssh -i <path_to_ssh_key> root@<freeipa-server-ip-address>"
endif::[]

vakwetu: This looks weird in the formatting.  Ping me for details.

docs_user/modules/openstack-tls_everywhere.adoc
```diff
@@ -0,0 +1,211 @@
...
+----
+IPA_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@<freeipa-server-ip-address>"
+----
+
+In this example the FreeIPA instance is running on a separate host, in a container.
+
+== Transfer the CA certificate and key from FreeIPA to the cert-manager Operator
+
+This is only going to work if:
```
vakwetu, line 29: Replace with:

These instructions explain how to extract the CA signing certificate from the FreeIPA instance that is used to provide the certificates in the source environment and import it into certmanager for use in the target environment.  In this way, disruption on the compute nodes can be minimized because a new chain of trust need not be installed.

It is expected that the old FreeIPA node will then be decommissioned and no longer used to issue certificates,  This may not be possible if the IPA server is used to issue certificates for non-Openstack systems.

This procedure will also need to be modified if the signing keys are stored in an HSM (hardware security module) instead of an NSSDB.  In that case, if the key is retrievable, special HSM utilities may be required.

docs_user/modules/openstack-tls_everywhere.adoc
```diff
@@ -0,0 +1,211 @@
...
+
+----
+oc get issuers -n openstack
+----
+
+----
+oc get secret rootca-internal -n openstack -o yaml
+----
+
+== Stop and disable certmonger, and remove the existing certificates
```
vakwetu, line 176: Replace with:

== Stop and disable certmonger, and remove the tracking requests for the existing certificates

tests/roles/development_environment/files/pre_launch.bash
```diff
@@ -1,5 +1,7 @@
 set -e
 
+alias openstack="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100 OS_CLOUD=standalone openstack"
```
vakwetu, line 3: This also needs to be conditional or it will fail the build.

vakwetu: (I think)  Maybe this applies only to docs

vakwetu:
changes are merged into https://github.com/openstack-k8s-operators/data-plane-adoption/pull/331

vakwetu:
closing as changes are merged into [331](https://github.com/openstack-k8s-operators/data-plane-adoption/pull/331)

PR #386, opened 2024-03-28
Add keys for Swift ring file ConfigMap

cschwede:
The swift-operator is going to use additional keys for the ring file ConfigMap. Adding these to the docs and tests.


jistr (approved):

cschwede:
recheck

PR #385, opened 2024-03-27
Beta docs restructure planning chapter

klgill:
I restructured almost all of the "Planning the new deployment chapter" except for the "Network" section because I felt that this PR was getting too big. 

Jiri and Fabricio: Let me know if this PR is too big, and I'll create smaller ones in the future. And of course, please add your feedback to the PR if you see any errors.

Peer reviewers (writers): Please follow the peer review checklist: https://docs.google.com/document/d/1hxKJdhGNqSOTg6WrbYjJaQIWDu7x5cl2QRYJ7G0HFt8/edit

PR #388, opened 2024-03-28
Beta docs restructure network section

klgill:
I restructured the "Network" section of the "Planning a new deployment" chapter. This is closely related to the following PR, which is in review: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/385

Peer reviewers (writers): Please follow the peer review checklist: https://docs.google.com/document/d/1hxKJdhGNqSOTg6WrbYjJaQIWDu7x5cl2QRYJ7G0HFt8/edit


docs_user/assemblies/openstack_adoption.adoc
rheslop, line 1: This assembly name does not adhere to mod-docs standards

docs_user/assemblies/assembly_network-requirements-and-configuration.adoc
```diff
@@ -0,0 +1,51 @@
+[id="network-requirements-and-configuration_{context}"]
+
+:context: network-requirements
+
+= Network requirements and configuration
```
rheslop, line 5: The title of this assembly looks like a conceptual module (and maybe one should be included with much of this leading information). But includes other assemblies with gerunds and a process module. 
Is this the best title here?

klgill: Maybe something like, "Configuring the network for the RHOSO deployment"?

docs_user/assemblies/openstack_adoption.adoc
klgill: I'm going to eventually delete this assembly. I'm "extracting" all the modules from this assembly first.

klgill:
Note to self: I will remove the "Network requirements and configuration" assembly from the "OpenStack adoption" assembly before merging. I only added the assembly there so the reviewers and I could see how it rendered.

docs_user/modules/proc_configuring-control-plane-services.adoc
```diff
@@ -0,0 +1,106 @@
+[id="configuring-control-plane-services_{context}"]
+
```
jistr, line 2: This is one thing that i've found strange on the downstream docs filesystem structure. The heading "Configuring control plane services" makes sense in the rendered document because it is a subheading under "Network requirements and configuration" so it's natural that we're talking about configuring control plane services in the broader context of network configuration. But the file name is "proc_configuring-control-plane-services.adoc" which may lead one to think that it is actually about configuring the control services themselves, rather than configuring the networking to make it ready for the control plane services. Configuration of the services themselves is handled elsewhere in the documentation, so that makes the file name somewhat misleading, to me at least.

So this is not a -1, if that's how we've been doing things (and from other docs i've seen it looks like that), then let's not break the pattern. So maybe this is just feedback for longer term consideration within the docs team. (I'm wondering whether you folks can navigate this comfortably :) .)

jistr (commented):
The new structure LGTM, we should remove the content from the previous place before merging (and there's a merge conflict to resolve, please let me know if i should help with that).

klgill:
> The new structure LGTM, we should remove the content from the previous place before merging (and there's a merge conflict to resolve, please let me know if i should help with that).

Yes, this "Networking" assembly needs to be put in the "Planning the new deployment" chapter (and removed from the "openstack adoption" assembly. I hadn't merged the other PR before I started this one, so the "planning the new deployment" assembly didn't exist in the repo yet. I think that's what's causing the conflict, and it should be pretty straightforward to resolve.

docs_user/modules/proc_configuring-control-plane-services.adoc
klgill: In this case, you're right that we should be more specific, especially since I'm going to create a chapter called "Adopting OpenStack control plane services" in another PR. The similar names might cause confusion. Maybe this instead?
Title of module: Configuring the networking for control plane services
Filename: proc_configuring-the-networking-for-control-plane-services.adoc

docs_user/assemblies/assembly_network-requirements-and-configuration.adoc
slawqo: I was never good in finding names or titles but it "Configuring the network for the RHOSO deployment" sounds good for me

docs_user/modules/proc_configuring-control-plane-services.adoc
slawqo: +1

slawqo (commented):
this looks good for me but I think that @booxter should also check it

klgill:
> this looks good for me but I think that @booxter should also check it

Hi, I'd like to merge this today to maintain steady progress on the Beta docs and avoid merge conflicts. I think your review is sufficient for the restructuring, unless there is something specific Ihar should review. For GA, I will be rewriting the content, so there will still be an opportunity for you and Ihar to do a thorough review.

PR #394, opened 2024-04-03
Add TEST_ARGS for 'tags', 'skip-tags' etc.; add Makefile targets for rollback tests

jistr:
I put these two commits into a single PR because they would conflict with each other if separated out.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,25 +2,24 @@ TEST_INVENTORY ?= tests/inventory.yaml
 TEST_VARS ?= tests/vars.yaml
 TEST_SECRETS ?= tests/secrets.yaml
 TEST_CONFIG ?= tests/ansible.cfg
-# String with comma seperated tags which will be passed to skip-tags during ansible-playbok run to skip part of tests
-TESTS_TO_SKIP ?= 
+TEST_ARGS ?=
 
 ### TESTS ###
 
 test-minimal: TEST_OUTFILE := tests/logs/test_minimal_out_$(shell date +%FT%T%Z).log
 test-minimal:
 	mkdir -p tests/logs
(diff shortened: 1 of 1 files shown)


jistr:
@ciecierski Do you think it's ok to remove the TESTS_TO_SKIP if we have TEST_ARGS? I could also keep TESTS_TO_SKIP, but it makes the lines in the Makefile longer and it doesn't provide additional functionality AFAICT.

Makefile
```diff
@@ -2,25 +2,34 @@ TEST_INVENTORY ?= tests/inventory.yaml
...
+	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) $(TEST_ARGS) tests/playbooks/test_swift_migration.yaml 2>&1 | tee $(TEST_OUTFILE)
+
+test-rollback-minimal: TEST_OUTFILE := tests/logs/test_rollback_minimal_out_$(shell date +%FT%T%Z).log
+test-rollback-minimal:
+	mkdir -p tests/logs
+	ANSIBLE_CONFIG=$(TEST_CONFIG) ansible-playbook -v -i $(TEST_INVENTORY) -e @$(TEST_VARS) -e @$(TEST_SECRETS) $(TEST_ARGS) tests/playbooks/test_rollback_minimal.yaml 2>&1 | tee $(TEST_OUTFILE)
+
+test-rollback-with-ceph: TEST_OUTFILE := tests/logs/test_rollback_with_ceph_out_$(shell date +%FT%T%Z).log
+test-rollback-with-ceph:
+	mkdir -p tests/logsff
```
jistr, line 31: meh typo, will push a fix

ciecierski (approved):
LGTM

jistr:
Thank you, merging.

PR #398, opened 2024-04-04
Beta docs update attributes file

klgill:
I added variables for each OSP service for the upstream build and downstream build.

### DIFF
--- a/docs_user/adoption-attributes.adoc
+++ b/docs_user/adoption-attributes.adoc
@@ -4,27 +4,72 @@
 
 ifeval::["{build}" == "upstream"]
 :OpenShift: OpenShift
-:osp_prev_long: OpenStack
-:osp_acro: OSP
+:rhos_prev_long: OpenStack
+:rhos_acro: OSP
 :OpenStackShort: OSP
 :OpenShiftShort: OCP
-:osp_curr_ver: Antelope
-:osp_prev_ver: Wallaby
+:rhos_curr_ver: Antelope
+:rhos_prev_ver: Wallaby
 :OpenStackPreviousInstaller: TripleO
+
+//Components and services
+
+//Identity service (keystone)
+:identity_service: Keystone
+
+//Shared File System service (manila)
 :rhos_component_storage_file_first_ref: Manila
 :rhos_component_storage_file: Manila
+
+//OpenStack Key Manager (barbican)
+:key_manager: Barbican
+
+//OpenStack Networking service (neutron)
+:networking: Neutron
+
+//Object Storage service (swift)
+:object_storage: Swift
+
+//Image service (glance)
+:image_service: Glance
+
 endif::[]
 
 ifeval::["{build}" == "downstream"]
-:osp_long: Red Hat OpenStack Services on OpenShift
-:osp_acro: RHOSO
-:osp_prev_long: Red Hat OpenStack Platform
+:rhos_long: Red{nbsp}Hat OpenStack Services on OpenShift (RHOSO)
+:rhos_acro: RHOSO
+:rhos_prev_long: Red{nbsp}Hat OpenStack Platform
 :OpenStackShort: RHOSP
-:osp_curr_ver: 18.0
-:osp_prev_ver: 17.1
+:rhos_curr_ver: 18.0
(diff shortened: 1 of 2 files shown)


jistr (approved):

docs_user/adoption-attributes.adoc
```diff
@@ -4,27 +4,124 @@
...
 :OpenShiftShort: OpenShift Container Platform
 :OpenStackPreviousInstaller: director
+
+//Components and services
+
+//Identity service (keystone)
+:identity_service_first_ref: Identity service (keystone)
+:identity_service: Identity service
+
+//Shared File System service (manila)
```
Jenny-Anne, line 84: All good for manila, glance, and swift, thanks. In the comments, maybe keep the s at the end of Systems in Shared File Systems service for the sake of consistency or to avoid copy and paste errors, but that's only a very minor comment on the comments :) In fact, I think Shared File System service could potentially be a better name because it's easier to read than Systems, but that's another story. 

I'm not sure why some entries include OpenStack and others don't, or why the entries for horizon and Telemetry include "Red Hat OpenStack Platform (RHOSP)" when other entries don't and OpenStack isn't a platform any more?

docs_user/adoption-attributes.adoc
```diff
@@ -4,27 +4,124 @@
...
+:rhos_long: Red{nbsp}Hat OpenStack Services on OpenShift (RHOSO)
+:rhos_acro: RHOSO
+:rhos_prev_long: Red{nbsp}Hat OpenStack Platform
 :OpenStackShort: RHOSP
-:osp_curr_ver: 18.0
-:osp_prev_ver: 17.1
+:rhos_curr_ver: 18.0
+:rhos_prev_ver: 17.1
 :OpenShift: Red Hat OpenShift Container Platform
 :OpenShiftShort: OpenShift Container Platform
```
igallagh-redhat, line 75: The short form in technical docs is RHOCP.

docs_user/adoption-attributes.adoc
igallagh-redhat: You can drop any "OpenStack" prefix from service names.

docs_user/adoption-attributes.adoc
```diff
@@ -4,27 +4,124 @@
...
+//Object Storage service (swift)
+:object_storage_first_ref: Swift
+:object_storage: Swift
+
+//Image service (glance)
+:image_service_first_ref: Glance
+:image_service: Glance
+
+//Compute service (nova)
+:compute_service_first_ref: Nova
```
JoanneOFlynn2018, line 42: Hi Katie, the upstream service names all start with a capital in the PR, they should be lowercase, correct?

docs_user/adoption-attributes.adoc
klgill: I was going by the conventions that I saw in our downstream docs. For example, on first reference of the Telemetry service,  I saw that it was written as "Red  Red Hat OpenStack Platform (RHOSP) Telemetry service".[1]
But are you saying that I can just use "Telemetry service" on first reference instead?
[1]https://access.redhat.com/documentation/en-us/red_hat_openstack_platform/16.0/html/deployment_recommendations_for_specific_red_hat_openstack_platform_services/config-recommend-telemetry_config-recommend-telemetry

igallagh-redhat: Yes, you can just say Telemetry service.

docs_user/adoption-attributes.adoc
klgill: Hmm...maybe the service names are the same for both upstream and downstream then. "Examples of these guidelines in service names are: the Block Storage service (cinder), the Object Storage service (swift), or the Bare Metal service (ironic)."

Jenny-Anne: Yes, it looks that way from those guidelines. 

"Developer documentation may refer to the project name, but end-user, operator, administrator, and application developer documentation must refer to the service name. When reviewing service names, consider the consumers of the information."

klgill:
There is no distinction between service names between upstream and downstream. But these variables will help ensure that the names are consistent in both docs.

PR #409, opened 2024-04-17
fix typo then -> them

cubeek:
just a typo

I don't know why the last line is shown as changed, I didn't touch that.


jistr (approved):
Thanks

PR #414, opened 2024-04-18
moved retrieving services module

klgill:
Moved "Retrieving services topology specific-configuration" module to beginning of control plane chapter.


klgill:
@jistr @matbu This is just a quick PR where I moved the "Retrieving services from a topology specific-configuration" module to the beginning of the control plane chapter.

jistr (approved):
:+1: Thank you

PR #421, opened 2024-04-23
DNM:Test new version of pre-adoption-validation

gibizer:
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/634


gibizer:
recheck I updated the dependency

gibizer:
It showed that the new check was run properly

PR #425, opened 2024-04-26
consolidating sections and checking xrefs

klgill:
Minor maintenance of the Adoption guide. I consolidated "Variables" sections with "Prerequisites" and fixed some broken xrefs.

### DIFF
--- a/docs_user/assemblies/assembly_preparing-the-block-storage-service-for-adoption.adoc
+++ b/docs_user/assemblies/assembly_preparing-the-block-storage-service-for-adoption.adoc
@@ -4,9 +4,10 @@
 
 = Preparing the {block_storage} configurations for adoption
 
-As described in xref:planning-the-new-deployment_planning[Planning the new deployment], {block_storage_first_ref} is configured using
+The {block_storage_first_ref} is configured using
 configuration snippets instead of using configuration parameters
-defined by the installer.
+defined by the installer. For more information, see xref:planning-the-new-deployment_planning[Planning the new deployment].
+//kgilliga: Note to self: This xref does not work in the preview. Need to revisit.
 
 The recommended way to deploy {block_storage} volume backends has changed to remove old
 limitations, add flexibility, and improve operations.
--- a/docs_user/modules/proc_adopting-the-compute-service.adoc
+++ b/docs_user/modules/proc_adopting-the-compute-service.adoc
@@ -1,6 +1,6 @@
 [id="adopting-the-compute-service_{context}"]
 
-//Check xref contexts.
+//kgilliga: Note to self: "Adopting the data plane" xrefs do not work. Need to revisit.
 
 = Adopting the {compute_service}
 
@@ -21,15 +21,12 @@ must already be imported into the control plane MariaDB;
  ** the xref:adopting-the-image-service_{context}[Adopting the Image service] needs to be imported;
  ** the xref:migrating-ovn-data_migrating-databases[Migrating OVN data] need to be imported;
  ** the xref:adopting-the-networking-service_{context}[Adopting the Networking service] needs to be imported;
- ** the xref:adopting-the-bare-metal-provisioning-service_{context}[Adopting the Openstack Baremetal service] needs to be imported;
+ ** the xref:adopting-the-bare-metal-provisioning-service_{context}[Adopting the Openstack Baremetal service] needs to be imported; 
+//kgilliga:Need to revist this xref. Might rewrite this section anyway.
  ** Required services specific topology
-xref:pulling-the-openstack-configuration_{context}[Pulling the OpenStack configuration].
-//kgilliga: this xref should specifically point to the Get services topology specific configuration module when it's ready.
+xref:proc_retrieving-services-topology-specific-configuration_{context}[Retrieving services from a topology specific-configuration].
  ** {rhos_prev_long} services have been stopped. For more information, see xref:stopping-openstack-services_migrating-databases[Stopping {rhos_prev_long} services].
-
-.Variables
-
-Define the shell variables and aliases used in the steps below. The values are just illustrative, use values that are correct for your environment:
+* Define the shell variables and aliases used in the steps below. The values are just illustrative, use values that are correct for your environment:
 ----
 alias openstack="oc exec -t openstackclient -- openstack"
 ----
@@ -135,7 +132,7 @@ oc wait --for condition=Ready --timeout=300s Nova/nova
 ----
 +
 The local Conductor services will be started for each cell, while the superconductor runs in `cell0`.
-Note that `disable_compute_service_check_for_ffu` is mandatory for all imported Nova services, until the external data plane is imported, and until Nova Compute services fast-forward upgraded. For more information, see xref:adopting-dataplane_{context}[Data Plane Adoption].
+Note that `disable_compute_service_check_for_ffu` is mandatory for all imported Nova services, until the external data plane is imported, and until Nova Compute services fast-forward upgraded. For more information, see xref:adopting-data-plane_data-plane[Adopting the data plane].
 
 .Verification 
 
@@ -147,7 +144,7 @@ $ openstack endpoint list | grep nova
 $ openstack server list
 ----
 
-Compare the following outputs with the topology specific configuration in xref:pulling-the-openstack-configuration_{context}[Pulling the OpenStack configuration].
+Compare the following outputs with the topology specific configuration in xref:proc_retrieving-services-topology-specific-configuration_{context}[Retrieving services from a topology specific-configuration].
 
 * Query the superconductor for cell1 existance and compare it to pre-adoption values:
 +
@@ -164,4 +161,4 @@ The expected changes to happen:
  ** RabbitMQ transport URL no longer uses `guest`.
 
 [NOTE]
-At this point, the {compute_service} control plane services have yet taken control over existing {compute_service} compute workloads. That would become possible to verify only after the data plane adoption is completed. For more information, see xref:adopting-dataplane_{context}[Data Plane Adoption].
+At this point, the {compute_service} control plane services have yet taken control over existing {compute_service} compute workloads. That would become possible to verify only after the data plane adoption is completed. For more information, see xref:adopting-data-plane_data-plane[Adopting the data plane].
--- a/docs_user/modules/proc_adopting-the-compute-service.adoc
+++ b/docs_user/modules/proc_adopting-the-compute-service.adoc
@@ -26,7 +26,7 @@ must already be imported into the control plane MariaDB;
  ** Required services specific topology
 xref:proc_retrieving-services-topology-specific-configuration_{context}[Retrieving services from a topology specific-configuration].
  ** {rhos_prev_long} services have been stopped. For more information, see xref:stopping-openstack-services_migrating-databases[Stopping {rhos_prev_long} services].
-* Define the shell variables and aliases used in the steps below. The values are just illustrative, use values that are correct for your environment:
+* Define the following shell variables. The values that are used are examples. Replace these example values with values that are correct for your environment:
 ----
 alias openstack="oc exec -t openstackclient -- openstack"
 ----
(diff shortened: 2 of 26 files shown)


klgill:
@jistr @matbu  I have a question about the “Stopping OpenStack services” module. This procedure has the following prerequisite:
“Also collect the services topology specific configuration before stopping services required to gather it live. You will need it to compare it with the post-adoption values later on. For more information, see xref:proc_retrieving-services-topology-specific-configuration_adopt-control-plane[Retrieving services from a topology specific-configuration].”

However, the “Retrieving services from a topology specific-configuration” module comes *after* the “Stopping OpenStack services” module. Should the “Retrieving services from a topology specific-configuration” module be moved before “Stopping the OpenStack services”?

docs_user/assemblies/assembly_preparing-the-block-storage-service-for-adoption.adoc
```diff
@@ -4,9 +4,10 @@
 
 = Preparing the {block_storage} configurations for adoption
 
-As described in xref:planning-the-new-deployment_planning[Planning the new deployment], {block_storage_first_ref} is configured using
+The {block_storage_first_ref} is configured using
 configuration snippets instead of using configuration parameters
-defined by the installer.
+defined by the installer. For more information, see xref:planning-the-new-deployment_planning[Planning the new deployment].
```
gregraka, line 9: I notice in the xrefs, you have removed the context variable. Is that because the target resides in another assembly?

docs_user/modules/proc_adopting-the-compute-service.adoc
```diff
@@ -21,15 +21,12 @@ must already be imported into the control plane MariaDB;
...
  ** Required services specific topology
-xref:pulling-the-openstack-configuration_{context}[Pulling the OpenStack configuration].
-//kgilliga: this xref should specifically point to the Get services topology specific configuration module when it's ready.
+xref:proc_retrieving-services-topology-specific-configuration_{context}[Retrieving services from a topology specific-configuration].
  ** {rhos_prev_long} services have been stopped. For more information, see xref:stopping-openstack-services_migrating-databases[Stopping {rhos_prev_long} services].
-
-.Variables
-
-Define the shell variables and aliases used in the steps below. The values are just illustrative, use values that are correct for your environment:
+* Define the shell variables and aliases used in the steps below. The values are just illustrative, use values that are correct for your environment:
```
gregraka, line 29: Per the ISG:
Do not use [below] to refer to an earlier software version or fix level. Use "earlier". Do not use to indicate a relative location in a document, as in "the information below". Use "the information later in this section". Alternatively, you can use "following" to refer to a table, to art, or to a list. Example: "The following table shows ..."

docs_user/modules/proc_adopting-the-compute-service.adoc
```diff
@@ -164,4 +161,4 @@ The expected changes to happen:
  ** RabbitMQ transport URL no longer uses `guest`.
 
 [NOTE]
-At this point, the {compute_service} control plane services have yet taken control over existing {compute_service} compute workloads. That would become possible to verify only after the data plane adoption is completed. For more information, see xref:adopting-dataplane_{context}[Data Plane Adoption].
+At this point, the {compute_service} control plane services have yet taken control over existing {compute_service} compute workloads. That would become possible to verify only after the data plane adoption is completed. For more information, see xref:adopting-data-plane_data-plane[Adopting the data plane].
```
gregraka, line 164: These 2 sentences read strangely.
Consider this alternative:

````
At this point, the {compute_service} control plane services do not control the existing {compute_service} compute workloads. The control plane manages the data plane only after the successful completion of the data adoption process. For more information, see xref:adopting-data-plane_data-plane[Adopting the data plane].
````

File: docs_user/modules/proc_configuring-a-ceph-backend.adoc [hunk 4]
```diff
@@ -33,11 +33,7 @@ ceph auth caps client.openstack \
 .Prerequisites
 
 * The `OpenStackControlPlane` custom resource (CR) must already exist.
-
-.Variables
-
-Define the shell variables used in the steps below. The values are
-just illustrative, use values that are correct for your environment:
+* Define the following shell variables. The values are illustrative. Use values that are correct for your environment:
```
gregraka, line 36: I wonder if `illustrative` might challenge a non-native English speaker? Not sure, but I think we target a 7-8th grade reading level in our docs.

Maybe this?

````
Define the following shell variables. The values used here are examples. Replace these example values with values that are correct for your environment:
````

File: docs_user/modules/proc_deploying-the-bare-metal-provisioning-service.adoc [hunk 5]
```diff
@@ -27,11 +27,8 @@ It is critical that this configuration file comes from one of the controllers an
 //kgilliga: What is meant by "overcloud Ironic deployment? Can this be changed to the "RHOSP Bare Metal Provisioning service deployment"?
 
 * If adopting the Ironic Inspector service you need the value of the `IronicInspectorSubnets` {OpenStackPreviousInstaller} parameter. Use the same values to populate the `dhcpRanges` parameter in the target environment.
-
-.Variables
-
-Define the shell variables and aliases used in the steps below. The values are just illustrative, use values that are correct for your environment:
-
+* Define the following shell variables and aliases. The values are illustrative.  Use values that are correct for your environment:
```
gregraka, line 30: See earlier comment about this construction.

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
```diff
@@ -23,12 +23,8 @@ here this time).
 //kgilliga: this xref should specifically point to the Get services topology specific configuration module when it's ready.
  ** {OpenStackShort} services have been stopped. For more information, see xref:stopping-openstack-services_{context}[Stopping {rhos_prev_long} services].
  ** There must be network routability between the original MariaDB and the MariaDB for the control plane.
-
-.Variables
-
-Define the shell variables used in the steps below. The values are
-just illustrative, use values that are correct for your environment:
-
+* Define the following shell variables. The values are illustrative. Use values that are correct for your environment:
```
gregraka, line 26: Another instance.

docs_user/modules/proc_migrating-ovn-data.adoc
```diff
@@ -18,12 +18,8 @@ defined. Specifically, `openstack/internalapi` network is defined.
  ** There must be network routability between:
   *** The adoption host and the original OVN.
   *** The adoption host and the control plane OVN.
-
-.Variables
-
-Define the shell variables used in the steps below. The values are
-just illustrative, use values that are correct for your environment:
-
+* Define the following shell variables. The values are illustrative. Use values that are correct for your environment:
```
gregraka, line 21: Another instance.

docs_user/modules/proc_migrating-ovn-data.adoc
```diff
@@ -34,9 +30,9 @@ OVSDB_IMAGE=registry.redhat.io/rhosp-dev-preview/openstack-ovn-base-rhel9:18.0
 endif::[]
 SOURCE_OVSDB_IP=172.17.1.49
 ----
-
-The real value of the `SOURCE_OVSDB_IP` can be get from the puppet generated configs:
-
++
+You can get the real value of the `SOURCE_OVSDB_IP` from the puppet generated configurations:
```
gregraka, line 34: `real` doesn't seem like the correct term to use unless it has a nuance that I'm unaware of.

How about?

````
You can get the value to set `SOURCE_OVSDB_IP` by querying the puppet-generated configurations:
````

File: docs_user/modules/proc_stopping-openstack-services.adoc [hunk 9]
```diff
@@ -25,26 +25,21 @@ Note that you should not stop the infrastructure management services yet, such a
 
 .Prerequisites
 
-You can stop {OpenStackShort} services at any moment, but you might leave your environment in an undesired state. However, you should confirm that there are no long running operations that require other services.
-
-Ensure that there are no ongoing instance live migrations, volume migrations (online or offline), volume creation, backup restore, attaching, detaching,
+* You can stop {OpenStackShort} services at any moment, but you might leave your environment in an undesired state. However, you should confirm that there are no long running operations that require other services.
```
gregraka, line 28: Do we really want to document this content? If there's a risk of an "undesirable state" I wonder if this is wise?

- `moment` seems awkward to me. Maybe `time` instead? 
- Per ISG:
````
'Do not use [etc.]. Use "and so on" when you list a clear sequence of
elements, such as "1, 2, 3, and so on"....'
````
If you do decide to document this, maybe change it to something like this?
````
Before you stop {OpenStackShort} services, first confirm that there no long 
running operations that require these services.
WARNING: Avoid stopping {OpenStackShort} services. Doing so can leave
your environment in an undesired state.
````

File: docs_user/modules/proc_stopping-openstack-services.adoc [hunk 10]
```diff
@@ -25,26 +25,21 @@ Note that you should not stop the infrastructure management services yet, such a
...
+* Ensure that there are no ongoing instance live migrations, volume migrations (online or offline), volume creation, backup restore, attaching, detaching,
```
gregraka, line 29: Change `etc.` to `and so on`.

docs_user/modules/proc_stopping-openstack-services.adoc
```diff
@@ -25,26 +25,21 @@ Note that you should not stop the infrastructure management services yet, such a
...
 ----
-
-Also collect the services topology specific configuration before stopping services required to gather it live. You will need it to compare it with the post-adoption values later on. For more information, see xref:pulling-the-openstack-configuration_{context}[Pulling the OpenStack configuration].
-//kgilliga: this xref should specifically point to the Get services topology specific configuration module when it's ready.
-
-.Variables
-
-Define the shell variables used in the following steps. The values are illustrative and refer to a single node standalone {OpenStackPreviousInstaller} deployment. Use values that are correct for your environment:
-
+* Also collect the services topology specific configuration before stopping services required to gather it live. You will need it to compare it with the post-adoption values later on. For more information, see xref:proc_retrieving-services-topology-specific-configuration_adopt-control-plane[Retrieving services from a topology specific-configuration].
```
gregraka, line 39: - Comma after `Also`?
- Hyphen after `topology`?
- It's unclear to me what `required to gather it live` means.
   Sounds jargony. Maybe the jargon has a specific meaning?

docs_user/modules/proc_stopping-openstack-services.adoc
```diff
@@ -25,26 +25,21 @@ Note that you should not stop the infrastructure management services yet, such a
...
+//kgilliga: Question for Mathieu and Jiri: Should the "Retrieving services from a topology specific-configuration" module be moved to the beginning of the "Migrating databases to the control plane chapter?".
+* Define the following shell variables. The values are illustrative and refer to a single node standalone {OpenStackPreviousInstaller} deployment. Use values that are correct for your environment:
```
gregraka, line 41: Another instance of `illustrative` if you decide to change it.

docs_user/assemblies/assembly_preparing-the-block-storage-service-for-adoption.adoc
klgill: I used "{context}" when the target was in the same assembly. If the target was in a different assembly, I used the actual context variable, e.g., "planning" or "migrating databases", etc.

jistr:
@bogdando Regarding @klgill 's question above, the only thing i see that does something else than talk to MariaDB is this:
`$CONTROLLER_SSH sudo podman exec -it nova_api nova-manage cell_v2 list_cells`. Can this run nova-manage call be executed without the Nova services running? If yes then i'd say that rather than moving that whole section, we should just remove the prerequisite from the "stopping services" doc instead. I think we want to fetch the values for those config variables with the API services already stopped, to make sure nothing is changing?

klgill:
> @bogdando Regarding @klgill 's question above, the only thing i see that does something else than talk to MariaDB is this: `$CONTROLLER_SSH sudo podman exec -it nova_api nova-manage cell_v2 list_cells`. Can this nova-manage call be executed without the Nova services running? If yes then i'd say that rather than moving that whole section, we should just remove the prerequisite from the "stopping services" doc instead. I think we want to fetch the values for those config variables with the API services already stopped, to make sure nothing is changing?

I think this edit can be made in a separate PR after @bogdando answers the question. I want to merge this and continue checking cross-references and other links.

bogdando:
> @bogdando Regarding @klgill 's question above, the only thing i see that does something else than talk to MariaDB is this: `$CONTROLLER_SSH sudo podman exec -it nova_api nova-manage cell_v2 list_cells`. Can this nova-manage call be executed without the Nova services running? If yes then i'd say that rather than moving that whole section, we should just remove the prerequisite from the "stopping services" doc instead. I think we want to fetch the values for those config variables with the API services already stopped, to make sure nothing is changing?

Nova manage commands will not function on the source cloud when there is no Nova control planning services running

PR #431, opened 2024-05-01
Remove ovn_match_northd_version from samples

booxter:
`false` is the default value; and we are not relying on this ovn "feature" when upgrading ovs. (it is needed only when upgrading using so-called `fail-safe upgrade` procedure [1], which means upgrading skipping an LTS release. This is not what we do anyway.)

[1] https://docs.ovn.org/en/stable/intro/install/ovn-upgrades.html#fail-safe-upgrade


booxter:
Looks like I'm wrong saying the default is `false`. (it's `true` in edpm-ansible) Need to understand why it's `true` there; I don't think we (want to) do fail-safe upgrades. @karelyatin thoughts?

booxter:
Looks like we should change the defaults in edpm-ansible first to correct one (`false`): https://opendev.org/openstack/puppet-tripleo/commit/68b4a7c87d3af1200d18149365844d40683cb530?style=unified&whitespace=ignore-all&show-outdated=

Then we can clean this up here and in other samples.

booxter:
switch for the default value in edpm-ansible here: https://github.com/openstack-k8s-operators/edpm-ansible/pull/651

booxter:
May need a follow-up after https://github.com/openstack-k8s-operators/data-plane-adoption/pull/475 merges.

booxter:
recheck

```
Fetching gem metadata from https://rubygems.org/..
Retrying fetcher due to error (2/4): Errno::EPIPE Broken pipe - SSL_connect
.
Retrying fetcher due to error (3/4): Errno::EPIPE Broken pipe - SSL_connect
..
Retrying fetcher due to error (4/4): Errno::EPIPE Broken pipe - SSL_connect
..
cd docs_user; BUILD=upstream make html
make[1]: Entering directory '/home/zuul/src/github.com/openstack-k8s-operators/data-plane-adoption/docs_user'
cp images/.gitkeep ../docs_build/adoption-user/images
touch ../docs_build/adoption-user/.timestamp-images
bundle exec asciidoctor -a source-highlighter=highlightjs -a highlightjs-languages="yaml,bash" -a highlightjs-theme="monokai" --failure-level WARN -a build=upstream -b xhtml5 -d book -o ../docs_build/adoption-user/index-upstream.html main.adoc
bundler: command not found: asciidoctor
Install missing gem executables with `bundle install`
```

jistr (approved):

PR #441, opened 2024-05-06
Add libvirt customIssuer

xek:
libvirt CA was recently introduced, a customIssuer needs to be
set to overwrite the defaults with the IPA extracted CA.


vakwetu:
+2

PR #446, opened 2024-05-08
Cinder: Downstream only describe single backend

Akrog:
When adopting the Cinder volume service in an openstack-operator deployment we can include all backends in a single pod or having a pod for each backend.

This is a system administrator preference and both options will work, but we consider that the 1 backend per pod option is considerably superior, so for the downstream documentation we want it to be the only option.

This patch makes a couple of small conditional changes so that the quick and dirty solution of creating the cinder configuration is not mentioned downstream.


jistr (approved):
LGTM but will let @klgill review & merge

klgill (approved):

PR #447, opened 2024-05-08
[doc]Describe what to do when pre-adoption-validation fails

gibizer:
Implements: OSPRH-6875

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -96,7 +96,7 @@ export computes=(
 .Procedure
 
 //* _Temporary fix_ until the OSP 17 https://code.engineering.redhat.com/gerrit/q/topic:stable-compute-uuid[backport of the stable compute UUID feature]
-//lands. 
+//lands.
 //kgilliga: Revisit this step after 17.1.3 is on the CDN.
 . For each Compute node, write the UUID of the Compute service to the stable `compute_id` file in `/var/lib/nova/` directory:
 +
@@ -121,11 +121,11 @@ endif::[]
 done
 ----
 ifeval::["{build}" != "downstream"]
-. Create a https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets[ssh authentication secret] for the data plane nodes: 
+. Create a https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets[ssh authentication secret] for the data plane nodes:
 //kgilliga:I need to check if we will document this in Red Hat docs.
 endif::[]
 ifeval::["{build}" != "upstream"]
-. Create a ssh authentication secret for the data plane nodes: 
+. Create a ssh authentication secret for the data plane nodes:
 endif::[]
 +
 [subs=+quotes]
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -613,6 +613,21 @@ oc logs -l app=openstackansibleee -f --max-log-requests 20
 oc wait --for condition=Ready openstackdataplanedeployment/openstack-pre-adoption --timeout=10m
 ----
 
+. If the validation deployment failed then check the ansible execution logs to
```
marios, line 616: if any openstack-pre-adoption validations fail, you must first determine which ones were unsuccessful and then follow the instructions below for each case:

(just a suggestion feel free to adopt part/all/none)

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -613,6 +613,21 @@ oc logs -l app=openstackansibleee -f --max-log-requests 20
...
+* if the hostname validation failed then check that the hostname of the EDPM
+node is correctly listed in the `OpenStackDataPlaneNodeSet`
+
+* if the kernel argument check failed then make sure that the
+`OpenStackDataPlaneNodeSet` has the same kernel argument configuration in
+`edpm_kernel_args` and `edpm_kernel_hugepages` variables than what is used in
+the 17 node.
+
+* if the tuned profile check failed then make sure that the
+`edpm_tuned_profile` variable in the `OpenStackDataPlaneNodeSet` configured to
```
marios, line 628: is configured to

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -613,6 +613,21 @@ oc logs -l app=openstackansibleee -f --max-log-requests 20
...
+use the same profile as it used on the 17 node.
```
marios, line 629: same profile as set on the (source) OSP 17 node.

marios (commented):
thanks for posting it - just some inline suggestions but overall lgtm

wondering if we need to tell them more about 'how to determine which validation failed'

gibizer:
> thanks for posting it - just some inline suggestions but overall lgtm
> 
> wondering if we need to tell them more about 'how to determine which validation failed'

How to get the ansible logs is describe just above this. How to understand ansible logs can be an interesting doc topic, but I think it is out of scope of this patch as I think it requires a more general approach.

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
gibizer: applied your suggestion with a small tweak

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
gibizer: Done

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
gibizer: Applied

jistr (approved):

PR #456, opened 2024-05-13
Add mariadb-copy-data clean up.

holser:
When adoption fails on mariadb-copy-data it needs to be cleaned up.


jistr (approved):
Thanks :+1:

PR #459, opened 2024-05-14
Test new reboot strategy against dataplane-adoption validation

ciecierski:
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/646

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -441,6 +441,20 @@
 - name: Run pre-adoption validation
   when: run_pre_adoption_validation|bool
   block:
+    - name: Create OpenStackDataPlaneService/reboot-strategy
+      no_log: "{{ use_no_log }}"
+      ansible.builtin.shell: |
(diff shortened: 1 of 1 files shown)


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -470,6 +484,7 @@
           - openstack
           servicesOverride:
           - pre-adoption-validation
+          - reboot-strategy
```
jistr, line 487: I don't think we should be diverging from the docs like this. Let's just put the os-reboot role back into the service list on the NodeSet, both in tests and in docs?

jistr (changes requested):

tests/roles/dataplane_adoption/tasks/main.yaml
ciecierski: ack, I will put it back into pre-validation

ciecierski:
Closed in favor of https://github.com/openstack-k8s-operators/data-plane-adoption/pull/459

PR #462, opened 2024-05-15
Externalize a TripleO deployed Ceph cluster

fmount:
Implements: [OSPRH-832](https://issues.redhat.com/browse/OSPRH-832)
Recording: https://asciinema.org/a/659587


fmount:
@jistr this playbook is not executed by default, and not tied to our CI atm, so in general, if this patch meets the criteria and requirements we have for the project, I'd like you to review it and see if it's reasonable to get it merged.
I tested this code in a 3+3 environment, and there's a recording of the entire execution [1].
As a follow up we might need to refine something in the code to have this playbook triggered on a topology that @cescgina and @marios are working on, but I'm still not sure we want to integrate it in the current gating jobs (it probably makes more sense downstream).
Can we get some reviews and check if we can move forward with this work to unblock the next stage (integration)?

[1] https://asciinema.org/a/659587

fultonj:
I'm +2 on merging this playbook.

tests/roles/ceph_migrate/README.md
```diff
@@ -0,0 +1,143 @@
...
+intended to be run in CI, where a `multinode` TripleO environment with deployed
+Ceph exists.
+It **does not** migrate OSDs, hence no data movement is involved.
+It assumes that both `source` and `target` nodes are part of the Ceph cluster,
+and the high level procedure does the following:
+
+1. It Loads the information from the existing Ceph cluster (deployed daemons,
+   status of the cluster, hostmap, services, monmap, osdmap)
+2. It Validates the retrieved information (the cluster is healthy and it's in a
+   valid state), and fail is one of the defined conditions is not met
```
marios, line 14: fail if one

tests/roles/ceph_migrate/README.md
```diff
@@ -0,0 +1,143 @@
...
+perform (as an example) the migration of a single node using tags and passing
+overrides.
+To migrate only a subset of Mon/Mgr, it is possible to customize the list of
+`decomm_nodes` and `target_nodes`:
+
+```
+decomm_nodes:
+  - controller-0
+target_nodes:
+  - cephstorage-0
```
marios, line 38: who/what provides this node? I mean, the tripleo/source deployment will have 3 control 3 compute.

tests/roles/ceph_migrate/tasks/ceph_cli.yaml
```diff
@@ -0,0 +1,33 @@
+---
+# Copyright 2021 Red Hat, Inc.
```
marios, line 2: 2021

marios (commented):
first pass

tests/roles/ceph_migrate/README.md
fmount: Done

tests/roles/ceph_migrate/tasks/ceph_cli.yaml
fmount: Done

tests/roles/ceph_migrate/README.md
fmount: I added an additional sentence to clarify how, in the example, the `hostname(s)` of the nodes are gathered. In particular, the example shows how to pass overrides to decommission a single node, moving daemons (mons in particular) from src to the dest. This might be useful considering that we do not have a `3+3` right now, but we might want to try a single `mon` migration to prove that the flow is safe.

fmount: @marios as a follow up of this work we need to plug this playbook to our CI, which is where we might need a "glue" code that generates the overrides and runs the playbook. Given that we know in advance the topology, and the resulting hosts data is predictable in CI, this might just be something we pass from a config job.

tests/roles/ceph_migrate/README.md
```diff
@@ -0,0 +1,149 @@
...
+- `inventory` should be replaced by the inventory used to connect to the TripleO
+  deployed nodes.
+
+It is possible to execute the migration of a specific Ceph daemon type using the
+tags.
+For example, the RGW migration only can be triggered with the following:
+
+```
+ansible-playbook -i <inventory> $PLAYBOOK_DIR/test_externalize_ceph.yaml -e @overrides.yaml --tags ceph_rgw
+```
```
jistr, line 149: Nice doc :+1:

jistr (approved):

tests/plugins/module_utils/ca_common.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/python
+
+# -*- coding: utf-8 -*-
+# Copyright (c) 2020 OpenStack Foundation
```
marios, line 4: so is this a copy from an earlier file? do we need to point to it? do we need/want to add also 2024 here in an earlier line? There are a few more not just this one.

tests/roles/ceph_migrate/README.md
marios: ack lets talk revisit this when we are ready. Joan is working through the ceph wire up (for 3ctrl/1compute) currently so that is a prerequisite

marios (commented):
as discussed in yesterday's adoption call, this should be safe and iterate. It is not yet wired up to run in any of the make targets.

tests/plugins/module_utils/ca_common.py
fmount: Yes, this set of python files represent the core library imported from TripleO: tripleo used to manipulate the ceph spec, which is something required in this context.
I added that library here and extended to fit the use case. I'll fix the header in my next PS

fmount: I updated the headers at this point, thank you for checking!

fmount:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

fmount:
recheck

cescgina (approved):

PR #466, opened 2024-05-22
Fix Mariadb_address in new documentation

holser:
There were doc update after PR#458. This is follow up update to update this part of documentation.

### DIFF
--- a/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
+++ b/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
@@ -116,7 +116,7 @@ oc wait --for condition=Ready pod/mariadb-copy-data --timeout=30s
 
 .Procedure
 
-. Check that the Galera database cluster members are online and synced:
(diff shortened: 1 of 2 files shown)


docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
```diff
@@ -116,7 +116,7 @@ oc wait --for condition=Ready pod/mariadb-copy-data --timeout=30s
 
 .Procedure
 
-. Check that the Galera database cluster members are online and synced:
+. Check that the Source galera database cluster members are online and synced:
```
jistr, line 119: I think this should be capitalized differently: "source Galera". It's a small thing but now that these docs directly translate to customer docs, we need to care about such stuff :).

holser: updated in next update

jistr (approved):
:+1: Thanks

jistr:
recheck

PR #475, opened 2024-05-29
Document OVN gateway adoption procedure

booxter:
Basically a truncated repeat of Compute procedure, with an emphasis on the list of services to include and on the fact that a separate NodeSet and Deployment is needed.

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -159,7 +159,7 @@ rm -f id*
 cd -
 ----
 
-. Create a `nova-compute-extra-config` service (with local storage backend for lbivrt):
+. Create a `nova-compute-extra-config` service (with local storage backend for `libvirt`):
 +
 . If TLS Everywhere is enabled, append the following to the OpenStackDataPlaneService spec:
 +
@@ -211,7 +211,7 @@ The secret `nova-cell<X>-compute-config` is auto-generated for each
 That service removes pre-FFU workarounds and configures Compute
 services for local storage backend.
 
-* Or, create a `nova-compute-extra-config` service (with Ceph backend for libvirt):
+* Or, create a `nova-compute-extra-config` service (with Ceph backend for `libvirt`):
 +
 [source,yaml]
 ----
@@ -361,7 +361,7 @@ endif::[]
           nova_libvirt_network: internalapi
 
         # edpm_network_config
-        # Default nic config template for a EDPM compute node
+        # Default nic config template for a EDPM node
         # These vars are edpm_network_config role vars
         edpm_network_config_template: |
            ---
@@ -505,10 +505,10 @@ spec:
         readOnly: true
 "
 ----
-+
-* Replace `<bridge_mappings>` with the value of the bridge mappings in your configuration, for example, `"datacentre:br-ctlplane"`.
+Note that you should retain the original `OpenStackDataPlaneNodeSet` services
+composition, except the inserted `ceph-client` service.
 
-. Ensure that the `ovn-controller` settings that are configured in the `OpenStackDataPlaneNodeSet` CR are the same as were set in the Compute nodes before adoption. This configuration is stored in the `external_ids`` column in the `Open_vSwitch` table in the Open vSwitch database:
++ Ensure that the `ovn-controller` settings that are configured in the `OpenStackDataPlaneNodeSet` CR are the same as were set in the Compute nodes before adoption. This configuration is stored in the `external_ids` column in the `Open_vSwitch` table in the Open vSwitch database:
 +
 ----
 ovs-vsctl list Open .
@@ -516,9 +516,6 @@ ovs-vsctl list Open .
 external_ids        : {hostname=standalone.localdomain, ovn-bridge=br-int, ovn-bridge-mappings=<bridge_mappings>, ovn-chassis-mac-mappings="datacentre:1e:0a:bb:e6:7c:ad", ovn-encap-ip="172.19.0.100", ovn-encap-tos="0", ovn-encap-type=geneve, ovn-match-northd-version=False, ovn-monitor-all=True, ovn-ofctrl-wait-before-clear="8000", ovn-openflow-probe-interval="60", ovn-remote="tcp:ovsdbserver-sb.openstack.svc:6642", ovn-remote-probe-interval="60000", rundir="/var/run/openvswitch", system-id="2eec68e6-aa21-4c95-a868-31aeafc11736"}
 ...
 ----
-
-Note that you should retain the original `OpenStackDataPlaneNodeSet` services
-composition, except the inserted `ceph-client` service.
 
 +
 * Replace `<bridge_mappings>` with the value of the bridge mappings in your configuration, for example, `"datacentre:br-ctlplane"`.
@@ -678,6 +675,13 @@ spec:
   - openstack
 EOF
 ----
+
+Note: if you have other node sets to deploy (e.g. for networker nodes), you may
+add them above in `nodeSets` list, or create separate
+`OpenStackDataPlaneDeployment` CRs later.
+
+Note that once a `OpenStackDataPlaneDeployment` CR is deployed, it's impossible
+to add new node sets to it later.
 
 .Verification
 
--- a/docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
@@ -189,6 +189,8 @@ ifeval::["{build}" != "downstream"]
           # https://issues.redhat.com/browse/RHOSZUUL-1517
           dnf -y install crypto-policies
           update-crypto-policies --set FIPS:NO-ENFORCE-EMS
+          ./venv/bin/repo-setup current-podified -b antelope -d centos9 --stream
+          rm -rf repo-setup-main
 endif::[]
 ifeval::["{build}" == "downstream"]
         edpm_bootstrap_command: |
--- a/docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
@@ -1,0 +1,346 @@
+[id="adopting-networker-services-to-the-data-plane_{context}"]
+
+= Adopting Networker services to the {rhos_acro} data plane
+
+If you have `Networker` nodes in your existing {rhos_acro} deployment, you
+should adopt them to the {rhos_acro} data plane too. The procedure is largely
+the same as for `Compute` nodes.
+
+The main difference between `Networker` and `Compute` nodes is that `Networker`
+nodes do not run `nova-compute` and `libvirt` services.
+
+You won't need to run `neutron-sriov` service on `Networker` nodes either. But
+depending on your topology, you might need to run `neutron-metadata` service
+there. Specifically, when you'd like to serve metadata to SR-IOV ports hosted
+on `Compute` nodes.
+
+Assuming you'd like to continue running OVN gateway services on `Networker`
+nodes, you will keep `ovn` service in the list to deploy. (You may later decide
+to deprovision OVN gateway services from `Networker` nodes and run them on OCP
+nodes instead. This is currently not recommended and is out of scope for the
+adoption procedure.)
+
+You may also want to run `neutron-dhcp` service on `Networker` nodes instead of
+`Compute` nodes. (But note that using `neutron-dhcp` with OVN is not generally
+needed.)
+
+Whatever the final list of services you'd like to run on `Networker` nodes, you
+will create a separate `OpenStackDataPlaneNodeSet` CR just for these nodes.
+
+. Define the following shell variable. The value that is used is an example
+and refers to a single node standalone {OpenStackPreviousInstaller} deployment.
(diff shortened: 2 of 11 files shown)


docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
```diff
@@ -0,0 +1,349 @@
...
+oc logs -l app=openstackansibleee -f --max-log-requests 20
+----
+
+.. Wait for the deployment to reach the `Ready` status:
++
+----
+oc wait --for condition=Ready openstackdataplanedeployment/openstack-pre-adoption-networker --timeout=10m
+----
+
+. Deploy the `OpenStackDataPlaneDeployment` CR for `Networker` nodes:
```
slawqo, line 302: do we need new Deployment CR for that? What if we would add new NodeSet to the existing Deployment CR?

slawqo:
recheck - patch looks good for me and I am not sure if the failed job is related to this change really

docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
```diff
@@ -0,0 +1,349 @@
...
+vary depending on the services you've enabled):
++
+----
+oc exec openstackclient -- openstack network agent list
++--------------------------------------+------------------------------+------------------------+-------------------+-------+-------+----------------------------+
+| ID                                   | Agent Type                   | Host                   | Availability Zone | Alive | State | Binary                     |
++--------------------------------------+------------------------------+------------------------+-------------------+-------+-------+----------------------------+
+| 174fc099-5cc9-4348-b8fc-59ed44fcfb0e | DHCP agent                   | standalone.localdomain | nova              | :-)   | UP    | neutron-dhcp-agent         |
+| 10482583-2130-5b0d-958f-3430da21b929 | OVN Metadata agent           | standalone.localdomain |                   | :-)   | UP    | neutron-ovn-metadata-agent |
+| a4f1b584-16f1-4937-b2b0-28102a3f6eaa | OVN Controller agent         | standalone.localdomain |                   | :-)   | UP    | ovn-controller             |
```
karelyatin, line 347: the name i think will change to OVN Controller Gateway agent when chassis as gw=true, but can cross check and update here

docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
booxter: Yeah, I guess I went with simpler separation of two topics (compute vs networker) though both steps should be aligned better. I think we may want to document both. There are benefits with a separate Deployment (failure domain reduction), but it's helpful to demonstrate that the whole deployment can be done in a single step too.

booxter: I am adding a note to both sections of the document that you can either deploy separately, or reuse the same CR (plus a note that patching after deployment won't work.)

docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
booxter: Indeed, it's `OVN Controller Gateway agent`.

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -159,7 +159,7 @@ rm -f id*
 cd -
```
jistr, line 159: Changes here LGTM but also CCing @gibizer @bogdando .

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -504,10 +517,10 @@ spec:
         readOnly: true
 "
 ----
-+
-* Replace `<bridge_mappings>` with the value of the bridge mappings in your configuration, for example, `"datacentre:br-ctlplane"`.
```
jistr, line 508: Should we keep the note about bridge mappings?

jistr (commented):
LGTM just one question inline, it could perhaps also be addressed in a separate PR if this one passes CI.

jistr:
recheck

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
gibizer: Look OK to me too

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -205,13 +205,26 @@ spec:
 EOF
 ----
 +
+. If TLS Everywhere is enabled, append the following to the `OpenStackDataPlaneService` spec:
++
+[source,yaml]
+----
+  tlsCert:
```
bogdando, line 212: @gibi @slagle I haven't tried TLS-ed adoption yet, but don't we need to add some sesrvice name nova reference?

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
bogdando: indeed, looks like it got lost during copy-pasta?

docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
```diff
@@ -0,0 +1,356 @@
...
+          rm -f /run/virtlogd.pid
+          rm -rf repo-setup-main
+endif::[]
+ifeval::["{build}" == "downstream"]
+        edpm_bootstrap_command: |
+          subscription-manager register --username {{ subscription_manager_username }} --password {{ subscription_manager_password }}
+          subscription-manager release --set=9.2
+          subscription-manager repos --disable=*
+          subscription-manager repos --enable=rhel-9-for-x86_64-baseos-eus-rpms --enable=rhel-9-for-x86_64-appstream-eus-rpms --enable=rhel-9-for-x86_64-highavailability-eus-rpms --enable=openstack-17.1-for-rhel-9-x86_64-rpms --enable=fast-datapath-for-rhel-9-x86_64-rpms --enable=openstack-dev-preview-for-rhel-9-x86_64-rpms
+          # FIXME: perform dnf upgrade for other packages in EDPM ansible
```
bogdando, line 205: if we are not going to run libvirt on networkers, we need no this hack

docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
```diff
@@ -0,0 +1,356 @@
...
++
+----
+ovs-vsctl list Open .
+...
+external_ids        : {hostname=standalone.localdomain, ovn-bridge=br-int, ovn-bridge-mappings=<bridge_mappings>, ovn-chassis-mac-mappings="datacentre:1e:0a:bb:e6:7c:ad", ovn-cms-options=enable-chassis-as-gw, ovn-encap-ip="172.19.0.100", ovn-encap-tos="0", ovn-encap-type=geneve, ovn-match-northd-version=False, ovn-monitor-all=True, ovn-ofctrl-wait-before-clear="8000", ovn-openflow-probe-interval="60", ovn-remote="tcp:ovsdbserver-sb.openstack.svc:6642", ovn-remote-probe-interval="60000", rundir="/var/run/openvswitch", system-id="2eec68e6-aa21-4c95-a868-31aeafc11736"}
+...
+----
+
++
+* Replace `<bridge_mappings>` with the value of the bridge mappings in your configuration, for example, `"datacentre:br-ctlplane"`.
```
bogdando, line 236: ah, I see it has been moved here!

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
bogdando: or no, it was moved https://github.com/openstack-k8s-operators/data-plane-adoption/pull/475/files#diff-603a706779fa17d06015eb6775886bdaa06a6531c6dc1a2a4ee2edbb4f5e6c66R236

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -256,9 +262,12 @@
             os_net_config_iface: {{ dataplane_os_net_config_iface | default ('nic1') }}
             os_net_config_set_route: {{ dataplane_os_net_config_set_route | default(true) | bool }}
             os_net_config_dns: {{ dataplane_os_net_config_dns | default("") }}
+
+            {%+ if compute_adoption|bool +%}
             service_net_map:
```
bogdando, line 267: those are no longer needed and being removed in edpm

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -356,8 +368,10 @@
         - run-os
         - install-certs
         - ceph-client
+        {%+ if compute_adoption|bool +%}
```
bogdando, line 371: this should be somehow aligned with docs, maybe based on ENV vars, or if-else sections?

tests/vars.sample.yaml
```diff
@@ -49,3 +49,6 @@ ironic_adoption: false
 
 # Run pre-adoption validation before the deploying
 run_pre_adoption_validation: true
+
+# Whether the adopted node will host compute services
+compute_adoption: true
```
bogdando, line 54: what are the cases when we have that set to False?

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
bogdando: FTR https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/services/dataplane_v1beta1_openstackdataplaneservice_libvirt.yaml (doesn't mention any service = nova spec values, so perhaps this is fine

bogdando: see also https://github.com/openstack-k8s-operators/architecture/blob/39f6b870a53bbe13c9ca6ca7e14868232a490e59/lib/dataplane/nodeset/kustomization.yaml#L11

FWIW, perhaps we should try to cover TLS-e support for compute services adoption in this patch, this looks like a major gap to address

booxter: This is just a mistake after a git conflict resolution, this text is also listed above in lines 164-177. I will remove it. Thanks for spotting.

docs_user/modules/proc_adopting-networker-services-to-the-data-plane.adoc
booxter: also above, correct? (lines 192-197).

tests/roles/dataplane_adoption/tasks/main.yaml
booxter: @bogdando could you please clarify what you mean? The docs have a separate section for Networker nodes, and this section doesn't list the libvirt and nova-compute* services in the example snippet. Anything else that I am missing?

tests/vars.sample.yaml
booxter: @bogdando one can set it to `false` and validate that `test-minimal` target still passes without deploying `libvirt` and `nova-compute`. Does it address your concern?

tests/roles/dataplane_adoption/tasks/main.yaml
booxter: This would belong to a separate PR I think.

tests/roles/dataplane_adoption/tasks/main.yaml
bogdando: There is a rule that code in this repo should look the same way in docs. @jistr WDYT for this case, how should we proceed?

booxter:
recheck

booxter:
recheck

booxter:
recheck

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -341,8 +353,10 @@
         - run-os
         - install-certs
         - ceph-client
```
karelyatin, line 355: seems we would also  not need ceph-client on networkers, any reason to keep it outside of if block below?

jistr (approved):
Good comments by Yatin but if you folks want to land this and then do the tweaks in separate PR, to make things easier to review, i think that would be fine.

booxter:
@jistr will post commits on top without force push to ease reviewers' work. Please don't merge yet.

tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -98,12 +98,14 @@ edpm_bootstrap_command: |
   # https://issues.redhat.com/browse/RHOSZUUL-1517
   dnf -y install crypto-policies
   update-crypto-policies --set FIPS:NO-ENFORCE-EMS
+  {%+ if compute_adoption|bool +%}
   # FIXME: perform dnf upgrade for other packages in EDPM ansible
   # here we only ensuring that decontainerized libvirt can start
   ./venv/bin/repo-setup current-podified -b antelope -d centos9 --stream
```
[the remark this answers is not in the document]
booxter: Indeed, I was too greedy. We should configure the repos; just not execute libvirt hack.

booxter:
@karelyatin @jistr addressed the two remaining comments about repos not installed and ceph-client (but not running the task at all.)

PR #476, opened 2024-05-30
Fix tls connection issues in tlse job

cescgina:
The jobs with tlse keeps failing because of a SSL error on nova_compute.
The fixes for this problem [1][2] suggest that we should add a value for
EDPMServiceType.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/896
[2] https://github.com/openstack-k8s-operators/architecture/pull/247

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -173,6 +173,7 @@ cd -
       - ctlplane
     issuer: osp-rootca-issuer-internal
   caCerts: combined-ca-bundle
+  edpmServiceType: nova
 ----
 +
 [source,yaml]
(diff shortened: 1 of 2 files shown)


lewisdenny (approved):
/lgtm

Here is another example of fixing the SSL error as another example: https://github.com/openstack-k8s-operators/ci-framework/pull/1757

lewisdenny:
@slagle Could you please double check this patch?

fao89 (approved):

fao89:
we should update the docs also: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/83f8cb0727955a3b270483d44b91c0fe9b37709e/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc?plain=1#L162

slagle (commented):
LGTM

jistr (approved):

marios (approved):

PR #478, opened 2024-06-03
Add ceph-hci-pre service when using ceph

cescgina:
When trying to run tempest tests with the standalone jobs, we found that after the edpm deployment, the firewall was blocking access to the ceph monitors, adding this service should configure the firewall to allow the traffic to the mons.

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -355,6 +355,7 @@
         - configure-os
         - run-os
         - install-certs
+        - ceph-hci-pre
         - ceph-client
         - libvirt
         - nova-compute-extraconfig


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -355,6 +355,7 @@
         - configure-os
         - run-os
         - install-certs
+        - ceph-hci-pre
```
fmount, line 358: +1 from me, if Ceph is part of the deployment we need the firewall in place. However I think this might apply to standalone only, because in multinode we're not going to enroll the Controllers as edpm nodes, right? In any case it doesn't hurt, I'm just wondering if we want to put a comment to explain why we need that and which is the use case / scenario we're covering.

frenzyfriday:
Thanks, testing here https://review.rdoproject.org/r/c/testproject/+/53007

frenzyfriday:
recheck

frenzyfriday:
/lgtm

The jobs passed in TP: https://review.rdoproject.org/zuul/build/9501b596ab534ddfbc8df33abecd564f with this patch as depends on

fmount:
@cescgina do you want to rebase and remove the work in progress label?

fmount (approved):
/lgtm

cescgina:
recheck

cescgina:
recheck

jistr (approved):

cescgina:
recheck

cescgina:
recheck

marios (approved):

PR #480, opened 2024-06-04
[WIP] Move heat fields to initial deployment

cescgina:
Recent validation changes fail the heat adoption because the
databaseInstance cannot be patched, so this change moves it to the
initial deployed CR.


bshephar (changes requested):
Failed here again:
```
❯ curl -s https://logserver.rdoproject.org/80/480/6e198c0124dfc0241c463ed624e1ebd5aab56d7e/github-check/data-plane-adoption-osp-17-to-extracted-crc/c4a9d24/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/heats.heat.openstack.org/heat.yaml | yq '.spec | "dbAccount=\(.databaseAccount), DBInstance=\(.databaseInstance)"'
dbAccount=heat, DBInstance=
```

```
❯ curl -s https://logserver.rdoproject.org/80/480/6e198c0124dfc0241c463ed624e1ebd5aab56d7e/github-check/data-plane-adoption-osp-17-to-extracted-crc/c4a9d24/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/heats.heat.openstack.org/heat.yaml | yq '.status.conditions[0]'
lastTransitionTime: "2024-06-04T19:19:21Z"
message: DB create job error occurred MariaDB CR name is not present
reason: Error
severity: Warning
status: "False"
type: Ready
```

Maybe something in the ci-framework is enabling Heat prior to the dataplane-adoption workflow kicking off. As I showed before, when heat is disabled in the control plane, there is no Heat CR in the cluster. So if we consider what the webhook is doing, it is taking the existing CR and compares it to the incoming CR:
https://github.com/openstack-k8s-operators/heat-operator/blob/main/api/v1beta1/heat_webhook.go#L189

So the existence of a Heat CR implies that it's being enabled somewhere prior to the deploy podified heat task in the data-plane-adoption workflow.

I can only assume something in ci-framework is doing it if we're not doing it anywhere in `data-plane-adoption`. But will need to dig into that.

Maybe what makes your [initial patch work](https://github.com/openstack-k8s-operators/data-plane-adoption/commit/bae3d2c4a005be44bef2de4b57800269da422a52) is the fact that you're disabling Heat first and changing the databaseInstance at the same time. Which would work because there is no Heat object, therefore no Heat operator webhook to validate the change. Because just disabling it doesn't change the `databaseInstance`:
```
[m3@localhost ~]$ oc get oscp openstack-galera-network-isolation -o jsonpath='{.spec.heat}' | jq '. | "enabled=\(.enabled), databaseInstance=\(.template.databaseInstance)"'
"enabled=false, databaseInstance=openstack"
```

So updating the `databaseInstance` now while it's disabled would be fine. But once it's enabled, it will be enabled with the existing `databaseInstance` which in this jobs case appears to be `""`. So when we try to patch it, we're changing it from `""` to `openstack` which triggers the webhook validation to fail.

Let's go back to your previous iteration on this PR:
https://github.com/openstack-k8s-operators/data-plane-adoption/commit/bae3d2c4a005be44bef2de4b57800269da422a52

That seems like a sensible path forward here.

fao89:
I notice some services initialize the databaseInstance at the openstack-operator:
https://github.com/search?q=repo%3Aopenstack-k8s-operators%2Fopenstack-operator+DatabaseInstance+language%3AGo&type=code&l=Go
I've proposed this:
https://github.com/openstack-k8s-operators/openstack-operator/pull/826

cescgina:
recheck

fao89:
https://github.com/openstack-k8s-operators/openstack-operator/pull/822 will probably fix this issue

cescgina:
> [openstack-k8s-operators/openstack-operator#822](https://github.com/openstack-k8s-operators/openstack-operator/pull/822) will probably fix this issue

ack, thanks @fao89 let's wait for the CI results there, at least we know that this PR works, so tomorrow we can resolve the issue

bshephar (approved):

bshephar:
> We should make the change also in docs but since this is a CIX fix we can merge this as-is first.

I think the documented procedure is fine. If you're a customer doing this, the Heat will presumably be disabled which is what it says here:
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/proc_adopting-the-orchestration-service.adoc?plain=1#L6

The problem we're having with this job is that _something_ is enabling Heat prior to the `data-plane-adoption` `heat_adoption` workflow.
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/480#pullrequestreview-2097617056

I think we would need to figure out what is enabling Heat in the first place to better understand why we end up in this situation.

I could probably document the validation bypass annotation if we suspect users will end up in this situation during adoption:
https://github.com/openstack-k8s-operators/heat-operator/blob/main/api/v1beta1/heat_webhook.go#L183-L185

son-vyas:
recheck

cescgina:
the job is passing in https://github.com/openstack-k8s-operators/openstack-operator/pull/822, so we don't need this, closing

PR #490, opened 2024-06-14
use the datasources field for the nova dataplane service

fao89:


docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -249,13 +250,15 @@ metadata:
...
-    - nova-migration-ssh-key
+  dataSources:
+    - secretRef:
+        name: nova-cell1-compute-config
+    - secretRef:
+        name: nova-migration-ssh-key
+    - configMapRef:
+        name: nova-compute-extraconfig
+    - configMapRef:
+        name: ceph-nova
```
fao89, line 261: should we add it as optional in the default nova service?

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -195,12 +195,13 @@ metadata:
   name: nova-compute-extraconfig
   namespace: openstack
 spec:
-  label: nova.compute.extraconfig
-  configMaps:
-    - nova-compute-extraconfig
-  secrets:
-    - nova-cell1-compute-config
-    - nova-migration-ssh-key
+  dataSources:
```
fao89, line 198: I kept nova-compute-extraconfig service, to keep the simmetry with the ceph test

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
SeanMooney: no infact i would like to do the opiciete and adjust all our test code to use nova-extra-config

that is the name that we want to standardise on in our docs and test code.

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
SeanMooney: well i would prefer to chnage the ceph test to use nova-extra-config insetad

SeanMooney (changes requested):

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -242,22 +221,6 @@ data:
     images_rbd_glance_copy_timeout=600
     rbd_user=openstack
     rbd_secret_uuid=$CEPH_FSID
----
```
SeanMooney, line 245: ya so this is exactly what i would have done too.

we should not need to use customer service defieitons of nova or nova with cpeh anymore

we just need to use a single 

nova-extra-config configmap which we update as required in the procedure.

the only expction to that is if you need multiple ndoes sets with different nova config per nodeset in that case we will need a separate nova service per nodeset but +1 on this apporoch in general.

fao89:
recheck

jistr (approved):

bogdando:
This change has left the documentation irrelevant. It sill mentions creation of a service, while there is no more service created, but a serviceOverride to a "stock osdpdservice" nova instead. Furthermore, it mentions adding ``tlsCert`` to the *removed* services, which cannot be done as documented, and serviceOverride doesn't help with that?

PR #494, opened 2024-06-18
Docs: fix RHEL versions in downstream rendering

jistr:


docs_user/rhoso_attributes.adoc
```diff
@@ -5,11 +5,11 @@
 // Example: Use {osp_long} to display "OpenStack Platform".
 
 // RHEL version attributes
```
jistr, line 7: @klgill I'm curious if we need this file at all? We have the `rhel_*_ver` variables in adoption-attributes.adoc too, and they were correct there but the settings got overriden by the incorrect values in this file. We should have the variables in one place only for each variant (upstream vs. downstream).

docs_user/adoption-attributes.adoc
```diff
@@ -2,7 +2,7 @@
 :context: assembly
 :build: downstream
 
-ifeval::["{build}" == "upstream"]
+ifeval::["{build}" != "downstream"]
```
klgill, line 5: Can you help me understand why you switched this ifeval to downstream? The names listed in the variables below it are upstream-specific, e.g., we don't use "Wallaby" downstream.

klgill (commented):
@jistr This is correct to update the "rhel_prev_ver", "rhel_curr_ver", and "rhelvernum" variables upstream. We should merge that change. 

The problem actually originated downstream. I had duplicate files of the rhoso-attributes.adoc and adoption-attributes.adoc in the two different folders. The rhoso-attributes file in the "common" folder had the correct values, but the rhoso-attributes file in the adopting-data-plane folder was out of date.

I deleted the files in this MR downstream (not merge yet): https://gitlab.cee.redhat.com/rhci-documentation/docs-Red_Hat_Enterprise_Linux_OpenStack_Platform/-/merge_requests/12768/diffs#note_11252720

I'm still trying to get clarity on whether I need both attributes files. But even if we keep both, I just need to make sure that the rhoso-attributes files in upstream and downstream is in sync.

jistr:
@klgill Ok thanks i'll merge this then, to fix the upstream preview of downstream docs at least.

If we find we do need the rhoso-attributes file, we could move all the downstream variant variables there, to remove duplication between adoption-attributes and rhoso-attributes, so the adoption-attributes file could contain only upstream values perhaps. I don't have a strong opinion on how this should be solved going forward, just that having the same variables for downstream variant in two different places, and one of the places being ignored, is not ideal :).

PR #501, opened 2024-06-24
Issue a token before adoption and testing it after

afaranha:


jistr (approved):

tests/roles/keystone_adoption/tasks/main.yaml
```diff
@@ -87,3 +87,15 @@
     for service in aodh heat heat-cfn barbican cinderv3 glance manila manilav2 neutron nova placement swift ironic-inspector ironic; do
       ${BASH_ALIASES[openstack]} service list | awk "/ $service /{ print \$2; }" | xargs ${BASH_ALIASES[openstack]} service delete || true
     done
+
+- name: debug token
+  ansible.builtin.debug:
+    var: before_adoption_token
```
odyssey4me, line 93: This assumes that this adoption process is running on a dev environment, which is true for CI but not for all adoptions. This set of tasks should be changed to either not make that assumption or skip if before_adoption_token is not set.

PR #505, opened 2024-06-25
Remove download-cache-openstack service

archana203:
During Dataplane adoption, download-cache service tries to update libvirt related packages which is not needed.

Resolves: [OSPRH-7874](https://issues.redhat.com/browse/OSPRH-7874)

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -248,7 +248,6 @@ spec:
   preProvisioned: true
   services:
     - bootstrap
-    - download-cache
     - configure-network
     - validate-network
     - install-os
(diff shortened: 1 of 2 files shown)


bogdando:
Does this fix a regression? this used to work w/o issues so far?

jistr (dismissed):
LGTM pending a CI pass. For reviewers' context: this is aiming to fix this bug https://issues.redhat.com/browse/OSPRH-7874

bogdando:
I think we need the latest openstack-selinux and latest libvirt during adoption?

@SeanMooney

SeanMooney:
so libvirt is not installed on the host pre-adoption and it is installed on the host during adoption as part of the greenfield deployment

the download-cache just pre-downloads the package that will be installed by edpm_libvirt during adoption so I don't think this will help.
i would expect this to just delay the package install slightly.

repo config is meant to be done via bootstrap so by the time download-cache or edpm_libvirt run the repo/subscription management is already done so neither role interact with subscription manager.

bogdando (changes requested):
please see above

archana203:
Based on the discussion, this PR is not valid for the issue. Closing it.

PR #508, opened 2024-06-26
Add pre-adoption optional reboot check

ciecierski:
Add OpenStackDataPlaneDeployment with reboot service to check if there is any pre-adoption 
reboot to be performed. It will not start reboot and only display information in ansibleee logs.

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -544,6 +544,26 @@ oc logs -l app=openstackansibleee -f --max-log-requests 20
 oc wait --for condition=Ready openstackdataplanedeployment/openstack-pre-adoption --timeout=10m
 ----
 
+.. Optional: Create a OpenStackDataPlaneDeployment CR that runs the reboot service in check-mode only.
+Reboot service is going to check if there is any reboot to be performed.:
++
+[source,yaml]
+----
+oc apply -f - <<EOF
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneDeployment
+metadata:
+  name: openstack-pre-adoption-reboot-check
+spec:
+  nodeSets:
+  - openstack
+  servicesOverride:
+  - reboot-os
+  ansibleExtraVars:
+  - edpm_reboot_strategy: never
+EOF
+----
+
 . If any openstack-pre-adoption validations fail, you must first determine
 which ones were unsuccessful based on the ansible logs and then follow the
 instructions below for each case:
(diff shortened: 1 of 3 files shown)


docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -560,6 +580,9 @@ the 17 node.
 `edpm_tuned_profile` variable in the `OpenStackDataPlaneNodeSet` is configured
 to use the same profile as set on the (source) OSP 17 node.
 
+* if reboot check returned that reboot is required then make sure that the
```
gibizer, line 583: probably worth describing how to get the returned information from the ansibleee pod logs.

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -385,6 +385,57 @@
...
+        kind: OpenStackDataPlaneDeployment
+        metadata:
+          name: openstack-pre-adoption-reboot-check
+        spec:
+          nodeSets:
+          - openstack
+          servicesOverride:
+          - reboot-os
+          ansibleExtraVars:
+          - edpm_reboot_strategy: never
```
gibizer, line 405: does this cause the ansible code to fail if reboot is required?  If not then we need some additional logic below that stops the automated process if reboot is required.

jistr:
Is this something that's needed? The os-reboot role now has a check that should always make it skip reboots on Adoption, even if `edpm_reboot_strategy` is left at default i think? And we already have a validation Deployment that aims to verify that the reboot-worthy parameters are not changing. So this would add another step into the procedure but i'm not sure if it adds additional validation beyond what we already have?

ciecierski:
> Is this something that's needed? The os-reboot role now has a check that should always make it skip reboots on Adoption, even if `edpm_reboot_strategy` is left at default i think? And we already have a validation Deployment that aims to verify that the reboot-worthy parameters are not changing. So this would add another step into the procedure but i'm not sure if it adds additional validation beyond what we already have?

It was raised by @archana203 , that we are not testing 'never' strategy and this is the only place where it applies. I'm not fully sure if this is needed, that's why I made it optional. Yes, reboot skips on Adoption.  The only  value  it adds is about checking reboot that was deferred and never run( as an example after minor update)

jistr:
If we add an optional test like this, nothing is going to run it unless we add a new job for it (which we IMO shouldn't given how long these jobs take). I think the test for the `never` strategy should probably be some smaller (molecule?) test if possible. In the adoption test suite which is used for end-to-end testing, i think we should test the "default path" which is relying on the `auto` strategy behavior.

PR #515, opened 2024-07-01
refining deploying backend services Pt1

klgill:
https://issues.redhat.com/browse/RHOSPDOC-1891

### DIFF
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -121,7 +121,7 @@ $ oc set data secret/osp-secret "PlacementPassword=$PLACEMENT_PASSWORD"
 $ oc set data secret/osp-secret "SwiftPassword=$SWIFT_PASSWORD"
 ----
 
-. If you enabled TLS-e in your {OpenStackShort} environment, in the `spec:tls` section, set the `enabled` parameter to `true`: 
+. If you enabled TLS-e in your {OpenStackShort} environment, in the `spec:tls` section, set the `enabled` parameter to `true`:
 +
 [source,yaml]
 ----
@@ -164,7 +164,7 @@ spec:
 ----
 
 . Deploy the `OpenStackControlPlane` CR. Ensure that you only enable the DNS, MariaDB, Memcached, and RabbitMQ services. All other services must
-be disabled: 
+be disabled:
 +
 [source,yaml]
 ----
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -46,7 +46,7 @@ ADMIN_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AdminPassword:
 ----
 * Set the service password variables to match the original deployment.
 Database passwords can differ in the control plane environment, but
-synchronizing the service account passwords is required.
+you must synchronize the service account passwords.
 +
 For example, in developer environments with {OpenStackPreviousInstaller} Standalone, the passwords can be extracted:
 +
@@ -148,8 +148,7 @@ spec:
       enabled: true
 ----
 
-. If your {OpenStackShort} environment does not enable TLS-e, in the
-`spec:tls`` section, set the `enabled` parameter to `false`:
+. If you did not enable TLS-e, in the `spec:tls`` section, set the `enabled` parameter to `false`:
 +
 [source,yaml]
 ----
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -1,18 +1,11 @@
 [id="deploying-backend-services_{context}"]
 
-= Deploying backend services
+= Deploying back-end services
 
-Create the `OpenStackControlPlane` custom resource (CR) with basic backend services deployed, and all the {rhos_prev_long} ({OpenStackShort}) services disabled. This will be the foundation of the control plane.
-
-In subsequent steps, you import the original databases and then add
-{OpenStackShort} control plane services.
+Create the `OpenStackControlPlane` custom resource (CR) with the basic back-end services deployed, and disable all the {rhos_prev_long} ({OpenStackShort}) services. This CR is the foundation of the control plane.
 
 .Prerequisites
 
-* The cloud that you want to adopt is up and running, and it is on the
-{OpenStackShort} {rhos_prev_ver} release.
-* All control plane and data plane hosts of the source cloud are up
-and running, and continue to run throughout the adoption procedure.
 * The `openstack-operator` is deployed, but `OpenStackControlPlane` is
 not deployed.
 ifeval::["{build}" != "downstream"]
@@ -26,7 +19,7 @@ endif::[]
 For production environments, the deployment method will likely be
 different.
 
-* If TLS Everywhere is enabled on the source environment, the `tls` root CA from the source environment must be copied over to the rootca-internal issuer.
+* If you enabled TLS Everywhere on the {OpenStackShort} environment, you must copy the `tls` root CA from the {OpenStackShort} environment to the `rootca-internal` issuer.
 
 * There are free PVs available to be claimed (for MariaDB and RabbitMQ).
 ifeval::["{build}" != "downstream"]
@@ -35,7 +28,7 @@ For developer/CI environments driven by install_yamls, make sure
 you've run `make crc_storage`.
 endif::[]
 * Set the desired admin password for the control plane deployment. This can
-be the original deployment's admin password or something else.
+be the admin password from your original deployment or a different password:
 +
 ----
 ADMIN_PASSWORD=SomePassword
@@ -46,12 +39,11 @@ To use the existing {OpenStackShort} deployment password:
 ----
 ADMIN_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AdminPassword:' | awk -F ': ' '{ print $2; }')
 ----
-* Set service password variables to match the original deployment.
+* Set the service password variables to match the original deployment.
 Database passwords can differ in the control plane environment, but
-synchronizing the service account passwords is a required step.
+synchronizing the service account passwords is required.
 +
-For example, in developer environments with {OpenStackPreviousInstaller} Standalone, the
-passwords can be extracted like this:
+For example, in developer environments with {OpenStackPreviousInstaller} Standalone, the passwords can be extracted:
 +
 ----
 AODH_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AodhPassword:' | awk -F ': ' '{ print $2; }')
@@ -73,14 +65,14 @@ SWIFT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' SwiftPassword:
 
 .Procedure
 
-. Make sure you are using the {OpenShift} namespace where you want the
-control plane deployed:
+. Ensure that you are using the {OpenShift} namespace where you want the
+control plane to be deployed:
 +
 ----
-oc project openstack
+$ oc project openstack
 ----
 
-. Create OSP secret.
+. Create the OSP secret.
 ifeval::["{build}" != "downstream"]
 +
 The procedure for this will vary, but in developer/CI environments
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -1,18 +1,11 @@
...
-{OpenStackShort} control plane services.
+Create the `OpenStackControlPlane` custom resource (CR) with the basic back-end services deployed, and disable all the {rhos_prev_long} ({OpenStackShort}) services. This CR is the foundation of the control plane.
 
 .Prerequisites
 
-* The cloud that you want to adopt is up and running, and it is on the
-{OpenStackShort} {rhos_prev_ver} release.
-* All control plane and data plane hosts of the source cloud are up
-and running, and continue to run throughout the adoption procedure.
 * The `openstack-operator` is deployed, but `OpenStackControlPlane` is
```
klgill, line 9: I removed the first two prerequisites because they seemed more general to adoption rather than specific to this procedure. I will likely move those general prereqs to an “adoption preparation checklist”, which I’m developing.

Is the prereq, “The `openstack-operator` is deployed, but `OpenStackControlPlane` is
not deployed," more generic or specific to deploying back-end services?

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -26,16 +19,16 @@ endif::[]
 For production environments, the deployment method will likely be
```
klgill, line 19: Can we elaborate on how the deployment method is different?

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -73,14 +65,14 @@ SWIFT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' SwiftPassword:
...
+. Ensure that you are using the {OpenShift} namespace where you want the
+control plane to be deployed:
 +
 ----
-oc project openstack
+$ oc project openstack
 ----
 
-. Create OSP secret.
+. Create the {OpenStackShort} secret.
```
klgill, line 75: Can we give an example of one common way to create the secret downstream?

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -92,43 +84,40 @@ make input
...
+$ oc set data secret/osp-secret "ManilaPassword=$MANILA_PASSWORD"
+$ oc set data secret/osp-secret "NeutronPassword=$NEUTRON_PASSWORD"
+$ oc set data secret/osp-secret "NovaPassword=$NOVA_PASSWORD"
+$ oc set data secret/osp-secret "OctaviaPassword=$OCTAVIA_PASSWORD"
+$ oc set data secret/osp-secret "PlacementPassword=$PLACEMENT_PASSWORD"
+$ oc set data secret/osp-secret "SwiftPassword=$SWIFT_PASSWORD"
 ----
 
-. Deploy `OpenStackControlPlane`. Make sure to only enable DNS,
+. Deploy the `OpenStackControlPlane` CR. Ensure that you only enable the DNS,
```
klgill, line 116: Can we provide the command for this? I'm not sure if something like the following is correct:
$ oc patch openstackcontrolplane openstack --type=merge --patch '
spec:
  dns:
      enabled: true
      apiOverride:
          route: {}
…

$ oc patch openstackcontrolplane openstack --type=merge --patch '
spec:
 MariaDB:
      enabled: true
      apiOverride:
          route: {}
…
and so on...

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -253,18 +241,18 @@ spec:
     templates:
       openstack:
         secret: osp-secret
-        replicas: 1
```
klgill, line 256: Just FYI: I updated the replicas as described in https://issues.redhat.com/browse/OSPRH-5645?focusedId=25012945&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25012945

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -155,9 +144,8 @@ spec:
       enabled: true
 ----
 
-. If the source environment does not enable TLS Everywhere, modify
-spec:tls section with the following override before applying it:
-
+. If your {OpenStackShort} environment does not enable TLS-e, in the
```
klgill, line 147: This step says that you only have to set the enabled parameter to "false" in the spec:tls section, but there's also a spec for the osp-secret that has all the services' enabled parameters set to false. 
Should we have another step below this one that says, "If your RHOSP environment does not enable TLS-e, in the osp-secret spec, set the enabled parameter in each service to "false"?

klgill:
@jistr I'm not sure if you are the right person to review this procedure. I think a lot of engineers contributed. Let me know if I should tag anyone else.

docs_user/modules/proc_deploying-backend-services.adoc
klgill: Also, are you modifying the spec:tls section in the openstack_control_plane.yaml file in steps 6 and 7?

docs_user/modules/proc_deploying-backend-services.adoc
jistr: I think we should actually keep these, they were an explicit request based on past experience. Folks were sometimes starting the upgrade procedure when some nodes were broken/offline but not properly removed from the OpenStack deployment, and it caused a failure and escalation.

docs_user/modules/proc_deploying-backend-services.adoc
jistr: I think we should have an `ifeval` here for downstream content, and instead of this sentence we should have "Install the OpenStack Operators according to section 1 of the deployment guide." and link to https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0-beta/html-single/deploying_red_hat_openstack_services_on_openshift/index#assembly_installing-and-preparing-the-Operators

docs_user/modules/proc_deploying-backend-services.adoc
jistr: We should either link to section 2.2 "Providing secure access to the Red Hat OpenStack Services on OpenShift services" of the deployment guide or pull the same content in here: https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0-beta/html-single/deploying_red_hat_openstack_services_on_openshift/index#proc_providing-secure-access-to-the-RHOSO-services_preparing

docs_user/modules/proc_deploying-backend-services.adoc
jistr: The command for the creation of the OpenStackControlPlane CR is below, it's the one that starts:

```
oc apply -f - <<EOF
apiVersion: core.openstack.org/v1beta1
kind: OpenStackControlPlane
...
```

Over time these TLS-specific configuration amendments were added prior to it, so this sentence got shifted out of place. It should immediately precede the `oc apply` command.

docs_user/modules/proc_deploying-backend-services.adoc
jistr: >  there's also a spec for the osp-secret that has all the services' enabled parameters set to false

I'm not aware of this. Can you link what you're referring to?

> are you modifying the spec:tls section in the openstack_control_plane.yaml file in steps 6 and 7?

This is about modifying the content of the OpenStackControlPlane CR creation snippet below, the one starting `oc apply`. We don't have the openstack_control_plane.yaml file in the Adoption guide presently but the `oc apply` snippet has the same purpose.

docs_user/modules/proc_deploying-backend-services.adoc
jistr: +1 that's a good catch, we could perhaps delete the replicas key instead though and rely on the default which should be 3. The `replicas: 1` here came from us deploying slimmed-down environments for development.

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -326,17 +314,17 @@ spec:
       swiftRing:
         ringReplicas: 1
       swiftStorage:
-        replicas: 0
```
jistr, line 329: We need to keep zero replicas here. Swift storage adoption has its own procedure separate from the main OpenStack adoption.

jistr (commented):
Thanks, some comments inline and otherwise LGTM.

docs_user/modules/proc_deploying-backend-services.adoc
klgill: > > there's also a spec for the osp-secret that has all the services' enabled parameters set to false
> 
> I'm not aware of this. Can you link what you're referring to?

In step 7, there are 2 code blocks one after another.
https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0-beta/html/adopting_a_red_hat_openstack_platform_17.1_deployment/migrating-databases-to-the-control-plane_storage-requirements#deploying-backend-services_migrating-databases

klgill: > > there's also a spec for the osp-secret that has all the services' enabled parameters set to false
> 
> I'm not aware of this. Can you link what you're referring to?
> 
I realize now that the "oc apply -f - <<EOF" command does not belong in step 7. It needs to be shifted up to step 5.

jistr: @klgill Step 6 should come before the `oc apply` though ("modify tls:spec section *before applying it*"). So rather than moving the `oc apply` to step 5, let's get rid of step 5 by moving that description to step 7. (Then step 6 becomes step 5 and step 7 becomes step 6 :).)

jistr: Also Step 7 now combines the override for nonTLS with the actual `oc apply`. I think it would make more sense for what is now Step 6 to be the "tls:spec" section overrides, both for TLS and non-TLS variants, and then after that comes the deployment step with `oc apply`.

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -46,12 +44,11 @@ To use the existing {OpenStackShort} deployment password:
...
 Database passwords can differ in the control plane environment, but
-synchronizing the service account passwords is a required step.
+synchronizing the service account passwords is required.
 +
-For example, in developer environments with {OpenStackPreviousInstaller} Standalone, the
-passwords can be extracted like this:
+For example, in developer environments with {OpenStackPreviousInstaller} Standalone, the passwords can be extracted:
 +
 ----
 AODH_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AodhPassword:' | awk -F ': ' '{ print $2; }')
```
jistr, line 54: I may send another PR to amend this example for better downstream suitability and remove the replica numbers below but i think this PR is good to merge.

jistr (approved):

docs_user/modules/proc_deploying-backend-services.adoc
jistr: I created a bug so that i don't forget about it https://issues.redhat.com/browse/OSPRH-8880

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -46,12 +44,11 @@ To use the existing {OpenStackShort} deployment password:
 ----
 ADMIN_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' AdminPassword:' | awk -F ': ' '{ print $2; }')
 ----
-* Set service password variables to match the original deployment.
+* Set the service password variables to match the original deployment.
 Database passwords can differ in the control plane environment, but
-synchronizing the service account passwords is a required step.
+synchronizing the service account passwords is required.
```
rheslop, line 49:
```suggestion
Synchronize the service account passwords.
```

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -73,15 +70,15 @@ SWIFT_PASSWORD=$(cat ~/tripleo-standalone-passwords.yaml | grep ' SwiftPassword:
 
 .Procedure
 
-. Make sure you are using the {OpenShift} namespace where you want the
-control plane deployed:
+. Ensure that you are using the {OpenShift} namespace where you want the
```
rheslop, line 73: This looks like it will expand to "Ensure that you are using the Red Hat OpenShift Container Platform namespace where..."
This isn't incorrect, but if there's a chance the customer won't be confused at this point to simply say "namespace", then it might be easier to read through.

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -91,44 +88,40 @@ you use `install_yamls`:
...
 ----
 endif::[]
+ifeval::["{build}" == "downstream"]
+. Create the {OpenStackShort} secret. For more information, see link:https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/{rhos_curr_ver}/html-single/deploying_red_hat_openstack_services_on_openshift/index#proc_providing-secure-access-to-the-RHOSO-services_preparing[Providing secure access to the Red Hat OpenStack Services on OpenShift services] in _Deploying Red Hat OpenStack Services on OpenShift_.
+endif::[]
 
-. If the `$ADMIN_PASSWORD` is different than the already set password
-in `osp-secret`, amend the `AdminPassword` key in the `osp-secret`
-correspondingly:
+. If the `$ADMIN_PASSWORD` is different than the password you set
```
rheslop, line 95: Do steps like this typically begin with "Optional:" ?

klgill: I don't think it would be labeled "optional" because the step is required for those who have a different password from the osp-secret. It's a conditional step.

PR #521, opened 2024-07-09
Fail if mons in quorum are less than expected

fmount:
This patch refines the Ceph migration flow to take into account situations relevant for CI integration.
1. Improved modularity: `backup.yaml` and `firewall.yaml` have been added to collect and reuse the same tasks in different parts of the playbook
2. Fixed the check related to `mon_quorum`, and wait for `mons` to appear in `Ceph` (it might require more time depending on the environment) 
3. Deduplicate code and turn off some debugging bool enabled by default
4. optionally skip the IP migration in case we do not require to keep mon IP addresses (false by default)

Jira: https://issues.redhat.com/browse/OSPRH-2540

### DIFF
--- a/tests/roles/ceph_migrate/tasks/mon.yaml
+++ b/tests/roles/ceph_migrate/tasks/mon.yaml
@@ -11,13 +11,13 @@
   ansible.builtin.debug:
     msg: "Migrate mon: {{ cur_mon }} to node: {{ target_node }}"
 
-- name: MON - Get current mon IP address from mon_map override
-  when: mon_map is defined
+- name: MON - Get current mon IP address from node_map override
+  when: node_map is defined
   ansible.builtin.set_fact:
     mon_ipaddr: "{{ mon_ip | trim | ansible.utils.ipaddr }}"
   vars:
     mon_ip: |-
-      {% for mon in mon_map %}
+      {% for mon in node_map %}
       {%   if mon.hostname == cur_mon %}
       {{ mon.ip }}
       {%   endif %}
@@ -27,13 +27,13 @@
   become: true
   ansible.builtin.command: "{{ ceph_cli }} -s -f json"
   register: monmap
-  retries: 20
-  delay: 2
-  # >= is related to the fact that the current monmap might still have
-  # references to the old mon daemon coming from the drained node
-  until: (monmap.stdout | from_json | community.general.json_query('monmap.num_mons') | int) >= ((decomm_nodes |default([]) | length | int) | default(3))
-  loop_control:
-    label: "MON - check mons quorum"
+  tags:
+    - ceph_mon_quorum
+
+- name: ansible.builtin.fail if there's no quorum
+  ansible.builtin.fail:
+    msg: "Mons are not in quorum"
+  when: (monmap.stdout | from_json | community.general.json_query('monmap.num_mons') | int) < ((decomm_nodes |default([]) | length | int) | default(3))
   tags:
     - ceph_mon_quorum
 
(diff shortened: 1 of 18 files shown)


jistr (approved):

jistr:
(Still a draft? If so we should probably convert it to a draft PR.)

fmount:
> (Still a draft? If so we should probably convert it to a draft PR.)

Yes, sorry, I think it deserves a few more CI runs. I'll resume this one after pto

fmount:
I'll turn to draft and unblock when I have a full green run in CI via teatproject

danpawlik:
recheck

danpawlik:
recheck

tests/roles/ceph_migrate/tasks/rbd.yaml
```diff
@@ -85,7 +85,6 @@
         target_node: "{{ node.1 }}"
       # This condition might be a different one
       loop: "{{ decomm_nodes | zip(target_nodes) }}"
-      # loop: "{{ decomm_nodes|zip(hostmap.keys() | difference(decomm_nodes) | sort) | list }}"
```
fmount, line 88: removed an old comment

tests/roles/ceph_migrate/tasks/mon.yaml
```diff
@@ -263,8 +235,8 @@
   become: true
   ansible.builtin.command: "{{ ceph_cli }} -s -f json"
   register: monmap
-  retries: 20
-  delay: 2
+  retries: 30
```
fmount, line 238: increasing a bit these values: I'll turn them into defaults in a follow up.

tests/roles/ceph_migrate/tasks/mon.yaml
```diff
@@ -169,53 +178,23 @@
...
-    # takes initiative to redeploy a mon, on an arbitrary ip address (if many)
-    # between the `ceph orch rm ...` command and the `ceph orch daemon add ...`
-    # one. For this reason, the deamon redeploy follows this flow:
-    # 1. update and unmanage the spec
-    # 2. redeploy on the right network as per the official doc:
-    #    https://docs.ceph.com/en/quincy/cephadm/services/mon/
-    # 3. manage the spec
-    # Note: a pause between the steps is required to allow the orchestrator to
-    #       process the spec and update the daemons ref
-    - name: Unmanage mons
```
fmount, line 197: Moving this task at the beginning to have a more clean flow (and easy to read / understand)

tests/roles/ceph_migrate/tasks/mon.yaml
```diff
@@ -103,6 +102,9 @@
     - ceph_drain
 
 - name: MON - Get current mon IP address
+  when:
+    - mon_ipaddr is not defined
+    - discover_mons | bool | default(true)
```
fmount, line 107: discover_mons is experimental, right now the rdo job defines a static `mon_map` override to make sure we have predictable results.

tests/roles/ceph_migrate/tasks/mon.yaml
```diff
@@ -41,30 +37,13 @@
...
 
+# Backup data
 - name: Backup data for client purposes
-  delegate_to: "{{ cur_mon }}"
+  ansible.builtin.include_tasks: backup.yaml
+  vars:
+    host: "{{ cur_mon }}"
   tags:
     - ceph_backup
-  block:
```
fmount, line 48: some code cleanup: the same, duplicated block, has been moved to `backup.yaml`

fmount:
This patch is getting bigger, and I'm worried to lose the scope the more I work with the CI integration.
Before adding an extra patch on top related to network and IP migration (which seems still problematic in CI), I'd like to merge the changes proposed here because I had a green execution from a testproject.

fmount:
> This patch is getting bigger, and I'm worried to lose the scope the more I work with the CI integration. Before adding an extra patch on top related to network and IP migration (which seems still problematic in CI), I'd like to merge the changes proposed here because I had a green execution from a testproject.

@cescgina @marios I'm wondering if we can land this patch so I don't risk to diverge from the topic of this PR. Still the job is not executed, but I'm refining the procedure more to deal with upstream multinode CI other than IR (that won't be used for this purpose).

cescgina (approved):
Since this is not running ans has no risk to CI, I'm fine merging and iterating if it makes dev work easier

fmount:
@cescgina thank you!

PR #525, opened 2024-07-12
Creating an empty template does not allow telemetry to set up properl…

jlarriba:
…y its defaults


stuggi:
lgtm @jistr we'd need this to unblock an openstack-operator bump. In general I think we could remove all the `template: {}`, but need to validate. therefore should do it in a follow up.

jistr (approved):

PR #526, opened 2024-07-12
Introduce Glance NFS adoption in the test suite

fmount:
`Glance` previously had a `local` backend use case. It can be easily turned into the `NFS` backend through `extraMounts`.
This patch is supposed to delete that (unsupported) use case and introduce the `GlanceNFS` backend one.

Jira: https://issues.redhat.com/browse/OSPRH-8594

### DIFF
--- a/tests/roles/glance_adoption/tasks/main.yaml
+++ b/tests/roles/glance_adoption/tasks/main.yaml
@@ -1,9 +1,23 @@
 - name: deploy podified Glance with local backend
-  ansible.builtin.shell: |
-    {{ shell_header }}
-    {{ oc_header }}
-    oc patch openstackcontrolplane openstack --type=merge --patch-file={{ role_path }}/files/glance_local.yaml
-  when: glance_backend == 'local'
+  when: glance_backend == 'nfs'
+  block:
+    - name: Fail if no nfs_server_addr/nfs_server_path are definied
+      when:
+        - nfs_server_addr is not defined
+        - nfs_server_path is not defined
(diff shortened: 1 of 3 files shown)


tests/roles/glance_adoption/tasks/main.yaml
```diff
@@ -1,9 +1,23 @@
 - name: deploy podified Glance with local backend
-  ansible.builtin.shell: |
-    {{ shell_header }}
-    {{ oc_header }}
-    oc patch openstackcontrolplane openstack --type=merge --patch-file={{ role_path }}/files/glance_local.yaml
-  when: glance_backend == 'local'
+  when: glance_backend == 'nfs'
```
konan-abhi, line 2: `glance_backend` where this variable is defined?

Because glance doesn't have any backend named nfs.

fmount: This is defined in the job and passed to the playbook to identify the use case we're going to cover (it's not something that goes to enabled_backends, where we have File)

konan-abhi (approved):

tests/roles/glance_adoption/tasks/main.yaml
```diff
@@ -1,9 +1,23 @@
 - name: deploy podified Glance with local backend
```
jistr, line 1: Nit: this should now say "with NFS backend"? But we can land the patch and tweak in another one, since the CI is green.

jistr (approved):

PR #528, opened 2024-07-12
refined placement procedure

klgill:
https://issues.redhat.com/browse/RHOSPDOC-1915

### DIFF
--- a/docs_user/modules/proc_adopting-the-placement-service.adoc
+++ b/docs_user/modules/proc_adopting-the-placement-service.adoc
@@ -7,8 +7,7 @@ To adopt the Placement service, you patch an existing `OpenStackControlPlane` cu
 .Prerequisites
 
 * You import your databases to MariaDB instances on the control plane. For more information, see xref:migrating-databases-to-mariadb-instances_migrating-databases[Migrating databases to MariaDB instances].
-* You adopt the {identity_service_first_ref} to the {rhos_long} the xref:adopting-the-identity-service_adopt-control-plane[Adopting the Identity service].
-* You must deploy the Memcached operator. For more information, see xref:deploying-backend-services_migrating-databases[Deploying back-end services].
+* You adopt the {identity_service_first_ref}. For more information, see xref:adopting-the-identity-service_adopt-control-plane[Adopting the Identity service].
 
 .Procedure
 
--- a/docs_user/modules/proc_adopting-the-placement-service.adoc
+++ b/docs_user/modules/proc_adopting-the-placement-service.adoc
@@ -2,21 +2,20 @@
 
 = Adopting the Placement service
 
+To adopt the Placement service, you patch an existing `OpenStackControlPlane` custom resource (CR) that has the Placement service disabled. The patch starts the service with the configuration parameters that are provided by the {rhos_prev_long} ({OpenStackShort}) environment.
+
 .Prerequisites
 
-* Previous Adoption steps completed. Notably,
- ** the xref:migrating-databases-to-mariadb-instances_migrating-databases[Migrating databases to MariaDB instances]
-must already be imported into the control plane MariaDB.
- ** the xref:adopting-the-identity-service_adopt-control-plane[Adopting the Identity service] needs to be imported.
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_adopting-the-placement-service.adoc
```diff
@@ -2,21 +2,20 @@
...
 .Prerequisites
 
-* Previous Adoption steps completed. Notably,
- ** the xref:migrating-databases-to-mariadb-instances_migrating-databases[Migrating databases to MariaDB instances]
-must already be imported into the control plane MariaDB.
- ** the xref:adopting-the-identity-service_adopt-control-plane[Adopting the Identity service] needs to be imported.
- ** the Memcached operator needs to be deployed (nothing to import for it from
-the source environment).
+* You import your databases to MariaDB instances on the control plane. For more information, see xref:migrating-databases-to-mariadb-instances_migrating-databases[Migrating databases to MariaDB instances].
+* You adopt the {identity_service_first_ref} to the {rhos_long} the xref:adopting-the-identity-service_adopt-control-plane[Adopting the Identity service].
```
klgill, line 10: Can these prerequisites be removed? Customers should be following the guide in sequence, so they would have done these steps already. But if there’s a prerequisite that you think we should emphasize, please let me know.

klgill:
Hi @SeanMooney , can you please review my edits to the Placement service adoption procedure?

docs_user/modules/proc_adopting-the-placement-service.adoc
```diff
@@ -41,13 +40,13 @@ spec:
...
 .Verification
 
-* Check that Placement endpoints are defined and pointing to the
-control plane FQDNs and that Placement API responds:
+* Check that the Placement service endpoints are defined and pointing to the
+control plane FQDNs, and that the Placement API responds:
 +
 ----
-alias openstack="oc exec -t openstackclient -- openstack"
+$ alias openstack="oc exec -t openstackclient -- openstack"
```
SeanMooney, line 47: as a genera comment adding `$` to the start of a line to indifcat a prompt signifcally reduced the ux of using the docs as you cannot simply copy paste the lines simply.

the rendering hack that tries to remove those automatically in some cases does not help with this. escipally not in the pdf version fo the docs or when they are sued directly form github.

so i would reconsider making this change and reevaluate the style rule that is driving it.

docs_user/modules/proc_adopting-the-placement-service.adoc
SeanMooney: i think they can be removed however the guide was not originally intended to have a strictly defined order 

note that some service have depencies one other and each service section was originally intended to be completed independently.


if you want the doc to be followed in order then you need to do the adtoping in depency order not simply alpahpetically for example

placement depends on mariadb, and keystone

so keystone must be adopted before placement. nova depend on placement so placement should be adopted before nova is.

placement does not depend on memcache so 
```
You must deploy the Memcached operator. For more information, see xref:deploying-backend-services_migrating-databases[Deploying back-end services].
```
is incorrect

bogdando: keeping preresuisites in line, helps to avoid breakages caused by moving the docs pages around. But I have no strong opinion here. Just to note that that should be a global decision to be made

docs_user/modules/proc_adopting-the-placement-service.adoc
bogdando: I agree with Sean

docs_user/modules/proc_adopting-the-placement-service.adoc
klgill: @bogdando Can I remove the prerequisite to deploy the Memcached operator? It was commented that it is incorrect.

docs_user/modules/proc_adopting-the-placement-service.adoc
klgill: I will bring this up at my next team meeting. This change would affect all documentation, and I shouldn't make this decision on my own.

PR #531, opened 2024-07-16
Delete secrets from previous installation

holser:
The secrets may have old passwords that are not compatible with current installation. This patch removes secrets that are left from previous installation

Closes: OSPRH-8599


jistr:
Looks like the CI failure is relevant.

holser:
@jistr It should be OK now

jistr (approved):

PR #536, opened 2024-07-21
Disable OVN adoption

holser:


jistr (changes requested):
IMO this is not a path forward, adoption of OVN is crucial e.g. for workload testing.

PR #290, opened 2024-02-16
Update telemetry adoption guide

yadneshk:

PR #540, opened 2024-07-25
Update the service list and its ordering to match greenfield

jistr:
References:
https://github.com/openstack-k8s-operators/openstack-operator/blob/76b853500f9f8867044db03e4fa1408539a48643/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml

https://github.com/openstack-k8s-operators/openstack-operator/blob/76b853500f9f8867044db03e4fa1408539a48643/config/samples/dataplane/ceph/values.yaml

https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/dataplane/pre_ceph_hci/values.yaml

Resolves: https://issues.redhat.com/browse/OSPRH-7875

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -157,12 +157,12 @@
         - run-os
         - reboot-os
         - install-certs
+        - ovn
+        - neutron-metadata
         {%+ if compute_adoption|bool +%}
         - libvirt
         - nova
         {%+ endif +%}
-        - ovn
-        - neutron-metadata
         - telemetry
       env:
         - name: ANSIBLE_CALLBACKS_ENABLED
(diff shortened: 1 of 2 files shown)


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -258,16 +258,16 @@
         - configure-network
         - validate-network
         - install-os
+        - ceph-hci-pre
```
jistr, line 261: @fmount Is it ok to move ceph-hci-pre here? It matches where it is in this example:

https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/dataplane/pre_ceph_hci/values.yaml#L145

jistr:
Rebased after #290

fao89 (approved):

tests/roles/dataplane_adoption/tasks/main.yaml
fmount: +1 from me.
Besides, that role is supposed to generate and apply firewall rules in preparation to Ceph deploy [1], so I think it's ok to have it after `install-os`.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ceph_hci_pre/tasks/firewall.yml

fmount (approved):
/lgtm

PR #543, opened 2024-07-25
refining glance procedures

klgill:
The procedures in the following Jiras have been refined: https://issues.redhat.com/browse/RHOSPDOC-1635
https://issues.redhat.com/browse/RHOSPDOC-1637

@fmount Please review my edits and address the questions I added on this PR.

### DIFF
--- a/docs_user/assemblies/assembly_adopting-the-image-service.adoc
+++ b/docs_user/assemblies/assembly_adopting-the-image-service.adoc
@@ -17,15 +17,11 @@ To complete the {image_service} adoption, ensure that your environment meets the
 * You have a Single Node OpenShift or OpenShift Local that is running in the {OpenShift} cluster.
 * Optional: You can reach an internal/external `Ceph` cluster by both `crc` and {OpenStackPreviousInstaller}.
 
-ifeval::["{build}" != "downstream"]
-The {image_service} adoption follows the same pattern as the {identity_service_first_ref} adoption. For more information, see xref:adopting-the-identity-service_adopt-control-plane[Adopting the {identity_service}]
-endif::[]
-
 include::../modules/proc_adopting-image-service-with-object-storage-backend.adoc[leveloffset=+1]
 
 include::../modules/proc_adopting-image-service-with-block-storage-backend.adoc[leveloffset=+1]
 
-include::../modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc[leveloffset=+1]
+include::../modules/proc_adopting-image-service-with-nfs-backend.adoc[leveloffset=+1]
 
 include::../modules/proc_adopting-image-service-with-ceph-backend.adoc[leveloffset=+1]
 
--- a/docs_user/assemblies/assembly_adopting-the-image-service.adoc
+++ b/docs_user/assemblies/assembly_adopting-the-image-service.adoc
@@ -8,8 +8,8 @@ To adopt the {image_service_first_ref} you patch an existing `OpenStackControlPl
 
 The {image_service} adoption is complete if you see the following results:
 
-* The `GlanceAPI` service up and running. 
-* The {identity_service} endpoints are updated, and the same back end of the source cloud is available. 
+* The `GlanceAPI` service up and running.
+* The {identity_service} endpoints are updated, and the same back end of the source cloud is available.
 
 .Prerequisites
 
--- a/docs_user/assemblies/assembly_adopting-the-image-service.adoc
+++ b/docs_user/assemblies/assembly_adopting-the-image-service.adoc
@@ -18,8 +18,7 @@ To complete the {image_service} adoption, ensure that your environment meets the
 * Optional: You can reach an internal/external `Ceph` cluster by both `crc` and {OpenStackPreviousInstaller}.
 
 ifeval::["{build}" != "downstream"]
-//This link goes to a 404.
-As already done for https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/keystone_adoption.md[Keystone], the Glance Adoption follows the same pattern.
+The {image_service} adoption follows the same pattern as the {identity_service_first_ref} adoption. For more information, see xref:adopting-the-identity-service_adopt-control-plane[Adopting the {identity_service}]
 endif::[]
 
 include::../modules/proc_adopting-image-service-with-object-storage-backend.adoc[leveloffset=+1]
--- a/docs_user/assemblies/assembly_adopting-the-image-service.adoc
+++ b/docs_user/assemblies/assembly_adopting-the-image-service.adoc
@@ -11,11 +11,11 @@ The {image_service} adoption is complete if you see the following results:
 * The `GlanceAPI` service up and running.
 * The {identity_service} endpoints are updated, and the same back end of the source cloud is available.
 
-.Prerequisites
+To complete the {image_service} adoption, ensure that your environment meets the following criteria:
 
-* A {OpenStackPreviousInstaller} environment (the source cloud) is running.
-* A Single Node OpenShift or OpenShift Local is running in the {OpenShift} cluster.
-* Optional: You can reach an internal/external `Ceph` cluster is by both `crc` and {OpenStackPreviousInstaller}.
+* You have a running {OpenStackPreviousInstaller} environment (the source cloud).
+* You have a Single Node OpenShift or OpenShift Local that is running in the {OpenShift} cluster.
+* Optional: You can reach an internal/external `Ceph` cluster by both `crc` and {OpenStackPreviousInstaller}.
 
 ifeval::["{build}" != "downstream"]
 //This link goes to a 404.
--- a/docs_user/assemblies/assembly_adopting-the-image-service.adoc
+++ b/docs_user/assemblies/assembly_adopting-the-image-service.adoc
@@ -4,21 +4,21 @@
 
 = Adopting the {image_service}
 
-Adopting {image_service_first_ref} means that an existing `OpenStackControlPlane` custom resource (CR), where {image_service}
-is supposed to be disabled, should be patched to start the service with the
-configuration parameters provided by the source environment.
+To adopt the {image_service_first_ref} you patch an existing `OpenStackControlPlane` custom resource (CR) that has the {image_service} disabled. The patch starts the service with the configuration parameters that are provided by the {rhos_prev_long} ({OpenStackShort}) environment.
 
-When the procedure is over, the expectation is to see the `GlanceAPI` service
-up and running: the {identity_service} endpoints are updated and the same backend of the source Cloud is available. If the conditions above are met, the adoption is considered concluded.
+The {image_service} adoption is complete if you see the following results:
 
-This guide also assumes that:
+* The `GlanceAPI` service up and running. 
+* The {identity_service} endpoints are updated, and the same back end of the source cloud is available. 
 
-* A {OpenStackPreviousInstaller} environment (the source Cloud) is running on one side.
-* A `SNO` / `CodeReadyContainers` is running on the other side.
-* (optional) An internal/external `Ceph` cluster is reachable by both `crc` and {OpenStackPreviousInstaller}.
+.Prerequisites
+
+* A {OpenStackPreviousInstaller} environment (the source cloud) is running.
+* A Single Node OpenShift or OpenShift Local is running in the {OpenShift} cluster.
+* Optional: You can reach an internal/external `Ceph` cluster is by both `crc` and {OpenStackPreviousInstaller}.
 
 ifeval::["{build}" != "downstream"]
-//This link goes to a 404. Do we need this text downstream?
+//This link goes to a 404.
 As already done for https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/keystone_adoption.md[Keystone], the Glance Adoption follows the same pattern.
 endif::[]
 
--- a/docs_user/modules/proc_adopting-image-service-with-block-storage-backend.adoc
+++ b/docs_user/modules/proc_adopting-image-service-with-block-storage-backend.adoc
@@ -2,7 +2,7 @@
 
 = Adopting the {image_service} that is deployed with a {block_storage} back end
 
-Adopt the {image_service_first_ref} that you deployed with a {block_storage_first_ref} back end in the {rhos_prev_long} ({OpenStackShort}) environment. The control plane `glanceAPI` instance is deployed with the following configuration:
+Adopt the {image_service_first_ref} that you deployed with a {block_storage_first_ref} back end in the {rhos_prev_long} ({OpenStackShort}) environment. The control plane `glanceAPI` instance is deployed with the following configuration. You use this configuration in the patch manifest that deploys the {image_service} with the block storage back end:
 
 ----
 ..
@@ -31,8 +31,7 @@ spec
 
 .Procedure
 
-. Write the patch manifest into a file, for example `glance_cinder.patch`.
-For example:
+. Create a new file, for example `glance_cinder.patch`, and include the following content:
 +
 ----
 spec:
@@ -41,6 +40,7 @@ spec:
     apiOverride:
       route: {}
     template:
+      secret: osp-secret
       databaseInstance: openstack
       storage:
         storageRequest: 10G
--- a/docs_user/modules/proc_adopting-image-service-with-block-storage-backend.adoc
+++ b/docs_user/modules/proc_adopting-image-service-with-block-storage-backend.adoc
@@ -1,8 +1,8 @@
 [id="adopting-image-service-with-block-storage-backend_{context}"]
 
-= Adopting the {image_service} that is deployed with a {block_storage} backend
+= Adopting the {image_service} that is deployed with a {block_storage} back end
 
-Adopt the {image_service_first_ref} that you deployed with an {block_storage_first_ref} backend. When {image_service} is deployed with {block_storage} as a backend in the {rhos_prev_long} environment based on {OpenStackPreviousInstaller}, the control plane `glanceAPI` instance is deployed with the following configuration:
+Adopt the {image_service_first_ref} that you deployed with a {block_storage_first_ref} back end in the {rhos_prev_long} ({OpenStackShort}) environment. The control plane `glanceAPI` instance is deployed with the following configuration:
 
 ----
 ..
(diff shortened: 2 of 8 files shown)


docs_user/assemblies/assembly_adopting-the-image-service.adoc
```diff
@@ -4,21 +4,21 @@
...
+
+* You have a running {OpenStackPreviousInstaller} environment (the source cloud).
+* You have a Single Node OpenShift or OpenShift Local that is running in the {OpenShift} cluster.
+* Optional: You can reach an internal/external `Ceph` cluster by both `crc` and {OpenStackPreviousInstaller}.
 
 ifeval::["{build}" != "downstream"]
-//This link goes to a 404. Do we need this text downstream?
+//This link goes to a 404.
 As already done for https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/keystone_adoption.md[Keystone], the Glance Adoption follows the same pattern.
 endif::[]
```
klgill, line 23: The github link goes to a 404. It is upstream only, but I thought I would point it out.

docs_user/modules/proc_adopting-image-service-with-block-storage-backend.adoc
```diff
@@ -77,8 +75,9 @@ spec:
             - storage
```
klgill, line 75: I have two questions on step 1:

1. Does the template also need “secret:” ?
2. (Hackfest feedback) Does the indentation look correct for “spec: type: LoadBalancer”?

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
```diff
@@ -25,7 +23,7 @@ In particular, find among the {OpenStackPreviousInstaller} heat templates the fo
 
 ----
 +
-In the example above, as the first variable shows, the {image_service} has no notion of NFS Ganesha backend: the `File` driver is used in this scenario, and behind the scenes, the `filesystem_store_datadir` which usually points to `/var/lib/glance/images/` is mapped to the export value provided by the `GlanceNfsShare` variable.
+In the example above, as the first variable shows, the {image_service} has no notion of NFS Ganesha back end: the `File` driver is used in this scenario, and behind the scenes, the `filesystem_store_datadir` which usually points to `/var/lib/glance/images/` is mapped to the export value provided by the `GlanceNfsShare` variable.
```
klgill, line 26: The whole section that starts with “In the example above…” seems like background information. In the prerequisites section, we usually want to keep the information focused on the action that the customer needs to take. Is this information necessary for downstream? Can we put it in its own module in this assembly?

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
```diff
@@ -153,7 +149,7 @@ Mounts:
 * Check the mountpoint:
```
klgill, line 149: “Check the mountpoint” : Can we elaborate on what customers should be looking for? For example, “Check that the mountpoint is pointing to the correct location/directory/something else?

docs_user/modules/proc_adopting-image-service-with-object-storage-backend.adoc
```diff
@@ -1,8 +1,8 @@
 [id="adopting-image-service-with-object-storage-backend_{context}"]
 
-= Adopting the {image_service} that is deployed with a {object_storage} backend
+= Adopting the {image_service} that is deployed with a {object_storage} back end
 
-Adopt the {image_service_first_ref} that you deployed with an {object_storage_first_ref} backend. When {image_service} is deployed with {object_storage_first_ref} as a backend in the {rhos_prev_long} environment based on {OpenStackPreviousInstaller}, the control plane `glanceAPI` instance is deployed with the following configuration:
+Adopt the {image_service_first_ref} that you deployed with an {object_storage_first_ref} back end in the {rhos_prev_long} ({OpenStackShort}) environment. The control plane `glanceAPI` instance is deployed with the following configuration:
```
klgill, line 6: Can we provide more context as to what this glanceAPI configuration will be used for? Something like, “You use this configuration in the patch manifest that deploys the image service with the object storage back end.”

docs_user/modules/proc_adopting-image-service-with-block-storage-backend.adoc
```diff
@@ -27,8 +27,7 @@ spec
 
 .Prerequisites
 
-* Previous Adoption steps completed. Notably, MariaDB, {identity_service_first_ref} and {key_manager_first_ref}
-should be already adopted.
+* You have completed the previous adoption steps.
 
 .Procedure
```
klgill, line 33: Step 1“Write the patch manifest into a file” - Does this mean "create a new file called, for example, `glance_swift.patch`, and paste the following content into it"?

docs_user/modules/proc_adopting-image-service-with-object-storage-backend.adoc
```diff
@@ -25,8 +25,7 @@ spec
 
 .Prerequisites
 
-* Previous Adoption steps completed. Notably, MariaDB, {identity_service_first_ref} and {key_manager_first_ref}
-should be already adopted.
+* You have completed the previous adoption steps.
 
 .Procedure
```
klgill, line 31: Step 1 “Write the patch manifest into a file” - Does this mean "create a new file called, for example, glance_swift.patch, and paste the following content into it"?
(Hackfest feedback) Step 1: The template is missing “secret:” Where does this go under “template:” and is the value “osp-secret”?

docs_user/modules/proc_adopting-image-service-with-object-storage-backend.adoc
```diff
@@ -73,8 +71,9 @@ spec:
             - storage
```
klgill, line 71: (Hackfest feedback) Does the indentation look correct for “spec: type: LoadBalancer”?

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
```diff
@@ -2,19 +2,19 @@
 
 = Verifying the {image_service} adoption
 
-Verify that you successfully adopted your {image_service_first_ref} to the {rhos_long} {rhos_curr_ver} deployment.
+Verify that you adopted the {image_service_first_ref} to the {rhos_long} {rhos_curr_ver} deployment.
 
 .Procedure
 
-. Test the {image_service_first_ref} from the {rhos_prev_long} CLI. You can compare and make sure the configuration has been correctly applied to the {image_service} pods:
+. Test the {image_service} from the {rhos_prev_long} command line interface (CLI). You can compare and ensure that the configuration is applied to the {image_service} pods:
```
klgill, line 9: Step 1: Can you clarify what are we comparing?

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
```diff
@@ -2,19 +2,19 @@
...
 +
 ----
-os-diff diff /etc/glance/glance.conf.d/02-config.conf glance_patch.yaml --frompod -p glance-api
+$ os-diff diff /etc/glance/glance.conf.d/02-config.conf glance_patch.yaml --frompod -p glance-api
 ----
 +
-If no line appears, then the configuration is correctly done.
+If no line appears, then the configuration is correct.
```
klgill, line 15: What should customers do if a line appears and the configuration is incorrect?

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
```diff
@@ -2,19 +2,19 @@
...
 
-. Inspect the resulting glance pods:
+. Inspect the resulting {image_service} pods:
 +
 ----
 GLANCE_POD=`oc get pod |grep glance-default-external-0 | cut -f 1 -d' '`
```
klgill, line 20: Comment from hackfest: "In my test environment, I see the glance pod name is "glance-default-single-0", we can think to change the grep string" 
Link for reference: https://docs.google.com/document/d/17GzJf8uTfpe7TJrEpf9NomZ98LcnNvUs2V3j9jgDw2w/edit?disco=AAABQ54c_1s

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
```diff
@@ -31,15 +31,15 @@ rbd_store_pool=images
...
 +
 ----
-oc exec -t $GLANCE_POD -c glance-api -- ls /etc/ceph
+$ oc exec -t $GLANCE_POD -c glance-api -- ls /etc/ceph
 ceph.client.openstack.keyring
 ceph.conf
 ----
 
-. Check that the service is active and the endpoints are properly updated in the {OpenStackShort} CLI:
+. Check that the service is active, and that the endpoints are updated in the {OpenStackShort} CLI:
```
klgill, line 42: (Comment from hackfest) Link to comment: https://docs.google.com/document/d/17GzJf8uTfpe7TJrEpf9NomZ98LcnNvUs2V3j9jgDw2w/edit?disco=AAABQ54c_1w

(kgilliga): I think the following link would help: https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0-beta/html-single/deploying_red_hat_openstack_services_on_openshift/index#proc_accessing-the-OpenStackClient-pod_cloud-access-admin

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
```diff
@@ -53,7 +53,7 @@ ceph.conf
 | 709859219bc24ab9ac548eab74ad4dd5 | regionOne | glance       | image        | True    | admin     | http://glance-admin-openstack.apps-crc.testing      |
```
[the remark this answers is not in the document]
fmount: I commented in that doc: no need to mention the admin endpoint, and I would just remove L53.

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: Correct, I think it's ok to mention that you can get an `openstack` cli through the `openstackclient` Pod, Can we just refer to that doc to avoid duplicating content?

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: Yes, if this part is generic (e.g. valid for all the backends), it makes sense to relax the grep and match more.
However, considering you don't have a single Pod, you can get the first one with the following:

```
GLANCE_POD=`oc get pod |grep glance-default | cut -f 1 -d' ' | head -n 1`
```

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: How can we be notified of an incorrect line? (I have no context on the `os-diff` tool and how the warnings are generated).

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: I think this refers to the resulting configuration that is applied to the Pods, but tbh I would  compare it with the applied Glance CR where we explicitly inject the configuration!
@jistr do you have any hint on this part?

docs_user/modules/proc_adopting-image-service-with-object-storage-backend.adoc
fmount: I remember I fixed that as part of https://issues.redhat.com/browse/OSPRH-7862

docs_user/modules/proc_adopting-image-service-with-object-storage-backend.adoc
fmount: Yes to both, and +1 to add the `osp-secret` template (it doesn't hurt to have it explicitly mentioned)

docs_user/modules/proc_adopting-image-service-with-object-storage-backend.adoc
fmount: +1 to have more context and I'm ok to the sentence you proposed. I think it would be enough as the following snippet providing the glance config is clear enough.

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
fmount: Check with the `mount` command that the mountpoint that points to /var/lib/glance/images  is mapped to the expected `<nfs server ip> : <nfs path>`

docs_user/assemblies/assembly_adopting-the-image-service.adoc
fmount: This is a leftover of the previous versions: can we update it to point to the keystone adoption section?

klgill: I'll update it in the next commit. It's still for upstream only. Is there any reason why downstream customers would need to know that glance adoption follows the same pattern as keystone adoption? I think it's unnecessary, but I'm curious.

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
klgill: How about : "Check that the mountpoint that points to `/var/lib/glance/images` is mapped to the expected `nfs server ip` and `nfs path` that you defined in the new default GlanceAPI instance?

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
klgill: @matbu Can you provide an answer? How does the os-diff tool indicate an incorrect configuration?

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
klgill: Actually, maybe we could do this instead:

. Check that the service is active, and that the endpoints are updated in the {OpenStackShort} CLI:
+
----
$ oc rsh openstackclient -n openstackclient
$ openstack service list | grep image

| fc52dbffef36434d906eeb99adfc6186 | glance    | image        |

$ openstack endpoint list | grep image

| 569ed81064f84d4a91e0d2d807e4c1f1 | regionOne | glance       | image        | True    | internal  | http://glance-internal-openstack.apps-crc.testing   |
| 5843fae70cba4e73b29d4aff3e8b616c | regionOne | glance       | image        | True    | public    | http://glance-public-openstack.apps-crc.testing     |
| 709859219bc24ab9ac548eab74ad4dd5 | regionOne | glance       | image        | True    | admin     | http://glance-admin-openstack.apps-crc.testing      |
----

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
klgill: Jiri is on PTO until Aug 25th. Is there someone else who would know? Or we could table it for now and clarify in another PR when he comes back.

docs_user/assemblies/assembly_adopting-the-image-service.adoc
fmount: thinking better about this, I think we don't necessarily need to refer to keystone, and the link can be removed (both upstream and downstream). When we  started this doc we had a very few services, and that link was added to highlight the pattern that is followed to enable an "adopted" service.
Feel free to revisit this part and remove that link that is not really required!

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
fmount: wait, why this mentions "Ganesha" ? the NFS share is not provided by `Ganesha` (and there's no Ceph involved here), which is only used by `Manila` only. This is a "generic" NFS service, so I would rename it and avoid mentioning Ganesha.
This is a background information that clarifies the fact that in general Glance has no notion of NFS: it uses a `File` (aka `local`) backend and the underlying filesystem points to the /var/lib/glance/images that is mapped to this generic, external, nfs share.
I think we can turn it into a `Note` if it helps, but considering that customers are doing a brownfield operation (moving from the baremetal world to openshift), it's important they understand what they are doing (without too much detail of course).
Do you think we can rework this sentence into a Note instead of another assembly (other than fixing the "Ganesha" part)?

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
fmount: Sounds good to me!

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: I agree to check this on a different PR. 
I read about the tool and this seems something that the human operator should do: in general the human operator needs to compare the 17.x Glance configuration with the one provided in the resulting/adopted Pod, and make sure it's aligned to what we had previously.
I don't see this one as a critical step for Glance, because most of the defaults are provided by the `glance-operator` and there might be different defaults compared to the previous release (e.g. glance-direct), so I prefer to focus on this detail on a different PR.

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: My concern is that we compare a 17.x config w/ a 18 config that might be different in terms of defaults injected by the `glance-operator` (and it's totally fine). I didn't try the `os-diff` tool but I'm wondering if we're able to `ignore` part of the config that would result in false positive issues.

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
fmount: Works for me (we only need to remove the glance-admin endpoint that is not provided in 18)

docs_user/modules/proc_adopting-image-service-with-nfs-ganesha-backend.adoc
klgill: OK, I changed this to a note and edited the text. I also removed references to "Ganesha" throughout the file.

fmount (approved):
/lgtm

klgill:
> /lgtm

Thanks! I sent this PR to my team for peer review.

docs_user/modules/proc_verifying-the-image-service-adoption.adoc
rheslop: I can see you added _command line interface_, it might not be needed here:  the IBM style guide says "Do not spell out an abbreviation that is commonly known to the target audience."

PR #545, opened 2024-07-29
Use os_user variable in development_environment role

archana203:
For the HA RHEV env, we are using a non-root user for accessing the source cloud.

### DIFF
--- a/tests/roles/development_environment/tasks/main.yaml
+++ b/tests/roles/development_environment/tasks/main.yaml
@@ -30,7 +30,7 @@
   no_log: "{{ use_no_log }}"
   environment:
     INSTALL_YAMLS_PATH: "{{ install_yamls_path }}"
-    CONTROLLER1_SSH="{{ controller1_ssh }}"
+    CONTROLLER1_SSH: "{{ controller1_ssh }}"
     OS_CLOUD_NAME: "{{ os_cloud_name }}"
   ansible.builtin.shell:
     cmd: |
--- a/tests/roles/development_environment/tasks/main.yaml
+++ b/tests/roles/development_environment/tasks/main.yaml
@@ -30,14 +30,12 @@
   no_log: "{{ use_no_log }}"
   environment:
     INSTALL_YAMLS_PATH: "{{ install_yamls_path }}"
-    EDPM_PRIVATEKEY_PATH: "{{ edpm_privatekey_path }}"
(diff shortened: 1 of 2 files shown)


xek (approved):
/lgtm

archana203:
recheck

afaranha (approved):

jistr (approved):

archana203:
recheck

tests/roles/development_environment/tasks/main.yaml
```diff
@@ -6,10 +6,11 @@
     EDPM_PRIVATEKEY_PATH: "{{ edpm_privatekey_path }}"
     OS_CLOUD_IP: "{{ standalone_ip | default(edpm_node_ip) }}"
     OS_CLOUD_NAME: "{{ os_cloud_name }}"
+    OS_USER: "{{ os_user }}"
```
jistr, line 9: There is a "OS_USERNAME" variable which means "the user account in OpenStack that is used for API calls". To avoid confusion with that one, i think we should probably call this something like "EDPM_SSH_USER"? But let's first run through the test whether this can fix the RHEL job before we rename? Or i'd be even ok landing this and then renaming in a follow-up PR.

jistr: Actually yes let's land this and rename in a follow-up. Sorry i didn't realize this on the first pass of the review :)

archana203: EDPM_SSH_USER seems to be an user used to ssh compute nodes. Here, we are doing ssh to the node from where we can access the openstack cloud configs to generate a token.

jistr: Fair enough, but personally i'd still at least include "SSH" later, to not be too close to the openstack client vars: https://docs.openstack.org/newton/user-guide/common/cli-set-environment-variables-using-openstack-rc.html 

Not something that needs to change right now anyway :)

jistr:
recheck

tests/roles/development_environment/tasks/main.yaml
archana203: OS_USER_SSH sounds good though

ciecierski: I echo with what Jiri commented, anything with OS_ prefix should be reserved only to OpenStack related environment variable.

archana203: What about OSP_SSH_USER ?

PR #546, opened 2024-07-29
Fix adoption fernet session token test

afaranha:
https://issues.redhat.com/browse/OSPRH-6854

### DIFF
--- a/tests/roles/keystone_adoption/defaults/main.yaml
+++ b/tests/roles/keystone_adoption/defaults/main.yaml
@@ -6,6 +6,9 @@ keystone_patch: |
       apiOverride:
         route: {}
       template:
+        customServiceConfig: |
+          [token]
+          expiration = 360000
         override:
           service:
             internal:
--- a/tests/roles/keystone_adoption/tasks/main.yaml
+++ b/tests/roles/keystone_adoption/tasks/main.yaml
@@ -69,7 +69,7 @@
       ${BASH_ALIASES[openstack]} service list | awk "/ $service /{ print \$2; }" | xargs -r ${BASH_ALIASES[openstack]} service delete || true
     done
 
-- name: debug token
+- name: Print session test token
   ansible.builtin.debug:
     var: before_adoption_token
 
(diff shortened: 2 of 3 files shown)


afaranha:
recheck

afaranha:
recheck

tests/roles/backend_services/templates/openstack_control_plane.j2
```diff
@@ -57,8 +57,10 @@ spec:
       ironicConductors: []
 
   keystone:
-    enabled: false
-    template: {}
+    template:
```
fmount, line 60: Restore this file to have `enabled: false`.
Keystone should be patched during the adoption phase, which means you have to apply that config here [1]

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/keystone_adoption/defaults/main.yaml

tests/roles/keystone_adoption/tasks/main.yaml
```diff
@@ -73,13 +73,32 @@
...
+    var: test_token
+
 - name: Verify that pre-adoption token still works
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    OS_TOKEN={{ before_adoption_token.stdout }}
-    alias openstack="oc exec -t openstackclient -- openstack"
-    ${BASH_ALIASES[openstack]} endpoint list
+    oc exec -n openstack -t openstackclient -- bash -c 'curl -s -I -H "X-Auth-Token: {{ before_adoption_token.stdout }}" https://keystone-public-openstack.apps-crc.testing/v3/users'
```
fmount, line 99: This will represent an issue when you're working in an environment != crc. 
The endpoint should either be retrieved by the existing ctlplane, or ... it might be parametrized in the form of:
`<urischeme | default('https')>://keystone-public-<namespace>.<cluster_domain>/v3/users`

fmount (changes requested):

tests/roles/keystone_adoption/defaults/main.yaml
```diff
@@ -6,6 +6,10 @@ keystone_patch: |
       apiOverride:
         route: {}
       template:
+        KeystoneAPI:
```
xek, line 9: I'm not sure, but it looks to me like the  customServiceConfig should be directly below "template:", so this line is not needed. This is how it works in glance, ex.: https://github.com/openstack-k8s-operators/glance-operator/blob/f66b05484df8df1df9cba1fc7560fbd2f4ce04ae/config/samples/backends/nfs/nfs.yaml#L10

fmount: correct, this needs to be changed and stay under template.

tests/roles/keystone_adoption/tasks/main.yaml
xek: There is {{ auth_url }} in the config, so I used that.

afaranha:
recheck

xek:
recheck

xek:
recheck

fmount (approved):
/lgtm

jistr (approved):

PR #548, opened 2024-08-01
Add OK / CHECK FAILED output to MariaDB checks

jistr:
Resolves: https://issues.redhat.com/browse/OSPRH-7859

### DIFF
--- a/tests/roles/mariadb_copy/templates/post_checks.bash
+++ b/tests/roles/mariadb_copy/templates/post_checks.bash
@@ -8,19 +8,19 @@
 {% endif %}
 
 # use 'oc exec' and 'mysql -rs' to maintain formatting
-dbs=$(oc exec openstack-galera-0 -c galera -- mysql -rs -uroot -p"$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;')
-echo $dbs | grep -Eq '\bkeystone\b'
+dbs=$(oc exec openstack-galera-0 -c galera -- mysql -rs -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;')
+echo $dbs | grep -Eq '\bkeystone\b' && echo "OK" || echo "CHECK FAILED"
 
 # ensure neutron db is renamed from ovs_neutron
 echo $dbs | grep -Eq '\bneutron\b'
-echo $PULL_OPENSTACK_CONFIGURATION_DATABASES | grep -Eq '\bovs_neutron\b'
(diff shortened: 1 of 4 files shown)


jistr:
Ready to merge i think.

fao89 (approved):

tests/roles/mariadb_copy/templates/post_checks.bash
```diff
@@ -8,19 +8,19 @@
 {% endif %}
 
 # use 'oc exec' and 'mysql -rs' to maintain formatting
-dbs=$(oc exec openstack-galera-0 -c galera -- mysql -rs -uroot -p"$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;')
-echo $dbs | grep -Eq '\bkeystone\b'
+dbs=$(oc exec openstack-galera-0 -c galera -- mysql -rs -uroot "-p$PODIFIED_DB_ROOT_PASSWORD" -e 'SHOW databases;')
```
ciecierski, line 11: "-p$PODIFIED_DB_ROOT_PASSWORD" ?

holser: -p"$PODIFIED_DB_ROOT_PASSWORD"

jistr: There is no difference between these two approaches in Bash.

jistr: The reason this is edited in this patch is that i re-copy-pasted what was in docs into the tests, to make them 100% in sync again. Both of the approaches that you guys quoted work the same, they're both correct AFAIK.

holser (approved):

tests/roles/mariadb_copy/templates/post_checks.bash
jistr: Actually the `"-p$PODIFIED_DB_ROOT_PASSWORD"` variant is probably the best because `-p"$PODIFIED_DB_ROOT_PASSWORD"` might make the admin tempted to put a space after the -p like `-p "$PODIFIED_DB_ROOT_PASSWORD"` and i think that would break. IIRC the mariadb client requires the password to be included in the same command line argument as the `-p` flag, not split into two by a space.

jistr:
Rebasing before the merge because the CI result is almost a month old.

tests/roles/mariadb_copy/templates/post_checks.bash
ciecierski: ack, thank you

PR #553, opened 2024-08-06
Refactor adoption Clean Up procedure

holser:
Sync clean up with install_yaml recomendations

Closes-Bug: OSPRH-9120

### DIFF
--- a/tests/roles/pcp_cleanup/tasks/main.yaml
+++ b/tests/roles/pcp_cleanup/tasks/main.yaml
@@ -2,28 +2,35 @@
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
+    # Cleanup OpenStackControlPlane object
+    oc delete --ignore-not-found=true osctlplane -l core.openstack.org/openstackcontrolplane -n openstack
 
-    oc delete --ignore-not-found=true --wait=false openstackdataplanedeployment/openstack
-    oc delete --ignore-not-found=true --wait=false openstackdataplanedeployment/openstack-nova-compute-ffu
-    oc delete --ignore-not-found=true --wait=false openstackcontrolplane/openstack
-    oc patch openstackcontrolplane openstack --type=merge --patch '
-    metadata:
-      finalizers: []
-    ' || true
-
-    while oc get pod | grep rabbitmq-server-0; do
-        sleep 2
-    done
-    while oc get pod | grep openstack-galera-0; do
-        sleep 2
+    # Ensure that all pods in openstack namespace are deleted
+    while oc get pod -n openstack | grep -E 'rabbitmq-server-0|openstack-galera-0'; do
+      sleep 2;
     done
 
+    # Cleanup OpenStackDataplane objects
+    oc delete --ignore-not-found=true OpenStackDataPlaneDeployment --all -n openstack
+    oc delete --ignore-not-found=true OpenStackDataPlaneNodeSet --all -n openstack
+    oc delete --ignore-not-found=true OpenStackDataPlaneService --all -n openstack
+
+    # Delete Adoption helper pods
     oc delete --ignore-not-found=true --wait=false pod mariadb-copy-data
     oc delete --ignore-not-found=true --wait=false pvc mariadb-data
     oc delete --ignore-not-found=true --wait=false pod ovn-copy-data
-    oc delete --ignore-not-found=true secret osp-secret
-    oc delete issuer rootca-internal --ignore-not-found
-    oc delete secret rootca-internal --ignore-not-found
+
+    # Delete secrets
+    for secret in $(oc -o name get secrets -n openstack); do
+      echo "Deleting secret ${secret}";
+      #(TODO: holser) oc patch can be removed when OSPRH-10262 is merged
+      oc patch ${secret} -p '{"metadata":{"finalizers":[]}}' --type=merge
+      oc delete ${secret} -n openstack;
+    done
+
+    oc delete --ignore-not-found issuer rootca-internal
+    oc delete --ignore-not-found secret rootca-internal
+
     oc delete subscription cluster-observability-operator -n openshift-operators --ignore-not-found
   when: pcp_cleanup_enabled|bool
 


holser (commented):
recheck

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -2,6 +2,16 @@
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
+    cd {{ install_yamls_path }}
+    # Delete the operator, but does not cleanup the service resources
+    make openstack_cleanup
```
cescgina, line 7: does this work ok? I've always done `openstack_deploy_cleanup` before `openstack_cleanup`

gibizer: if you remove the operators first then the below resource cleanup wont be complete as the operators won't do the cleanup tasks. So I think it is wrong.

jistr (changes requested):
As discussed on Slack this is unlikely to work in general. The `openstack_deploy_cleanup` target depends on having the CR YAMLs generated into `install_yamls/out/...`, and since during Adoption we don't use install_yamls to deploy, those files won't be generated. We should continue to rely on explicit `oc delete` cleanup for the things that the test suite created.

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -2,29 +2,49 @@
...
+    oc delete --ignore-not-found=true secret -l libvirt-secret -n openstack
+    oc delete --ignore-not-found=true secret -l octavia-ca-passphrase -n openstack
+
+    oc delete --ignore-not-found issuer rootca-internal
+    oc delete --ignore-not-found secret rootca-internal
+
     oc delete subscription cluster-observability-operator -n openshift-operators --ignore-not-found
+
+    # Delete openstack namespace
+    oc delete --ignore-not-found=true namespace openstack
```
gibizer, line 47: the overall cleanup process look good to me. Technically just doing the openstack namespace delete should be enough but I heard that is it is not always stable.

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -2,29 +2,49 @@
...
-      finalizers: []
-    ' || true
-
-    while oc get pod | grep rabbitmq-server-0; do
-        sleep 2
-    done
-    while oc get pod | grep openstack-galera-0; do
-        sleep 2
+    # Ensure that all pods in openstack namespace are deleted
+    while oc get pod -n openstack | grep -E 'rabbitmq-server-0|openstack-galera-0'; do
```
stuggi, line 9: checking for galera is probably ok, there might a situation where the deployment  is gone, but the finalizer on the CR got not removed, which the above you not catch. 

I think checking that there are no openstack related CRs, like the output of the following should not have a result:

```
for i in $(oc get crd -o custom-columns=Name:metadata.name| grep -E "openstack|rabbit"); do oc get $i -o name; done
```

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -2,29 +2,49 @@
...
+    # Cleanup OpenStackDataplane objects
+    oc delete --ignore-not-found=true OpenStackDataPlaneDeployment --all -n openstack
+    oc delete --ignore-not-found=true OpenStackDataPlaneNodeSet --all -n openstack
+    oc delete --ignore-not-found=true OpenStackDataPlaneService --all -n openstack
+
+    # Delete Dataplane CRDs
+    oc delete crd --ignore-not-found=true openstackdataplanedeployments.dataplane.openstack.org
+    oc delete crd --ignore-not-found=true openstackdataplanenodesets.dataplane.openstack.org
+    oc delete crd --ignore-not-found=true openstackdataplaneservices.dataplane.openstack.org
+
```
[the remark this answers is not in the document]
stuggi: what is meant with between runs? between rerun of a full update job? 

before we deploy the initial environment to which should get updated, we probably want to delete the CRDs from a previous run and it might be required, e.g. in the case where we do an update to an operator which introduces v1beta2. If you start with an operator version which only supports v1beta1 you do not want to have a v1beta2 from an older job installed. I think this version would not be removed when you install the operator which won't support it.

When you update the operators for a minor update you can not delete the CRDs. if you delete CRDs the CRs get deleted as well and your deployment is gone, yes a finalizer might block it. The updated operator will handle installing CRD changes for new versions like @gibizer said.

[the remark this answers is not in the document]
holser: "Between runs" means between 2 CI runs... It's full update job run eventually.

Our job doesn't deploy OCP on the run as developers crc. It reuses static OCP that is deployed once outside of the jobs. When CI is triggered, it will deploy OSP platform and will deploy operators on provided OCP. When job fails it means that all operators, crds/crs, csvs, secrets are left. When clean up is not perform the next CI run will have a failure. When I debug it I got the following situation

```
oc describe subscription openstack-operator

    Message:                error updating CRD "instancehas.instanceha.openstack.org": CustomResourceDefinition.apiextensions.k8s.io "instancehas.instanceha.openstack.org" is invalid: spec.names.kind: Invalid value: "InstanceHa": field is immutable
    Reason:                 InstallComponentFailed
```
I was able to fix it by running 
```
oc delete crd instancehas.instanceha.openstack.org
```
and triggering CI again.

I understand minor update case. However, the job I am working on is not minor update. It's adoption job. The job deploys OSP 17.1 and tries to adopt environment to 18.0 upstream master content. If crd structure is changed the operator will fail as it won't be able to patch. See error above.

tests/roles/pcp_cleanup/tasks/main.yaml
holser: I implemented crd/cr removal in the job. So, I removed crd/cr removal from this job

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -2,28 +2,40 @@
...
+    # Delete Adoption pods
     oc delete --ignore-not-found=true --wait=false pod mariadb-copy-data
     oc delete --ignore-not-found=true --wait=false pvc mariadb-data
     oc delete --ignore-not-found=true --wait=false pod ovn-copy-data
-    oc delete --ignore-not-found=true secret osp-secret
-    oc delete issuer rootca-internal --ignore-not-found
-    oc delete secret rootca-internal --ignore-not-found
+
+    # Delete secrets
```
odyssey4me, line 26: This seems like it would need constant updating whenever there are new secrets. Can we not list the secrets and then loop through them?

gibizer: The way out of this is to delete the namespace so then all the resources in the namespace is deleted. Today the feeling is that deleting the namespace is not always works properly. Personally I think that is a bug and should be fixed. (And we fixed issues before in this area). But I don't have the details to how namespace deletion fails (It worked for me last time when I tried locally) so I cannot file a Jira bug tracking it.

tests/roles/pcp_cleanup/tasks/main.yaml
```diff
@@ -2,28 +2,35 @@
...
+    for secret in $(oc -o name get secrets -n openstack); do
+      echo "Deleting secret ${secret}";
+      #(TODO: holser) oc patch can be removed when OSPRH-10262 is merged
+      oc patch ${secret} -p '{"metadata":{"finalizers":[]}}' --type=merge
+      oc delete ${secret} -n openstack;
```
odyssey4me, line 28: Should there be some sort of wait/until condition before deleting the namespace, or does k8s serialise the process automagically?

holser: There shouldn't be wait as openshift is stuck deleting object if it cannot delete it. It actually checks it that object is definitely deleted. That's what OSPRH-10262 is about when we delete resource (secret) but "oc delete" waits for actual termination forever.

odyssey4me (approved):

stuggi (commented):
lgtm

holser:
recheck

holser:
recheck

tests/roles/pcp_cleanup/tasks/main.yaml
gibizer: @holser sure. Thanks for the bug report OSPRH-10262 Please keep sending bugs when you see a stuck namespace deletion.

jistr (approved):

PR #558, opened 2024-08-07
Make a switcher to disable metric storage

holser:
Metric Storage requires persistence storage which is not always possible in test environments. This patch makes a switcher to allow disable metric storage until persistance/ephemeral switcher is inmplemented in telemetry controller per [1]

[1] https://issues.redhat.com/browse/OSPRH-9252

Closes-Bug: OSPRH-9169

### DIFF
--- a/tests/roles/common_defaults/defaults/main.yaml
+++ b/tests/roles/common_defaults/defaults/main.yaml
@@ -11,3 +11,6 @@ use_no_log: false
 
 # Whether the adopted node will host compute services
 compute_adoption: true
+
+# Disable metric storage until ephemeral storage is implemented by setting to false
+telemetry_metric_storage: true
--- a/tests/roles/telemetry_adoption/tasks/main.yaml
+++ b/tests/roles/telemetry_adoption/tasks/main.yaml
@@ -27,6 +27,7 @@
   delay: 2
 
 - name: deploy metric storage
+  when: telemetry_metric_storage|bool
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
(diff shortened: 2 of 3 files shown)


mrunge:
What is supposed to happen with configured telemetry if there is no storage provided?

holser:
> What is supposed to happen with configured telemetry if there is no storage provided?

There is no configured telemetry. Also there is no data migration between 17.1 telemetry and 18.0. It means that telemetry is installed from scratch in 18.0. If we disable Metric Storage this component won't work which is fine until ephemeral storage is implemented.

odyssey4me (changes requested):
Firstly, the review does not disable metric storage. There are mixed messages all through this PR saying that metric storage is disabled, but the settings say otherwise. I suggest being clear and consistent throughout.

Secondly, it's not clear to me why/how ephemeral metric storage makes any sense. To my mind, if there is not adoption of metrics, then surely we should be able to just skip it as a service and it can be a post-adoption action to set it up? Is there a reason why we cannot just do that?

Thirdly, if it is not possible to skip the metric setup because it's a default service or something like that, then I suggest fixing the test environment rather than changing the adoption flow because the test environment has a deficiency.

jlarriba:
The idea of entering adoption was that if the user adopts the data plane, we want telemetry to be installed there by default. Can it be enabled as a day-2 post adoption service? Sure, but after the multitude of environments without metrics that we found on 17.1 and previous (telemetry being a day-2 optional step), there has been a strong push from product management to get telemetry enabled by default in any circumstance.

Adopting an existent control plane "should" entail deploying the telemetry control plane and the telemetry agents on the new adopted adataplanes by default, and thats why we have commited this.

This is definitively debatable, but this is the reason.

PR #562, opened 2024-08-13
Delete the subscription created during telemetry adoption

archana203:
To avoid subscriptions.operators.coreos.com \"cluster-observability-operator\" already exists while re-running the adoption on pre-deployed OCP cluster.


fao89 (approved):

jistr (approved):

jistr:
I was unsure whether we should be undoing subscriptions in the cleanup, but since this subscription is indeed created during the execution of the tests, i think it's ok to remove it. We can follow up if this causes too much churn in e.g. dev environments.

PR #566, opened 2024-08-14
Add edpm_container_registry_insecure_registries var

cescgina:
Add edpm_container_registry_insecure_registries to the dataplane nodeset
to be able to pull images from temporal registries like we do in the
component pipeline.


cescgina:
recheck

cescgina:
recheck

fao89 (approved):

PR #568, opened 2024-08-19
Make telemetry adoption conditional.

holser:
Since there are no real data migration telemetry role just creates a new telemetry installation. As it doesn't do any data migration it's not a mandatory for adoption. This patch makes a condition when telemetry can be turned on and off (default).


fao89 (approved):

holser:
recheck

holser:
recheck

holser:
recheck

holser:
recheck

holser:
recheck

PR #570, opened 2024-08-20
Add networker dataplane

fyanac:


tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -81,11 +81,27 @@ edpm_nodes:
         subnetName: subnet1
       - name: storagemgmt
         subnetName: subnet1
+edpm_nodes_networker:
```
cescgina, line 84: I'm not sure we should add this default. IIUC we won't test the networker use case with standalone right? In that case this variable would be wrong, and the code is already checks if `edpm_nodes_networker` is defined, so it should not break other use cases

slawqo: this is correct, with standalone we don't test networked nodes case

tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -81,11 +81,27 @@ edpm_nodes:
...
+  standalone:
+    hostName: "{{ edpm_node_networker_hostname }}"
+    ansible:
+      ansibleHost: "{{ edpm_node_networker_ip }}"
```
marios, line 88: where is that coming from
in https://review.rdoproject.org/r/c/rdo-jobs/+/54113/ you have edpm_node_networker_ips but it isn't being set there, so probably it comes from the job definition?

but what about edpm_node_networker_ip ? it also is not being set in 54113

jistr (approved):
LGTM :+1:

cescgina (approved):

PR #576, opened 2024-08-28
Add variables for services configuration patches

eshulman2:
Added variables for services configuration patches to allow overridding them and use different configuration


tests/roles/ironic_adoption/defaultes/main.yaml
```diff
@@ -0,0 +1,75 @@
...
+                    metallb.universe.tf/loadBalancerIPs: 172.17.0.80
+                spec:
+                  type: LoadBalancer
+        ironicConductors:
+        - replicas: 1
+          networkAttachments:
+          - baremetal
+          provisionNetwork: baremetal
+          storageRequest: 10G
+          storageClass: {{ storage_class_name }}
```
jistr, line 27: We may need to add a default for storage_class_name too, to make sure it is already defined when the ironic_patch value is being processed. I think when the template was in `tasks.yaml` it relied on the variable being included from OVN role defaults, but that may not be a reliable way anymore. We should probably copy the default from the OVN file into this file too. https://github.com/openstack-k8s-operators/data-plane-adoption/blob/eca663af5a80df64a6b78388ce5a7d9181615fa0/tests/roles/ovn_adoption/defaults/main.yaml#L2

pinikomarov:
right , this looks good to me

tests/roles/ironic_adoption/defaultes/main.yaml
```diff
@@ -0,0 +1,77 @@
+---
```
jistr, line 1: Just noticed a typo: the directory is called "defaultes". Ironic adoption is not tested in CI so if you want to have CI feedback, you can let the current CI jobs finish. We should just fix this before merging.

eshulman2: Completely missed it, thanks will fix after CI

eshulman2: fixed this and the neutron typo that failed the gate

jistr (approved):

PR #579, opened 2024-08-29
[pre_launch script] Update public subnet allocation pool

karelyatin:
Current allocation is quite low, increase it to also match what being done with os_net_setup[1].

Also enabled dhcp and update subnet name.

Related change:- https://github.com/openstack-k8s-operators/install_yamls/pull/906 [1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/os_net_setup/defaults/main.yml


jistr (approved):
Thanks

holser (approved):
LGTM

PR #581, opened 2024-09-02
Add notes for finding the SOURCE_MARIADB_IP from openstack cloud

archana203:
Ref: [OSPRH-7860](https://issues.redhat.com/browse/OSPRH-7860)


jistr (approved):
I was wondering whether we should perhaps have the explanation only at the first occurrence of the variable, but i don't have a strong opinion one way or another. +1

klgill (approved):

klgill:
LGTM. Since the variable appears in 2 separate procedures, I think it's a good idea to add the explanation to both occurrences. If it appeared twice in the same procedure, I'd suggest including the explanation only once.

PR #583, opened 2024-09-03
refined configuring ceph back end procedure

klgill:
https://issues.redhat.com/browse/RHOSPDOC-1644

### DIFF
--- a/docs_user/modules/proc_configuring-a-ceph-backend.adoc
+++ b/docs_user/modules/proc_configuring-a-ceph-backend.adoc
@@ -74,7 +74,7 @@ stringData:
     mon_host = 10.1.1.2,10.1.1.3,10.1.1.4
 ----
 
-. In your `OpenStackControlPlane` CR, configure the {Ceph} back end from your {OpenStackShort} deployment by using the `extraMounts` key. For example:
+. In your `OpenStackControlPlane` CR, inject `ceph.conf` and `ceph.client.openstack.keyring` to the {OpenStackShort} services that are defined in the propagation list. For example:
 +
 [source,yaml]
 ----
--- a/docs_user/modules/proc_configuring-a-ceph-backend.adoc
+++ b/docs_user/modules/proc_configuring-a-ceph-backend.adoc
@@ -1,40 +1,30 @@
 [id="configuring-a-ceph-backend_{context}"]
 
-= Configuring a Ceph backend
+= Configuring a {Ceph} back end
 
-If the original deployment uses a Ceph storage backend for any service
-(e.g. {image_service_first_ref}, {block_storage_first_ref}, {compute_service_first_ref}, {rhos_component_storage_file_first_ref}), the same backend must be used in the
-adopted deployment and custom resources (CRs) must be configured accordingly.
-
-If you use {rhos_component_storage_file_first_ref}, on {OpenStackPreviousInstaller} environments, the CephFS driver in {rhos_component_storage_file} is configured to use
-its own keypair. For convenience, modify the `openstack` user so that you
-can use it across all {rhos_prev_long} services.
-
-Using the same user across the services serves two purposes:
-
-* The capabilities of the user required to interact with {rhos_component_storage_file}
-became far simpler and hence, more became more secure with {rhos_acro} {rhos_curr_ver}.
-* It is simpler to create a common ceph secret (keyring and ceph config
-file) and propagate the secret to all services that need it.
+If your {rhos_prev_long} ({OpenStackShort}) {rhos_prev_ver} deployment uses a {Ceph} back end for any service, such as {image_service_first_ref}, {block_storage_first_ref}, {compute_service_first_ref}, or {rhos_component_storage_file_first_ref}, you must configure the custom resources (CRs) to use the same back end in the {rhos_long} {rhos_curr_ver} deployment.
 
 [TIP]
-To run `ceph` commands, you must use SSH to connect to a Ceph
-storage node and run `sudo cephadm shell`. This brings up a ceph orchestrator
-container that allows you to run administrative commands against the ceph
-cluster. If you deployed the ceph cluster by using {OpenStackPreviousInstaller}, you may launch the `cephadm` shell from an {OpenStackShort} controller node.
+If you use {rhos_component_storage_file} on {OpenStackPreviousInstaller} environments, the CephFS driver in {rhos_component_storage_file} is configured to use its own key pair. For convenience, modify the `openstack` user so that you
+can use it across all {OpenStackShort} services.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_configuring-a-ceph-backend.adoc
```diff
@@ -1,40 +1,30 @@
 [id="configuring-a-ceph-backend_{context}"]
 
-= Configuring a Ceph backend
+= Configuring a {Ceph} back end
```
klgill, line 3: @fmount  Should we be referring to the back end as the "Ceph back end" or the "Red Hat Ceph Storage back end"?

docs_user/modules/proc_configuring-a-ceph-backend.adoc
```diff
@@ -87,11 +77,11 @@ stringData:
     mon_host = 10.1.1.2,10.1.1.3,10.1.1.4
 ----
 
-. Configure `extraMounts` within the `OpenStackControlPlane` CR:
+. In your `OpenStackControlPlane` CR, configure the {Ceph} back end from your {OpenStackShort} deployment by using the `extraMounts` key. For example:
```
klgill, line 80: Do customers need to adjust the `extraMounts` configuration to the values in their environment? That is, is the code in this step just an example, or are they supposed to use these exact values?

docs_user/modules/proc_configuring-a-ceph-backend.adoc
fmount: I think downstream we should keep referring to RHCS [1], while Ceph is sufficient for the upstream doc.
Do we want to put an `if` statement to make this distinction more explicit? 

[1] https://docs.redhat.com/en/documentation/red_hat_ceph_storage/7

docs_user/modules/proc_configuring-a-ceph-backend.adoc
```diff
@@ -1,40 +1,30 @@
...
-file) and propagate the secret to all services that need it.
+If your {rhos_prev_long} ({OpenStackShort}) {rhos_prev_ver} deployment uses a {Ceph} back end for any service, such as {image_service_first_ref}, {block_storage_first_ref}, {compute_service_first_ref}, or {rhos_component_storage_file_first_ref}, you must configure the custom resources (CRs) to use the same back end in the {rhos_long} {rhos_curr_ver} deployment.
 
 [TIP]
-To run `ceph` commands, you must use SSH to connect to a Ceph
-storage node and run `sudo cephadm shell`. This brings up a ceph orchestrator
-container that allows you to run administrative commands against the ceph
-cluster. If you deployed the ceph cluster by using {OpenStackPreviousInstaller}, you may launch the `cephadm` shell from an {OpenStackShort} controller node.
+If you use {rhos_component_storage_file} on {OpenStackPreviousInstaller} environments, the CephFS driver in {rhos_component_storage_file} is configured to use its own key pair. For convenience, modify the `openstack` user so that you
+can use it across all {OpenStackShort} services.
```
fmount, line 9: Do we want to call out the reason of this choice? One important reason is that `Manila`, which is used to be configured with its own Ceph keyring in 17,  doesn't require more privileges anymore compared to other OpenStack services backed by Ceph, hence using a single, openstack, keyring is simpler.

docs_user/modules/proc_configuring-a-ceph-backend.adoc
```diff
@@ -1,40 +1,30 @@
...
 
+[NOTE]
```
fmount, line 11: Not sure this should be a Note or the "procedure" that should be followed to update the openstack keyring.

docs_user/modules/proc_configuring-a-ceph-backend.adoc
fmount: They can use the following snippet as it is.
The only difference might be in the `propagation` list, which depends on the services that are backed by Ceph.
The snippet assumes the ceph.conf and keyring are propagated everywhere.
Note that this part that uses `extraMounts` is used to inject both `ceph.conf` and `ceph.client.openstack.keyring` to the Pods, but this is not providing the ceph backend configuration for each service: that part is present and defined for each service in their own adoption section.

docs_user/modules/proc_configuring-a-ceph-backend.adoc
klgill: We don't need the if statement. The {Ceph} attribute renders as "Ceph" upstream, while downstream it renders as "Red Hat Ceph Storage".

docs_user/modules/proc_configuring-a-ceph-backend.adoc
klgill: Should the note that starts with "To run ceph commands..." be a prerequisite, or the first step of the main procedure?

fmount: This command is not strictly connected to the order of services redeploy in the ctlplane, it can be treated as a prerequisite if Manila is part of the overcloud in the src cloud.
We can probably move it as a prerequisite rather then keeping it as a Note, and highlight that if Manila is part of the deployment, the openstack keyring must be updated with this `<command>` for this `<reason>`

klgill: This is the latest iteration of the introduction and prerequisites. It might be easier to see it all together in the preview: https://logserver.rdoproject.org/83/583/3b10bf2e3f08239c3c142ecc687247a528958938/github-check/adoption-docs-preview/2fb2d12/docs_build/docs_build/adoption-user/index-downstream.html#configuring-a-ceph-backend_migrating-databases

docs_user/modules/proc_configuring-a-ceph-backend.adoc
fmount: RHOSO Pods is generic enough to let the reader resolve it into a `CR`, a `Service`, a `Component` within a `Service`, a `Container`, etc.
To be consistent with the `extraMounts` wording, I would say "inject `ceph.conf` and `ceph.client.openstack.keyring` to the `OpenStack services` defined in the `propagation` list".
@klgill the propagation list is clearly visible to the example below, and contains the services that need those files because they're supposed to be configured with Ceph.

docs_user/modules/proc_configuring-a-ceph-backend.adoc
fmount: Yeah, in theory this is not a mandatory step for users that have reasons to keep the keyring separated. In practice, for greenfield, we support a single keyring for the openstack services, so keeping a different keyring might result in issues, and we don't even document that kind of use case, so this make it not really optional. I would remove that wording if it helps, and ask to make this change anyway.

docs_user/modules/proc_configuring-a-ceph-backend.adoc
fmount: Ok, thank you for sharing the preview. 
Definitely I think it's good to have this step as a prerequisite.

PR #584, opened 2024-09-04
Updating control_plane_rollback_verify_command var

frenzyfriday:


cescgina (approved):

tests/roles/control_plane_rollback/defaults/main.yaml
```diff
@@ -1,2 +1,2 @@
 control_plane_rollback_verify_command: |
-  $CONTROLLER1_SSH OS_CLOUD={{ os_cloud_name }} openstack user list
+  ssh root@{{ standalone_ip }} OS_CLOUD={{ os_cloud_name }} openstack user list
```
jistr, line 2: I'm wondering about this change, why is it needed and if it will be suitable for all envs. But anyway it's not a blocker, we can adjust if needed.

jistr (approved):

tests/roles/control_plane_rollback/defaults/main.yaml
cescgina: @jistr right now we needed this change because we need to run the openstack command from the undercloud and not a controller

PR #585, opened 2024-09-04
refined rollback procedure

klgill:
https://issues.redhat.com/browse/RHOSPDOC-1986

### DIFF
--- a/docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
+++ b/docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
@@ -2,27 +2,19 @@
 
 = Rolling back the control plane adoption
 
-If you encountered a problem during the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services that prevents you from completing the adoption procedure, you can roll back the control plane adoption.
+If you encountered a problem that prevents you from completing the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services, you can roll back the control plane adoption.
 
 [IMPORTANT]
-The roll back operation is only possible during the
-control plane parts of the adoption procedure. If you altered the data plane
-nodes in any way during the procedure, the roll back is not possible.
+You can only roll back the control plane adoption during the parts of the adoption procedure that alter the control plane. If you altered the data plane
+nodes in any way, you cannot perform the rollback procedure.
 
-During the control plane adoption, services on the source cloud's
-control plane are stopped but not removed. The databases on the source
-control plane are not edited by the adoption procedure. The
-destination control plane received a copy of the original
-control plane databases. The roll back procedure assumes that the data
-plane has not yet been touched by the adoption procedure and it is
-still connected to the source control plane.
+During the control plane adoption, services on the {OpenStackShort} control plane are stopped but not removed. The databases on the {OpenStackShort} control plane are not edited during the adoption procedure. The {rhos_long} control plane receives a copy of the original control plane databases. The rollback procedure assumes that the data plane has not yet been modified by the adoption procedure, and it is still connected to the {OpenStackShort} control plane.
 
 The rollback procedure consists of the following steps:
 
-* Restoring the functionality of the source control plane.
+* Restoring the functionality of the {OpenStackShort} control plane.
 
-* Removing the partially or fully deployed destination
-  control plane.
+* Removing the partially or fully deployed {rhos_acro} control plane.
 
 .Procedure
 
@@ -138,26 +130,23 @@ for i in {1..3}; do
 done
 ----
 
-. If the Ceph NFS service is running on the deployment as a {rhos_component_storage_file_first_ref}
-backend, you must restore the pacemaker ordering and colocation constraints
-involving the "openstack-manila-share" service:
+. If the Ceph NFS service is running on the deployment as a {rhos_component_storage_file_first_ref} back end, you must restore the Pacemaker order and colocation constraints for the `openstack-manila-share` service:
 +
 ----
 
-sudo pcs constraint order start ceph-nfs then openstack-manila-share kind=Optional id=order-ceph-nfs-openstack-manila-share-Optional
-sudo pcs constraint colocation add openstack-manila-share with ceph-nfs score=INFINITY id=colocation-openstack-manila-share-ceph-nfs-INFINITY
+$ sudo pcs constraint order start ceph-nfs then openstack-manila-share kind=Optional id=order-ceph-nfs-openstack-manila-share-Optional
+$ sudo pcs constraint colocation add openstack-manila-share with ceph-nfs score=INFINITY id=colocation-openstack-manila-share-ceph-nfs-INFINITY
 
 ----
 
-. Verify that the source cloud is operational again, e.g. by
-running `openstack` CLI commands or using the {dashboard_first_ref}.
-//kgilliga: Should we be more specific? Which CLI commands?
+. Verify that the source cloud is operational again, for example, you
+can run `openstack` CLI commands or check that you can access the {dashboard_first_ref}.
 
-. Remove the partially or fully deployed control plane so that another adoption attempt can be made later:
+. Remove the partially or fully deployed control plane so that you can attempt the adoption again later:
 +
 ----
-oc delete --ignore-not-found=true --wait=false openstackcontrolplane/openstack
-oc patch openstackcontrolplane openstack --type=merge --patch '
+$ oc delete --ignore-not-found=true --wait=false openstackcontrolplane/openstack
+$ oc patch openstackcontrolplane openstack --type=merge --patch '
 metadata:
   finalizers: []
 ' || true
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -138,26 +130,23 @@ for i in {1..3}; do
...
+$ sudo pcs constraint order start ceph-nfs then openstack-manila-share kind=Optional id=order-ceph-nfs-openstack-manila-share-Optional
+$ sudo pcs constraint colocation add openstack-manila-share with ceph-nfs score=INFINITY id=colocation-openstack-manila-share-ceph-nfs-INFINITY
 
 ----
 
-. Verify that the source cloud is operational again, e.g. by
-running `openstack` CLI commands or using the {dashboard_first_ref}.
-//kgilliga: Should we be more specific? Which CLI commands?
+. Verify that the source cloud is operational again, for example, you
+can run `openstack` CLI commands or check that you can access the {dashboard_first_ref}.
```
klgill, line 143: 1. Are we saying that they can run any openstack command, and if it’s successful, the cloud is running again? Similarly, if they can access the Dashboard, the cloud is operational?
2. What would the customer need to do if the source cloud is not operational?

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -169,20 +158,12 @@ while oc get pod | grep openstack-galera-0; do
...
-important to verify that the control plane resources have
-been removed and there are no leftovers which could affect the
-following adoption procedure attempt. Notably, the previously created
-copies of the database contents must not be used in another adoption
-attempt, and new copies must be made according to the adoption
-procedure documentation.
-
-//kgilliga: How do users verify that the control plane resources have all been removed?
+After you restore the {OpenStackShort} control plane services, their internal
+state might have changed. Before you retry the adoption procedure, verify that all the control plane resources are removed and that there are no leftovers which could affect the following adoption procedure attempt. You must not use previously created copies of the database contents in another adoption attempt. You must make new copies according to the adoption procedure documentation.
```
klgill, line 169: Instead of “adoption procedure documentation” can we link to something specific, such as “Migrating databases to MariaDB instances”? Or if we need to be more generic, I suggest something like: “You must make new copies according to the procedure for the service you are adopting.”

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
jistr: 1. It doesn't guarantee that all parts of the cloud are working ok, but it's a good smoke test. Really validating full functionality is only possible by something like running the Tempest test suite against the cloud, and we probably shouldn't be recommending that. So it's not a comprehensive validation but i don't think it's worth to aiming for one.
2. Start troubleshooting. Potentially it may lead to "restore from backup" situation. But i think it's very likely that the rollback would succeed, or require just minor tweaks after troubleshooting, without needing to restore from backup.

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
jistr: +1 to linking to a specific section, but i'd link to the higher level "Migrating databases to the control plane" because it's not only about MariaDB but also about OVN data.

jistr (approved):
Responses inline, we can link to the specific section for migrating databases but all LGTM, thanks!

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -2,27 +2,19 @@
 
 = Rolling back the control plane adoption
 
-If you encountered a problem during the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services that prevents you from completing the adoption procedure, you can roll back the control plane adoption.
+If you encountered a problem that prevents you from completing the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services, you can roll back the control plane adoption.
```
mickogeary, line 5: Maybe: "If you encountered a problem with the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services, you can roll back the control plane adoption."

or

"If you are unable to complete the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services, you can roll back the control plane adoption."

or

"If you are unable to complete the adoption of the {rhos_prev_long} ({OpenStackShort}) control plane services, you can roll back the control plane adoption, so that you can troubleshoot and try again."

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -2,27 +2,19 @@
...
-nodes in any way during the procedure, the roll back is not possible.
-
-During the control plane adoption, services on the source cloud's
-control plane are stopped but not removed. The databases on the source
-control plane are not edited by the adoption procedure. The
-destination control plane received a copy of the original
-control plane databases. The roll back procedure assumes that the data
-plane has not yet been touched by the adoption procedure and it is
-still connected to the source control plane.
+You can only roll back the control plane adoption during the parts of the adoption procedure that alter the control plane. If you altered the data plane
```
mickogeary, line 8: Maybe:
"If you altered the data plane nodes in any way, you cannot perform the rollback procedure. You can only roll back the control plane adoption when the control plane is altered. "

(I might put the kind of warning first as that seems more important to me)

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -2,27 +2,19 @@
...
+nodes in any way, you cannot perform the rollback procedure.
+
+During the control plane adoption, services on the {OpenStackShort} control plane are stopped but not removed. The databases on the {OpenStackShort} control plane are not edited during the adoption procedure. The {rhos_long} control plane receives a copy of the original control plane databases. The rollback procedure assumes that the data plane has not yet been modified by the adoption procedure, and it is still connected to the {OpenStackShort} control plane.
```
mickogeary, line 11: I guess this line is explaining the previous line. Does it mean you can't rollback (won't be able to do it) if data plane is altered, or you shouldn't rollback (if you do it, you'll break stuff)?

If it's the latter, maybe the previous line could be something like: "Do not attempt to rollback if you altered the data plane nodes in any way." Could double check with SME about this or what the potential issues are - data corruption maybe?

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -2,27 +2,19 @@
...
 
 The rollback procedure consists of the following steps:
 
-* Restoring the functionality of the source control plane.
+* Restoring the functionality of the {OpenStackShort} control plane.
 
-* Removing the partially or fully deployed destination
-  control plane.
+* Removing the partially or fully deployed {rhos_acro} control plane.
```
mickogeary, line 17: add a hyphen:
"Removing the partially or fully-deployed {rhos_acro} control plane."

Could you also just say "Removing the control plane." ?

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -138,26 +130,23 @@ for i in {1..3}; do
...
-sudo pcs constraint colocation add openstack-manila-share with ceph-nfs score=INFINITY id=colocation-openstack-manila-share-ceph-nfs-INFINITY
+$ sudo pcs constraint order start ceph-nfs then openstack-manila-share kind=Optional id=order-ceph-nfs-openstack-manila-share-Optional
+$ sudo pcs constraint colocation add openstack-manila-share with ceph-nfs score=INFINITY id=colocation-openstack-manila-share-ceph-nfs-INFINITY
 
 ----
 
-. Verify that the source cloud is operational again, e.g. by
-running `openstack` CLI commands or using the {dashboard_first_ref}.
-//kgilliga: Should we be more specific? Which CLI commands?
+. Verify that the source cloud is operational again, for example, you
```
mickogeary, line 142: Maybe you could show an example command, i.e?

. Verify that the source cloud is operational again:

$ example command

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -138,26 +130,23 @@ for i in {1..3}; do
...
+can run `openstack` CLI commands or check that you can access the {dashboard_first_ref}.
 
-. Remove the partially or fully deployed control plane so that another adoption attempt can be made later:
+. Remove the partially or fully deployed control plane so that you can attempt the adoption again later:
```
mickogeary, line 145: or "Remove the control plane in which the adoption failed to complete", which ever you prefer

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
```diff
@@ -169,20 +158,12 @@ while oc get pod | grep openstack-galera-0; do
...
-important to verify that the control plane resources have
-been removed and there are no leftovers which could affect the
-following adoption procedure attempt. Notably, the previously created
-copies of the database contents must not be used in another adoption
-attempt, and new copies must be made according to the adoption
-procedure documentation.
-
-//kgilliga: How do users verify that the control plane resources have all been removed?
+After you restore the {OpenStackShort} control plane services, their internal
+state might have changed. Before you retry the adoption procedure, verify that all the control plane resources are removed and that there are no leftovers which could affect the following adoption procedure attempt. You must not use previously created copies of the database contents in another adoption attempt. You must make new copies. For more information about making new copies, see xref:migrating-databases-to-the-control-plane_storage-requirements[Migrating databases to the control plane].
```
mickogeary, line 169: what is a "new copy"? Like a copy of a copy? Or a new DB instance? Maybe end user will know

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
klgill: I don't think it's possible to roll back if data plane nodes are altered. It's that they can't, not that they shouldn't. But @jistr can you confirm?

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
klgill: New copies of the database contents. So yes, I assume a new copy of the DB instance. @jistr Can you confirm?

jistr: Not a copy of a copy. A copy of the original source DB taken more recently, because the content might have changed. (After we restore the original services on the source control plane, the DB there starts to get modified again. When re-trying adoption, we need to copy the latest state of the source DB contents.)

docs_user/modules/proc_rolling-back-the-control-plane-adoption.adoc
jistr: Technically they could attempt the rollback but likely it would break stuff. So +1 on "Do not attempt to rollback if you altered the data plane nodes in any way."

PR #590, opened 2024-09-07
Fix broken URL in the network config section

elvgarrui:
The "Configuring the network for the RHOSO deployment" section had a broken link. It now points to the proper direction.


jistr (approved):

elvgarrui:
Done, thanks for the review! Should I leave it like this and assume it will work upstream once the changes are ready?

jistr:
I think we can merge and add the attributes in case the link happens to be broken upstream. IIUC the link was broken everywhere before, so we can't be worse off :).

PR #592, opened 2024-09-09
changes required for NFV adoption

eshulman2:
- Re-factored the way the openstack command is foremed for dev env
- Moved static code to vars
- Changed hosts for minimal job to use localhost instead of local which will allow avoiding using an inventory with just localhost in it


tests/vars.sample.yaml
```diff
@@ -1,6 +1,11 @@
 # Path to install_yamls repo
 install_yamls_path: ~/install_yamls #CUSTOMIZE_THIS
 
+src_env_bastion_key: ~/.ssh/id_src_cloud #CUSTOMIZE_THIS
```
jistr, line 4: Should we have these two src_env_bastion variables provided as defaults, or is it ok leaving them out from here? Maybe going forward we should prefer overriding the `openstack_ssh_command` (or `source_openstack_command`) as a whole, without going through the intermediate variables... It kinda evolved organically and we need them to not break existing jobs, but i don't know if it's the best pattern to keep expanding on...

tests/vars.sample.yaml
```diff
@@ -1,6 +1,11 @@
...
+src_env_bastion: "{{ standalone_ip }}" #CUSTOMIZE_THIS
+# ssh command for source env bastion
+openstack_ssh_comand: "ssh -i {{ src_env_bastion_key | default(edpm_privatekey_path) }} -o StrictHostKeyChecking=no {{ source_osp_ssh_user | default('stack') }}@{{ src_env_bastion | default(edpm_node_ip) }} OS_CLOUD={{ os_cloud_name | default('overcloud') }} openstack"
```
frenzyfriday, line 7: I think this should be "openstack_ssh_command"

tests/roles/development_environment/tasks/main.yaml
```diff
@@ -1,46 +1,26 @@
...
-    INSTALL_YAMLS_PATH: "{{ install_yamls_path }}"
-    EDPM_PRIVATEKEY_PATH: "{{ edpm_privatekey_path }}"
-    OS_CLOUD_IP: "{{ standalone_ip | default(edpm_node_ip) }}"
-    OS_CLOUD_NAME: "{{ os_cloud_name }}"
-    SOURCE_OSP_SSH_USER: "{{ source_osp_ssh_user }}"
   ansible.builtin.shell:
     cmd: |
       {{ shell_header }}
-      alias openstack="ssh -i {{ edpm_privatekey_path }} -o StrictHostKeyChecking=no ${SOURCE_OSP_SSH_USER}@${OS_CLOUD_IP} OS_CLOUD=${OS_CLOUD_NAME} openstack"
+      alias openstack={{ openstack_ssh_commmand }}
```
frenzyfriday, line 7: I think this should be "openstack_ssh_command"

tests/roles/development_environment/tasks/main.yaml
```diff
@@ -1,46 +1,26 @@
...
-    EDPM_PRIVATEKEY_PATH: "{{ edpm_privatekey_path }}"
-    OS_CLOUD_IP: "{{ standalone_ip | default(edpm_node_ip) }}"
-    OS_CLOUD_NAME: "{{ os_cloud_name }}"
-    SOURCE_OSP_SSH_USER: "{{ source_osp_ssh_user }}"
   ansible.builtin.shell:
     cmd: |
       {{ shell_header }}
-      alias openstack="ssh -i {{ edpm_privatekey_path }} -o StrictHostKeyChecking=no ${SOURCE_OSP_SSH_USER}@${OS_CLOUD_IP} OS_CLOUD=${OS_CLOUD_NAME} openstack"
-      ${BASH_ALIASES[openstack]} secret store --name testSecret --payload 'TestPayload'
+      {{ openstack_ssh_comand }} secret store --name testSecret --payload 'TestPayload'
```
frenzyfriday, line 16: I think this should be "openstack_ssh_command"

tests/roles/development_environment/tasks/main.yaml
```diff
@@ -1,46 +1,26 @@
...
-    EDPM_PRIVATEKEY_PATH: "{{ edpm_privatekey_path }}"
-    OS_CLOUD_IP: "{{ standalone_ip | default(edpm_node_ip) }}"
-    OS_CLOUD_NAME: "{{ os_cloud_name }}"
-    SOURCE_OSP_SSH_USER: "{{ source_osp_ssh_user }}"
   ansible.builtin.shell:
     cmd: |
       {{ shell_header }}
-      alias openstack="ssh -i {{ edpm_privatekey_path }} -o StrictHostKeyChecking=no ${SOURCE_OSP_SSH_USER}@${OS_CLOUD_IP} OS_CLOUD=${OS_CLOUD_NAME} openstack"
-      ${BASH_ALIASES[openstack]} token issue -f value -c id
+      {{ openstack_ssh_comand }} token issue -f value -c id
```
frenzyfriday, line 23: also here

eshulman2:
recheck

eshulman2:
recheck

jistr:
Looks like an issue coming from the refactoring rather than CI instability:

```
"+ secret store --name testSecret --payload TestPayload", "/bin/bash: line 4: secret: command not found"
```

tests/roles/backend_services/tasks/main.yaml
```diff
@@ -91,11 +95,9 @@
...
     {{ shell_header }}
     {{ oc_header }}
-    mkdir -p tmp
-    oc kustomize base > tmp/test_deployment.yaml
-    oc apply -f tmp/test_deployment.yaml
-  args:
-    chdir: "../config"
+    oc apply -f ../config/tmp/test_deployment.yaml
+  # args:
+  #   chdir: "../config"
```
marios, line 100: had to stare at this for a bit. removing the kustomization tls_overrides.j2 and merging it into the openstack_control_plane. 
i just question if it is really necessary to do it in this patch where you have lots of other nfv specific/required things going on.
It is the kind of change that might benefit of its own review.

for example, does it even make sense to keep a config/base directory any more then? this file is now no longer used https://github.com/marios/data-plane-adoption/blob/main/tests/config/base/openstack_control_plane.yaml lets remove that one too (i.e. you're now always generating it from the j2)

marios (commented):
fyi the last run on the jobs here is hitting a known issue https://issues.redhat.com/browse/OSPCIX-504 

(eg see https://logserver.rdoproject.org/92/592/20e171faa4189ed80e4ee9bc44e080eaeba567e4/github-check/adoption-standalone-to-crc-ceph/51ac3f1/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/openstackdataplanedeployments.dataplane.openstack.org/openstack.yaml)
```
    - lastTransitionTime: "2024-09-25T20:41:52Z"
      message: Deployment error occurred in ovn service error ConfigMap "ovncontroller-config"
        not found
```

jistr:
recheck

eshulman2:
recheck

jistr:
I think this failed on https://issues.redhat.com/browse/OSPCIX-503 which should now be worked around, will try another recheck.

jistr:
recheck

jistr (approved):
LGTM i'm just wondering if we should have a testproject for one of the multinode jobs before we merge.

tests/roles/backend_services/tasks/main.yaml
jistr: FWIW my instinct is the same re making this a separate PR, but i'm also thinking about timeframes and letting other patches rebase after these changes are made, so if this ends up working ok in CI or just needing minor adjustments then perhaps our best path could be to merge the "big things" fast and iterate on any leftovers / polishing?

cescgina: I agree, ideally it would be best to have this change into 2 or 3, but realistically, we should merge this when we get green runs for multinode jobs. I have a small PR that will need to be rebased on top of this to solve conflicts, I can include the cleanup of these unused files there

cescgina (approved):

PR #595, opened 2024-09-10
Document needed /64 prefix per OCP host

elvgarrui:
RFC 4291 https://datatracker.ietf.org/doc/html/rfc4291 section 2.5.1 forces all interface IDs in unicast IPv6 to be 64 bits long. This is now reflected on the documentation to make all the adopters aware of the limitations, since OVN won't be able to work with different interface ID lengths.

Jira: OSPRH-3329

### DIFF
--- a/docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
+++ b/docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
@@ -72,3 +72,18 @@ items:
       kubernetes.io/hostname: ocp-worker-0
       node-role.kubernetes.io/worker: ""
 ```
+
+// TODO: Move this to the IPv6 section once it is fully documented, both upstream and downstream.
+ifeval::["{build}" != "downstream"]
+[WARNING]
+In IPv6, {rhocp_long} worker nodes need a `/64` prefix allocation due to OVN
+limitations (RFC 4291). For dynamic IPv6 configuration, you need to change the
+prefix allocation on the Router Advertisement settings. If you want to use
+manual configuration for IPv6, define a similar CR to the
+`NodeNetworkConfigurationPolicy` CR example in this procedure, and define an
+IPv6 address and disable IPv4. Because the constraint for the `/64` prefix did
+not exist in {OpenstackPreviousInstaller}, your {OpenStackShort}
+control plane network might not have enough capacity to allocate these
+networks. If that is the case, allocate a prefix that fits a large enough number
+of addresses, for example, `/60`. The prefix depends on the number of worker nodes you have.
+endif::[]


jistr (approved):
Makes sense to me, thanks!

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -71,3 +71,16 @@ items:
...
+In IPv6, {OpenShift} worker nodes need a /64 prefix allocation due to OVN
+limitations (RFC 4291). For dynamic IPv6 configuration this needs to be changed
+on the Router Advertisement settings. In case you want to use manual
+configuration for IPv6, define a similar CR to the one seen above that enables
+and defines an IPv6 address and disables IPv4.  Considering that these
+{Openshift} worker nodes are serving as what we called Controllers in
+{OpenstackPreviousInstaller}, where this constraint for the /64 prefix didn't
+exist, there is a chance that your previous {OpenStackShort} Control Plane
+network does not have a size enough to allocate these networks.  If that is the
+case, allocate a prefix wide enough to fit them.
```
klgill, line 86:
```suggestion
[WARNING]
In IPv6, {OpenShift} worker nodes need a `/64` prefix allocation due to OVN limitations (RFC 4291). For dynamic IPv6 configuration, you need to change the prefix allocation on the Router Advertisement settings. If you want to use manual configuration for IPv6, define a similar CR to the `NodeNetworkConfigurationPolicy` CR example in this procedure, and define an IPv6 address and disable IPv4.  Because the constraint for the `/64` prefix did not exist in {OpenstackPreviousInstaller}, your previous {OpenStackShort} control plane network might not have enough capacity to allocate these networks. If that is the case, allocate a prefix large enough to fit them.
```

klgill: @elvgarrui Can you please ensure that I did not change the intended meaning of this note?
Also, would it make sense to add an example value in the last sentence? E.g., "If that is the case, allocate a prefix large enough to fit them, for example, 110."

klgill (changes requested):
I made a few edits.

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
elvgarrui: I updated it now, let me know what you think.

booxter:
The doc update makes sense; there's one issue with the fact that IPv6 as a whole is NOT documented yet (its documentation is being tracked in a different ticket.) This may mean that this warning will have to move to adjust to the broader IPv6 docs update. This is probably ok to merge with this caveat.

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -71,3 +71,16 @@ items:
...
+In IPv6, {OpenShift} worker nodes need a `/64` prefix allocation due to OVN
+limitations (RFC 4291). For dynamic IPv6 configuration, you need to change the
+prefix allocation on the Router Advertisement settings. If you want to use
+manual configuration for IPv6, define a similar CR to the
+`NodeNetworkConfigurationPolicy` CR example in this procedure, and define an
+IPv6 address and disable IPv4.  Because the constraint for the `/64` prefix did
+not exist in {OpenstackPreviousInstaller}, your previous {OpenStackShort}
+control plane network might not have enough capacity to allocate these
+networks. If that is the case, allocate a prefix that fits large enough number
+of addresses. Depends on the number of worker nodes you have, /60 might be enough.
```
klgill, line 86:
```suggestion
IPv6 address and disable IPv4. Because the constraint for the `/64` prefix did
not exist in {OpenstackPreviousInstaller}, your previous {OpenStackShort}
control plane network might not have enough capacity to allocate these
networks. If that is the case, allocate a prefix that fits a large enough number
of addresses, for example, `/60`. The prefix depends on the number of worker nodes you have.
```

klgill: Just some minor edits. Otherwise, looks good!

jistr:
> The doc update makes sense; there's one issue with the fact that IPv6 as a whole is NOT documented yet (its documentation is being tracked in a different ticket.) This may mean that this warning will have to move to adjust to the broader IPv6 docs update. This is probably ok to merge with this caveat.

In case we want to hide this in downstream for now, we can add an ifeval guard like this:

https://github.com/openstack-k8s-operators/data-plane-adoption/blame/23fb1153c70a52796852f3a553cc429258f3be44/docs_user/modules/proc_creating-a-ceph-nfs-cluster.adoc#L14

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
elvgarrui: Thanks, done! I also added the conditional suggested by Jiri.

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -72,3 +72,17 @@ items:
       kubernetes.io/hostname: ocp-worker-0
       node-role.kubernetes.io/worker: ""
 ```
+
+ifeval::["{build}" != "downstream"]
```
booxter, line 76: I dunno how docs team ultimately works, but I wonder if some TODO flag should be added here so that we don't forget to remove the `ifeval` from here when the rest of IPv6 docs are ready.

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -72,3 +72,17 @@ items:
...
+[WARNING]
+In IPv6, {OpenShift} worker nodes need a `/64` prefix allocation due to OVN
```
klgill, line 78:
```suggestion
In IPv6, {rhocp_long} worker nodes need a `/64` prefix allocation due to OVN
```

@elvgarrui I recently changed the "OpenShift" variable to match downstream. See above.

slawqo:
LGTM for me when @klgill's and @booxter's comments will be addressed

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
elvgarrui: Done

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
elvgarrui: Done!

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -72,3 +72,18 @@ items:
       kubernetes.io/hostname: ocp-worker-0
       node-role.kubernetes.io/worker: ""
 ```
+
+# TODO: Move this to the IPv6 section once it is fully documented, both upstream and downstream.
```
jistr, line 76: I think asciidoc uses `//` for comments rather than `#`.

elvgarrui: You are totally right, sorry for that! I updated it

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -72,3 +72,18 @@ items:
       kubernetes.io/hostname: ocp-worker-0
       node-role.kubernetes.io/worker: ""
 ```
+
+// TODO: Move this to the IPv6 section once it is fully documented, both upstream and downstream.
+ifeval::["{build}" != "downstream"]
+[WARNING]
+In IPv6, {rhocp_long} worker nodes need a `/64` prefix allocation due to OVN
```
booxter, line 79: (Minor point, not sure we would want to capture it) I think it's OVN-IC / OVN-Kubernetes limitation. (Which stems from the fact that SLAAC prefixes are used to allocate addresses.)

booxter (approved):

jistr (approved):

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
```diff
@@ -72,3 +72,18 @@ items:
...
+limitations (RFC 4291). For dynamic IPv6 configuration, you need to change the
+prefix allocation on the Router Advertisement settings. If you want to use
+manual configuration for IPv6, define a similar CR to the
+`NodeNetworkConfigurationPolicy` CR example in this procedure, and define an
+IPv6 address and disable IPv4.  Because the constraint for the `/64` prefix did
+not exist in {OpenstackPreviousInstaller}, your previous {OpenStackShort}
+control plane network might not have enough capacity to allocate these
```
klgill, line 86:
```suggestion
`NodeNetworkConfigurationPolicy` CR example in this procedure, and define an
IPv6 address and disable IPv4. Because the constraint for the `/64` prefix did
not exist in {OpenstackPreviousInstaller}, your {OpenStackShort}
control plane network might not have enough capacity to allocate these
```

klgill: Just some additional minor edits.

klgill (changes requested):

docs_user/modules/proc_configuring-openshift-worker-nodes.adoc
elvgarrui: Done, thanks

elvgarrui:
Rebased branch to be up to date with main

PR #597, opened 2024-09-10
Add scenarios for 17.1 env for adoption

cescgina:
Introduce a scenarios folder that will contain the needed input to
deploy a 17.1 environment using the cifmw role added in [1]. The
scenario is defined by a variable file, with undercloud specific
parameters, overcloud specific parameters, hooks that can be called
before or after both the undercloud and overcloud deployment, and two
maps that relate the groups in inventory produced by the playbook that
created the infra, to Roles and roles hostnames, to make it easier to
work with different roles in different scenarios.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2297


scenarios/hci.yaml
```diff
@@ -0,0 +1,64 @@
...
+    source: "path_to_playbook tbd"
+    inventory: "path_to_invenotry_tbd"
+cloud_domain: "localdomain"
+hostname_groups_map:
+  # map ansible groups in the inventory to role hostname format for
+  # 17.1 deployment
+  osp-computes: "overcloud-computehci"
+  osp-controllers: "overcloud-controller"
+overcloud:
+  stackname: "overcloud"
```
bogdando, line 37: this looks great, and allows multi-cell / dcn deployments in the future!

jistr:
:+1:

scenarios/hci.yaml
```diff
@@ -0,0 +1,64 @@
...
+      option: container_cli
+      value: podman
+    - section: DEFAULT
+      option: undercloud_enable_selinux
+      value: false
+    - section: DEFAULT
+      option: cleanup
+      value: false
+  undercloud_parameters_override: "hci/hieradata_overrides_undercloud.yaml"
+  undercloud_parameters_defaults: "hci/undercloud-parameter-defaults.yaml"
```
marios, line 23: since we're starting from scratch here maybe we can use either - or _ in the env files. like pick one and be consistent.

scenarios/hci.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - "/home/zuul/deployed_ceph.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/docker-ha.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/podman.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/debug.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/manila-cephfsnative-config.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/enable-legacy-telemetry.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/cephadm/ceph-mds.yaml"
+  services_file: "hci/overcloud_services.yaml"
```
marios, line 64: not sure about adding a 'services_file' as an interface/variable here
the others, like vip_data network_data, ceph_osh_spec are 'expected' and different from 'normal' tripleo-heat-templates environment files.

but this services_file and also config_download (at least the way we are using it here like an environment file) are just environment files. I am not sure we want to keep them as top level interfaces.

there is no 'services file' notion in tripleo. that's not a reason to not have one here but do we need it, is it different from any other environment file.

cescgina: that's a good point, I haven't noticed that both this file and the config-download one are doing pretty much the same thing and could be merged into one, I've updated the patch merging them

cescgina:
recheck

scenarios/hci/network_data.yaml
```diff
@@ -0,0 +1,61 @@
+---
+- name: Storage
+  mtu: 1500
+  vip: true
+  name_lower: storage
+  dns_domain: storage.mydomain.tld.
```
marios, line 6: for the TLS support (eventually) we'll need to jinjafy this for the domain

cescgina: I was working under this assumption that this files should be entirely static, i.e. there would be a 1:1 mapping of scenario and configuration. But your comment make we wonder, will we have a tlse and a non-tlse version of some scenario, or will we have two different scenarios, one for each case? I would go for the latter, if need to change the setup, we should create a new scenario, even if we incur in some duplication

scenarios/hci.yaml
cescgina: thanks @marios I've changed the files to all use underscores

scenarios/hci/network_data.yaml
fmount: @cescgina the problem I see is that TLS-e is enabled by default in the topology you land (unigamma). You can choose to disable it in the ctlplane/dataplane CRs, or make it configurable.
I don't think we want to duplicate the entire configuration to have something like  `hci-tls` and `hci-no-tls`, but making it a j2 parameter here means we can process it at ci-framework level where we consume those variables and generated pass it as an extravar to the playbook.
It seems a little effort compared to the complexity of maintaining two scenarios for the same HCI topology.

slawqo:
I am definitely not Tripleo expert here but this looks good for me in general.

scenarios/hci/network_data.yaml
cescgina: indeed templating this is simple enough from the ci-framework side, though I'm wary of adding too much complexity here, to avoid ending up with something similar to what we had in install_yamls devsetup, I'll give it a try. The important question here I think is, will we have variations in the scenarios like this, i.e a `hci-tls` and `hci-no-tls`? I worked under the assumption that  here we would target a small subset of well defined configurations

fmount: I don't think we have too many variations: VAs/DTs are supposed to be already an "opinionated" combination of components that reflects what customers use, so I don't expect much difference other than things like tls. We can keep this bit also for a follow up, so we can do good choices when we already have a stable codebase to work on. My comment is more like: let's avoid duplication at this stage, and we can consolidate scenarios in a follow up.

cescgina: ack, thanks for the clarification, as said, the change is simple enough, so I expect I'll be able to updated both PRs with it by the end of day, thanks for the review, it's really useful to clear this points now

scenarios/hci.yaml
```diff
@@ -0,0 +1,66 @@
...
+  vars:
+    - "/home/zuul/deployed_ceph.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/docker-ha.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/podman.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/low-memory-usage.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/debug.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/manila-cephfsnative-config.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/enable-legacy-telemetry.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml"
+    - "/usr/share/openstack-tripleo-heat-templates/environments/cephadm/ceph-mds.yaml"
```
jistr, line 61: I wondered if we should get rid of `args` and `vars` and just put the whole deploy command as a variable to set. I mean, will the deployment command *always* be in the format that is defined in the other patch:

https://github.com/openstack-k8s-operators/ci-framework/pull/2297/files#diff-9e5f7c63638bbac9bfa4df077fe45e16b3acb56feaf89599aacce7a307624498R64

Same for the ceph deploy command, should we just export that out in full as a variable rather than have it in the other patch:

https://github.com/openstack-k8s-operators/ci-framework/pull/2297/files#diff-9a8face42b21d9562691b08533c92de7db4832e6314cb0b9c5479f42dd3d448cR131

In general i tend to be more in favor of exporting larger command chunks as variables perhaps due to being burned by Infrared in the past, where it tried to be very smart and compose the commands for us from various pieces of info, but we almost always ended up fighting it, trying to "trick it" to produce the commands we needed. So its "smartness" quite often stood in the way rather than helped.

I realize there is more complexity to this though, e.g. some of the overcloud deploy input files are generated inside the deployment role rather than provided from the scenario here, so what i'm suggesting above is perhaps easier said then done.

Maybe it would be best to stay the course and just have this as food for thought when designing/amending the `overcloud deploy`, `ceph deploy` and similar commands, that when we hardcode the structure for them, we should be hardcoding something quite un-opinionated.

As a last resort, we can introduce an override mechanism later in the form "If this variable is specified, use it as `overcloud deploy` command. If it is not specified, use the command hardcoded in the deployment role."

jistr (approved):
Thank you Joan for putting this together i think it is very good. Now that i was able to think through the whole thing in combo with the other patch, i put in some very "futursitic concerns" as comments but they are not a -1, just some food for thought. LGTM thanks :).

scenarios/hci.yaml
cescgina: I see what you mean, but as you mentioned dealing with the files generated dynamically will be tricky, I like your last suggestion though, to have an optional argument that'll allow the user to override the whole command, I'll add that and update both patches after testing

scenarios/hci/network_data.yaml.j2
```diff
@@ -0,0 +1,61 @@
+---
+- name: Storage
+  mtu: 1500
+  vip: true
+  name_lower: storage
+  dns_domain: storage.{{ cloud_domain }}.tld.
```
marios, line 6: remove the .tld bit too but leave the trailing '.' 

storage.{{ cloud_domain }}.

cescgina: good catch, fixed, thanks!

scenarios/README.md
```diff
@@ -0,0 +1,121 @@
...
+- `network_data_file`:  path to the network_data file that defines the network
+to use in the overcloud, required.
+- `vips_data_file`:  path to the file defining the virtual ips to use in the
+overcloud, required.
+- `roles_file`: path to the file defining the roles of the different nodes
+used in the overcloud, required.
+- `config_download_file`: path to the config-download file used to pass
+environment variables to the overcloud, required.
+- `ceph_osd_spec_file`: path to the osd spec file used to deploy ceph when
+applicable, optional.
```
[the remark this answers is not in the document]
bogdando: right, you can as well just call them stack or stacks, omitting overcloud terminology.
The env files used for each stack deployment, alongside the pre and post execution hooks, must be instantiated as well. Let them holding only static values for the beginning, no templating yet.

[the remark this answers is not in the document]
cescgina: ack thanks @bogdando I'm preparing an update in this line, I'll update the PR once I've tested it

scenarios/hci/network_data.yaml.j2
```diff
@@ -0,0 +1,61 @@
...
+
+- name: External
+  mtu: 1500
+  vip: true
+  name_lower: external
+  dns_domain: external.{{ cloud_domain }}.
+  service_net_map_replace: external
+  subnets:
+    external_subnet:
+      gateway_ip: '172.21.0.1'
```
marios, line 58: just adding a note about what we discussed on friday; remove this one and add the External network to the controllers (otherwise tls setup fails during deployment)

scenarios/hci/roles.yaml
```diff
@@ -0,0 +1,258 @@
...
+    Controller role that has all the controler services loaded and handles
+    Database, Messaging and Network functions.
+  CountDefault: 1
+  tags:
+    - primary
+    - controller
+    # Create external Neutron bridge for SNAT (and floating IPs when using
+    # ML2/OVS without DVR)
+    - external_bridge
+  networks:
```
marios, line 19: add External here too

scenarios/hci/roles.yaml
```diff
@@ -0,0 +1,258 @@
+---
+###############################################################################
+# File generated by TripleO
```
marios, line 3: probably remove this comment to prevent any confusion? we 'generated' this ourselves via copy paste :)

scenarios/hci/network_data.yaml.j2
cescgina: done, will run a new test and upload the PR

scenarios/hci/roles.yaml
cescgina: done, will update the PR once current testing run finishes

scenarios/hci/roles.yaml
cescgina: done :)

scenarios/README.md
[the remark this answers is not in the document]
cescgina: I changed the `overcloud` section to a `stacks` one, that should support deploying multiple stacks (each has its own  network_data, config-download, etc.)

ciecierski (approved):
looks good to me

jistr:
@cescgina Ok to merge?

cescgina:
> @cescgina Ok to merge?

I'd say yes, I think this is a good first version

PR #608, opened 2024-09-17
Restart neutron pods after control plane adoption

karelyatin:
nova endpoints are missing(as created during nova_adoption) while neutron is adopted, and due to this nova port notifications fails as token is missing nova endpoints and this fails until token is refreshed after default token expiration i.e 3600.

This is a workaround to avoid random tempest failures as reported in related Jira Issue. Once we have some other fix this workaround can be removed.

Related-Issue: OSPCIX-453

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -6,6 +6,14 @@
     ceph_backend_configuration_fsid_shell_vars: |
       CEPH_FSID=$(oc get secret ceph-conf-files -o json | jq -r '.data."ceph.conf"' | base64 -d | grep fsid | sed -e 's/fsid = //')
 
+- name: Restart neutron pods after control plane adoption
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    {{ oc_login_command }}
+    oc delete pod -n openstack -l service=neutron
+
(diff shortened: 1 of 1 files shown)


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -6,6 +6,14 @@
     ceph_backend_configuration_fsid_shell_vars: |
       CEPH_FSID=$(oc get secret ceph-conf-files -o json | jq -r '.data."ceph.conf"' | base64 -d | grep fsid | sed -e 's/fsid = //')
 
+- name: Restart neutron pods after control plane adoption
```
jistr, line 9: As a workaround LGTM but can we create a Jira to track the root cause and link it from here? As it is now i think this would fix CI but not the docs, we could forget about this problem and later we'd get the same bug reported by an end user...

If we want to fix CI quickly we can add the link in a follow up PR and merge this one.

jistr (approved):

tests/roles/dataplane_adoption/tasks/main.yaml
karelyatin: @jistr root cause is basically how we adapt and create endpoints, reported https://issues.redhat.com/browse/OSPRH-10230 explaining this and to track fix on neutron side. Will clear the workaround after fixing OSPRH-10230. So this can be merged to clear adoption jobs for now.

fao89 (approved):

PR #609, opened 2024-09-17
Verify if the existing test VM instance is running

bogdando:
Add an extra verification of the test VM instance created on the source cloud, before adoption started.

This makes that failure mode verification more explicit, and the test suit to fail earlier.

Without that check, the next check to verify if the test VM instance can be stopped will fail, eventually. That is OK as well, yet obfuscates the real cause of an important failure mode - a test instance must survive adoption without downtime.

Related JIRA: [OSPRH-10215](https://issues.redhat.com/browse/OSPRH-10215)

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/55044

### DIFF
--- a/docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
+++ b/docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
@@ -120,6 +120,12 @@ $ oc exec -it nova-cell1-conductor-0 -- nova-manage db online_data_migrations
 $ oc rsh nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
 ----
 
+. Verify if the existing test VM instance is running:
++
+---
+${BASH_ALIASES[openstack]} server --os-compute-api-version 2.48 show --diagnostics test 2>&1 || echo FAIL
+---
+
 . Verify if the Compute services can stop the existing test VM instance:
 +
 ----
--- a/tests/roles/dataplane_adoption/tasks/nova_verify.yaml
+++ b/tests/roles/dataplane_adoption/tasks/nova_verify.yaml
@@ -5,7 +5,16 @@
       alias openstack="oc exec -t openstackclient -- openstack"
 
 # NOTE(bogdando): do not use 'set -o pipefail' for these verifications
-- name: verify if Nova services can stop the existing test VM instance
+- name: verify if the existing test VM instance is running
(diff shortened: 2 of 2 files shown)


docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
```diff
@@ -120,6 +120,12 @@ $ oc exec -it nova-cell1-conductor-0 -- nova-manage db online_data_migrations
 $ oc rsh nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
 ----
 
+. Verify if the existing test VM instance is running:
++
+---
+${BASH_ALIASES[openstack]} server --os-compute-api-version 2.48 show --diagnostics test 2>&1 || echo FAIL
+---
+
```
jistr, line 128: We may want to add upstream-only conditionals here (in another PR perhaps because this is about more than just this edit). AFAIK the downstream docs won't have the concept of a test instance.

jistr (approved):

bogdando:
The failure looks a real issue discovered by this change - the VM is in shut down after adoption

```
TASK [dataplane_adoption : set Nova services shell vars] ***********************
ok: [localhost] => {"ansible_facts": {"nova_header": "alias openstack=\"oc exec -t openstackclient -- openstack\"
"}, "changed": false}

TASK [dataplane_adoption : verify if the existing test VM instance is running] ***
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "alias openstack=\"oc exec -t openstackclient -- openstack\"

${BASH_ALIASES[openstack]} server --os-compute-api-version 2.48 show --diagnostics test 2>&1 || echo FAIL
",
"delta": "0:00:07.095347",
"end": "2024-10-09 12:43:58.418383",
"failed_when_result": true, "rc": 0, "start": "2024-10-09 12:43:51.323036",
"stdout": "No ServerDiagnostics found for None: Client Error for url: https://nova-public-openstack.apps-crc.testing/v2.1/servers/66db1b38-bb01-471c-8190-6f08f1f3a288/diagnostics, Cannot 'get_diagnostics' instance 66db1b38-bb01-471c-8190-6f08f1f3a288 while it is in power state shutdown
command terminated with exit code 1
FAIL",
```

cescgina:
recheck

bogdando:
Dependency is merged, please merge

PR #610, opened 2024-09-17
Test edpm_default_mounts empty

fyanac:


bogdando:
yes, this should be the right choice

PR #616, opened 2024-09-23
swift_adoption: Add check for swift-storage

rajathere:
When deploying the swift CR, we launch pods for swiftProxy, swiftStorage and swiftRing.
The swiftStorage pod runs multiple containers namely account-server, account-replicator, account-auditor etc.
We should wait for the swift storage pod to be active as well to ensure swift is running properly.
The reason for not including swift-ring is that it runs and terminates with Completed state.

swift Ring
swift-ring-rebalance-lpg5w  0/1  Completed  0  62d

swift Storage
oc get pod -l component=swift-storage -o jsonpath='{.items[0].status.phase}{"\n"}' | grep Running Running


rajathere:
swift-storage has 0 replicas so this is not needed in the current test runs.
but surely will be needed in other jobs deploying swift-storage pod.

PR #617, opened 2024-09-23
[DNM] Enable nicmapping ovn controller

karelyatin:


karelyatin:
was a test but nicmapping can be enabled already in tests as net-attach-defs already created

PR #620, opened 2024-09-24
Add Manila CephNFS support

fmount:
This patch introduces a set of basic tasks to perform the adoption of `Manila` with a `CephNFS` backend.
Before performing the actual adoption, a new `CephNFS` `cephadm` based cluster must be created.
For this reason, the `manila_nfs` tasks require a few parameters as input:

1. A new `CephNFS` `VIP` where the `CephIngress` daemon (made by `Haproxy` and `keepalived`) is created;
2. A set of Ceph `target_nodes` where the "nfs" label is added: these nodes are supposed to host the new `CephNFS` cluster and must be different from controller nodes that are going to be decommisioned;
~3. The `TripleO` managed `Ganesha` `VIP`: this input can be actually retrieved by the existing `TripleO` deployment, but we can gather this input as part of the next iteration on this patch~

After the new `CephNFS` cluster is created, it is possible to build a `manila-share` with the expected NFS configuration.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/955

### DIFF
--- a/tests/roles/manila_adoption/tasks/main.yaml
+++ b/tests/roles/manila_adoption/tasks/main.yaml
@@ -1,19 +1,42 @@
+- name: Check the required input when manila_backend is NFS
+  when: manila_backend == "cephnfs"
+  block:
+    - name: Set shell vars to connect to controller1
+      no_log: "{{ use_no_log }}"
+      ansible.builtin.set_fact:
+        controller_ssh: |
+          CONTROLLER1_SSH="{{ controller1_ssh }}"
+
+    - name: Get ceph-nfs IP Address
+      become: true
+      ansible.builtin.shell: |
+        {{ shell_header }}
+        {{ controller_ssh }}
+        ${CONTROLLER1_SSH} awk -F '[=;]' '/Bind_Addr/ {gsub(/ /, "", $2); print $2}' {{ ganesha_default_path }}
+      register: cephnfs_vip
+
+    - name: Fail if the OLD Ganesha VIP is not a good input value
+      when:
+        - not (cephnfs_vip.stdout | ansible.builtin.ipaddr)
+      ansible.builtin.fail:
+        msg: "The (TRIPLEO) gathered Ganesha server IP is malformed"
+
 - name: Deploy Podified Manila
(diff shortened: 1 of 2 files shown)


fmount:
@lkuchlan I might need your help to `testproject` this use case and define the set of parameters we need to pass as overrides when NFS is set as a backend.
@gouthampacha I start doing some sanity check against this code and try to reduce (as much as possible) the number of overrides we need to setup in the job.

tests/roles/manila_adoption/tasks/main.yaml
```diff
@@ -1,19 +1,63 @@
...
+      ansible.builtin.set_fact:
+        controller_ssh: |
+          CONTROLLER1_SSH="{{ controller1_ssh }}"
+
+    - name: Get ceph-nfs IP Address
+      become: true
+      ansible.builtin.shell: |
+        {{ shell_header }}
+        {{ controller_ssh }}
+        ${CONTROLLER1_SSH} sudo pcs constraint config | awk '/^  ip-.*ceph-nfs/{split($1,out,"-"); print out[2]}'
```
fmount, line 26: @gouthampacha I'm not sure we have a better (less error prone) way to gather this IP address. I can experiment a bit but I'm open to suggestions.
An idea is to grep it from `/etc/ganesha/ganesha.conf`: not sure we have other challenges there or is less complicated.

fmount: Looks better now

fmount:
recheck

fmount:
seems good to go now

gouthampacha (approved):
This looks good; thanks @fmount .. i like the strategy of using `ganesha.conf` rather than other sources for this VIP: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/d909b766375ff7b22fa3fe4828869e21034e517c/tests/roles/manila_adoption/templates/manila_cephfs.yaml.j2#L51

lkuchlan (approved):
Thanks Francesco, LGTM!

jistr (approved):

PR #624, opened 2024-09-25
Verify cinder resources after adoption

rajathere:
We should verify that the cinder volumes, snapshot and backup resources have consistent state before and after the adoption.


jistr:
`cinder_verify.yaml` looks good but i think you forgot to include it from `main.yaml`

jistr (approved):

cescgina (approved):
lgtm

PR #629, opened 2024-09-27
[WIP] Run network config in the end

karelyatin:
Running network config is re creating ovs bridges, leading to longer network downtime as ovs flows
misses until ovn is restarted.

Also disable gateway on computes, it got added by
mistake when adding networker support.

Related-Issue: https://issues.redhat.com/browse/OSPRH-10283

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -166,8 +166,6 @@
       services:
         - bootstrap
         - download-cache
-        - configure-network
-        - validate-network
         - install-os
         - configure-os
         - ssh-known-hosts
@@ -176,6 +174,8 @@
         - install-certs
         - ovn
         - neutron-metadata
+        - configure-network
+        - validate-network
       env:
         - name: ANSIBLE_CALLBACKS_ENABLED
           value: "profile_tasks"
--- a/tests/roles/dataplane_adoption/defaults/main.yaml
+++ b/tests/roles/dataplane_adoption/defaults/main.yaml
@@ -167,8 +167,6 @@ dataplane_cr: |
     services:
       - bootstrap
       - download-cache
-      - configure-network
-      - validate-network
       - install-os
       - configure-os
       - ssh-known-hosts
@@ -177,6 +175,8 @@ dataplane_cr: |
       - install-certs
       - ovn
       - neutron-metadata
+      - configure-network
+      - validate-network
       {%+ if compute_adoption|bool +%}
       - libvirt
       - nova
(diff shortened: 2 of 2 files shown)


jistr (commented):
+1 on the edpm_enable_chassis_gw fix.

On the service ordering i'd have two comments:

* If we want to do this, we should make matching edits in the docs. (Searching for 'configure-network' in https://openstack-k8s-operators.github.io/data-plane-adoption/user/ shows what we're documenting.)

* Another thing to consider is that this would mean we have different service ordering for adopted vs. greenfield environments. Not an absolute blocker perhaps, but it would be preferable if we could find a solution that keeps service ordering the same between greenfield and adoption.

karelyatin:
<< 

    If we want to do this, we should make matching edits in the docs. (Searching for 'configure-network' in https://openstack-k8s-operators.github.io/data-plane-adoption/user/ shows what we're documenting.)

Ok will update, for now just keeping it open for testing.

    Another thing to consider is that this would mean we have different service ordering for adopted vs. greenfield environments. Not an absolute blocker perhaps, but it would be preferable if we could find a solution that keeps service ordering the same between greenfield and adoption.

Yes we would need fix on os-net-config side, more details in https://issues.redhat.com/browse/OSPRH-10283. This could be considered as a workaround until then. I think we can have different order for adoption then green field, we already have requirements to update ovn controllers before controller, etc. But yes if possible any reorder should be avoided.

karelyatin:
recheck ping workaround landed

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -167,6 +166,8 @@
         - install-certs
         - ovn
         - neutron-metadata
+        - configure-network
```
bogdando, line 169: just to note, that we might as well change this ordering in VAs, to keep things synced, if that makes sense for greenfield

fyanac:
recheck

karelyatin:
> +1 on the edpm_enable_chassis_gw fix.
> 

Pushed https://github.com/openstack-k8s-operators/data-plane-adoption/pull/648 for remove edpm_enable_chassis_gw

> On the service ordering i'd have two comments:
> 
>     * If we want to do this, we should make matching edits in the docs. (Searching for 'configure-network' in https://openstack-k8s-operators.github.io/data-plane-adoption/user/ shows what we're documenting.)
> 
>     * Another thing to consider is that this would mean we have different service ordering for adopted vs. greenfield environments. Not an absolute blocker perhaps, but it would be preferable if we could find a solution that keeps service ordering the same between greenfield and adoption.

The change is temporary until os-net-config fix is available so going with https://review.rdoproject.org/r/c/rdo-jobs/+/55013 and accordingly update downstream jobs to switch the service run order temporary to get testing unblocked.

So will close this PR

karelyatin:
No longer needed as mentioned in last comment

PR #630, opened 2024-09-27
Avoid hardcoding parameters in pre_launch script

cescgina:
Don't hardcode connection information in the pre_launch script, but use
the variables that are already defined in the calling task.

### DIFF
--- a/tests/roles/development_environment/files/pre_launch.bash
+++ b/tests/roles/development_environment/files/pre_launch.bash
@@ -1,6 +1,6 @@
 set -e
 
-alias openstack="ssh -q -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@${OS_CLOUD_IP} OS_CLOUD=${OS_CLOUD_NAME} openstack"
+alias openstack="ssh -q -i ${EDPM_PRIVATEKEY_PATH} ${SOURCE_OSP_SSH_USER}@${OS_CLOUD_IP} OS_CLOUD=${OS_CLOUD_NAME} openstack"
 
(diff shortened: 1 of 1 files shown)


jistr (approved):
I guess you needed this for the work on the uni job?

This may get refactored further in #592 but it's already an improvement over the present state because it makes it consistent with what's in https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/development_environment/tasks/main.yaml . In fact we could also remove the "alias" line from there because it's being re-defined here, but since i know #592 is already refactoring that same spot and don't want to create more code conflict than necessary, perhaps we can stick to the minimal edit we already have.

cescgina:
> I guess you needed this for the work on the uni job?
> 
> This may get refactored further in #592 but it's already an improvement over the present state because it makes it consistent with what's in https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/development_environment/tasks/main.yaml . In fact we could also remove the "alias" line from there because it's being re-defined here, but since i know #592 is already refactoring that same spot and don't want to create more code conflict than necessary, perhaps we can stick to the minimal edit we already have.

exactly, I've noticed it trying to test the new role to deploy 17.1, it's not really blocking me and it could wait, but since 592 is much bigger, I think it makes sense to merge this smaller change before

fao89 (approved):

cescgina:
recheck

cescgina:
Closing this in favor of https://github.com/openstack-k8s-operators/data-plane-adoption/pull/592

PR #632, opened 2024-09-28
Add many more ping attempts

cescgina:
Add many more ping attempts to the test vm FIP to check if the missing
connectivity error is time-dependent.

Resolves: OSPCIX-503

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
+++ b/tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
@@ -51,4 +51,4 @@
 - name: verify connectivity to the existing test VM instance using Floating IP
   when: prelaunch_test_instance|bool
   ansible.builtin.shell: |
-      ping -c4 {{ lookup('env', 'FIP') | default('192.168.122.20', True) }}
+      ping -c10 {{ lookup('env', 'FIP') | default('192.168.122.20', True) }}


tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
```diff
@@ -51,4 +51,4 @@
 - name: verify connectivity to the existing test VM instance using Floating IP
   when: prelaunch_test_instance|bool
   ansible.builtin.shell: |
-      ping -c4 {{ lookup('env', 'FIP') | default('192.168.122.20', True) }}
+      ping -c10 {{ lookup('env', 'FIP') | default('192.168.122.20', True) }}
```
marios, line 54: pasting from slack

```
can we maybe do a wait_for with an explicit timeout (like 3/4 mins or so?)
10:07
wait_for_connection
10:07
i was looking for ansible module that would help https://docs.ansible.com/ansible/latest/collections/ansible/builtin/wait_for_connection_module.html
10:08
so, as a first pass i think we can merge the proposal
10:08
and that can be a task for improvement
10:08
for future adoption ci team

```

marios (approved):
ok for now but as discussed in slack it might be improved code wise but indeed it would be good to understand more about the root cause to decide if we need another fix somewhere else too.

tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
bogdando: please remember that there is a rule for any shell command in tests to match the same commands in docs. This is why we should avoid ansible smartness in test suits

PR #633, opened 2024-09-30
 Add variable to set the controlplane storage class

cescgina:
Add a new variable control_plane_storage_class that is  used by the
controlplane. This allows us to deploy using an storage class different
than local-storage.
    
Also increase slightly the storage request for the database, since in
some environment it seems to use slightly more than 500 Mb.

### DIFF
--- a/tests/roles/backend_services/templates/openstack_control_plane.j2
+++ b/tests/roles/backend_services/templates/openstack_control_plane.j2
@@ -4,7 +4,7 @@ metadata:
   name: openstack
 spec:
   secret: osp-secret
-  storageClass: local-storage
+  storageClass: {{ control_plane_storage_class  }}
 
   barbican:
     enabled: false
@@ -77,11 +77,11 @@ spec:
       openstack:
         secret: osp-secret
         replicas: 1
-        storageRequest: 500M
+        storageRequest: 1Gi
       openstack-cell1:
         secret: osp-secret
         replicas: 1
-        storageRequest: 500M
+        storageRequest: 1Gi
 
   memcached:
     enabled: true
--- a/tests/roles/backend_services/defaults/main.yaml
+++ b/tests/roles/backend_services/defaults/main.yaml
@@ -17,3 +17,4 @@ heat_auth_encryption_key: ''
 swift_password: ''
 dns_lb_ip: 192.168.122.80
 dns_server_ip: 192.168.122.1
+control_plane_storage_class: local-storage
--- a/tests/roles/backend_services/tasks/main.yaml
+++ b/tests/roles/backend_services/tasks/main.yaml
@@ -96,8 +96,6 @@
     {{ shell_header }}
     {{ oc_header }}
     oc apply -f ../config/tmp/test_deployment.yaml
-  # args:
-  #   chdir: "../config"
 
 - name: wait for services to start up
   ansible.builtin.shell: |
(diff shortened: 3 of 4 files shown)


fao89 (approved):

tests/roles/backend_services/templates/openstack_control_plane.j2
```diff
@@ -4,7 +4,7 @@ metadata:
   name: openstack
 spec:
   secret: osp-secret
-  storageClass: local-storage
+  storageClass: {{ control_plane_storage_class  }}
```
jistr, line 7: I'm wondering if we should have this new variable or if we should be using the existing `storage_class_name` here too. Is there a use case for differing values for these?

cescgina: `storage_class_name` defaults to `crc-csi-hostpath-provisioner` https://github.com/openstack-k8s-operators/data-plane-adoption/blob/07f2c16f363c0dc25e944657695db252f588d7c4/tests/roles/ovn_adoption/defaults/main.yaml#L4 and https://review.rdoproject.org/r/plugins/gitiles/rdo-jobs/+/refs/heads/master/playbooks/data_plane_adoption/vars.yaml#20 and using that storage class did not seem to work for the control plane. Also I found this old PR  where I got the impression that we wanted to use local-storage until  we can use lvms (we'll be able to use them easily in the uni-jobs) https://github.com/openstack-k8s-operators/data-plane-adoption/pull/227

cescgina: but I'm not very knowledgeable about the storage part, so maybe this PR can improved, this change is just the least different to what we current have in CI

jistr: I'm also not very knowledgeable about this but i wonder if we should try using just one storage class variable and perhaps change the default to `local-storage`. That's the value which can be found mentioned in architecture repo, the CI framework etc. The `crc-csi-hostpath-provisioner` approach seems to be specific to the Adoption test suite, i've not seen it anywhere else. IIUC the `local-storage` gets set up by `make crc-storage` when using the test topologies based on CRC. If we'd like to use `crc-csi-hostpath-provisioner` in dev envs, we can customize the variable value.

@bogdando you're the author of #227, does going with `local-storage` set up by `make crc-storage` sound ok to you?

jistr: IIUC the `storage_class_name` variable is currently used as "backend for `local-storage` storage class" while we should probably be using it for the actual storageClass value in the OpenStack CRs instead. And we should be probably defaulting to `local-storage` set up by `make crc-storage` like it's done in the rest of openstack-k8s project.

jistr: This probably needs more reading and a re-review of what was done in #227. I find the current state a bit confusing :)

jistr: Actually now i see #227 got reverted. What we have now is quite inconsistent use, somewhere we hardcode `local-storage`, elsewhere we use the value of `storage_class_name` variable. I'm doodling here with a possible fix: #643

cescgina:
Closing in favor of https://github.com/openstack-k8s-operators/data-plane-adoption/pull/643 which does this better, with a single variable

PR #636, opened 2024-10-01
[DNM] Check console log test vm

karelyatin:


gibizer:
we don't need this any more.

PR #639, opened 2024-10-02
[DNM] Revert ping increase to investigate CIX503

cescgina:


cescgina:
this is not needed anymore, closing

PR #641, opened 2024-10-03
started refining network config chapter

klgill:
I need to finish the IPAM assembly on Friday: https://issues.redhat.com/browse/RHOSPDOC-2014

### DIFF
--- a/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
+++ b/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
@@ -4,38 +4,24 @@ ifdef::context[:parent-context: {context}]
 
 :context: configuring-network
 
-= Configuring the network for the RHOSO deployment
+= Configuring the network for the {rhos_long_noacro} deployment
 
-With {rhocp_long}, the network is a very important aspect of the deployment, and
-it is important to plan it carefully. The general network requirements for the
-{rhos_prev_long} ({OpenStackShort}) services are not much different from the ones in a {OpenStackPreviousInstaller} deployment, but the way you handle them is.
-
-[NOTE]
-For more information about the network architecture and configuration, see
-link:{deploying-rhoso}/assembly_preparing-rhocp-for-rhoso[Preparing Red Hat OpenShift Container Platform for {rhos_long_noacro}] in the _ {deploying-rhoso-t}_guide and link:{defaultOCPURL}/networking/about-networking.html[About networking].
-
-When adopting a new {OpenStackShort} deployment, it is important to align the network
+When you adopt a new {rhos_long} deployment, you must align the network
 configuration with the adopted cluster to maintain connectivity for existing
 workloads.
 
-The following logical configuration steps will incorporate the existing network
-configuration:
+You need to incorporate the existing network configuration in the following steps:
 
-* configure **{OpenShiftShort} worker nodes** to align VLAN tags and IPAM
-  configuration with the existing deployment.
-* configure **Control Plane services** to use compatible IP ranges for
-  service and load balancing IPs.
-* configure **Data Plane nodes** to use corresponding compatible configuration for
-  VLAN tags and IPAM.
+* You configure {rhocp_long} worker nodes to align VLAN tags and IP Address Management (IPAM) configuration with the existing deployment.
+* You configure control plane services to use compatible IP ranges for service and load-balancing IP addresses.
+* You configure data plane nodes to use corresponding compatible configuration for VLAN tags and IPAM.
+* You can either reuse IPAM configuration from the existing deployment or, depending on IP address availability in the existing allocation pools, define new ranges to be used for the new control plane services. If you define new ranges, you configure IP routing between the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
+* You reuse VLAN tags from the existing deployment.
 
-Specifically,
+[NOTE]
+For more information about the network architecture and configuration, see
+link:{deploying-rhoso}/assembly_preparing-rhoso-networks[Preparing networks for Red Hat OpenStack Services on OpenShift] in _{deploying-rhoso-t}_ and link:{defaultOCPURL}/networking/about-networking[About networking] in _Networking_.
 
-* IPAM configuration will either be reused from the
-  **existing** deployment or, depending on IP address availability in the
-  existing allocation pools, **new** ranges will be defined to be used for the
-  new control plane services. If so, **IP routing** will be configured between
-  the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
-* VLAN tags will be reused from the existing deployment.
 
 include::../modules/proc_retrieving-network-information-from-your-existing-deployment.adoc[leveloffset=+1]
 
--- a/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
+++ b/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
@@ -12,7 +12,7 @@ workloads.
 
 Perform the following tasks to incorporate the existing network configuration:
 
-* You configure {rhocp_long} worker nodes to align VLAN tags and IP Address Management (IPAM) configuration with the existing deployment.
+* Configure {rhocp_long} worker nodes to align VLAN tags and IP Address Management (IPAM) configuration with the existing deployment.
 * You configure control plane services to use compatible IP ranges for service and load-balancing IP addresses.
 * You configure data plane nodes to use corresponding compatible configuration for VLAN tags and IPAM.
 * You can either reuse IPAM configuration from the existing deployment or, depending on IP address availability in the existing allocation pools, define new ranges to be used for the new control plane services. If you define new ranges, you configure IP routing between the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
--- a/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
+++ b/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
@@ -10,7 +10,7 @@ When you adopt a new {rhos_long} deployment, you must align the network
 configuration with the adopted cluster to maintain connectivity for existing
 workloads.
 
-You need to incorporate the existing network configuration in the following steps:
+Perform the following tasks to incorporate the existing network configuration:
 
 * You configure {rhocp_long} worker nodes to align VLAN tags and IP Address Management (IPAM) configuration with the existing deployment.
 * You configure control plane services to use compatible IP ranges for service and load-balancing IP addresses.
--- a/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
+++ b/docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
@@ -15,7 +15,7 @@ Perform the following tasks to incorporate the existing network configuration:
 * Configure {rhocp_long} worker nodes to align VLAN tags and IP Address Management (IPAM) configuration with the existing deployment.
 * You configure control plane services to use compatible IP ranges for service and load-balancing IP addresses.
 * You configure data plane nodes to use corresponding compatible configuration for VLAN tags and IPAM.
-* You can either reuse IPAM configuration from the existing deployment or, depending on IP address availability in the existing allocation pools, define new ranges to be used for the new control plane services. If you define new ranges, you configure IP routing between the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
+* Reuse IPAM configuration from the existing deployment or, depending on IP address availability in the existing allocation pools, define new ranges for the new control plane services. If you define new ranges, you configure IP routing between the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
 * You reuse VLAN tags from the existing deployment.
 
 [NOTE]
--- a/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
+++ b/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
@@ -13,7 +13,7 @@ The RHOSO control plane has different requirements for the number of IP addresse
 
 The total number of IP addresses that are required for the new control plane services in each isolated network is calculated as the sum of the following:
 
-* The number of {OpenShiftShort} worker nodes. Each node requires 1 IP address in the `NodeNetworkConfigurationPolicy` custom resources (CRs).
+* The number of {OpenShiftShort} worker nodes. Each worker node requires 1 IP address in the `NodeNetworkConfigurationPolicy` custom resource (CR).
 * The number of IP addresses required for the data plane nodes. Each node requires an IP address from the `NetConfig` CRs.
 * The number of IP addresses required for control plane services. Each service requires an IP address from the `NetworkAttachmentDefinition` CRs. This number depends on the number of replicas for each service.
 * The number of IP addresses required for load balancer IP addresses. Each service requires a Virtual IP address from the `IPAddressPool` CRs.
--- a/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
+++ b/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
@@ -9,7 +9,7 @@ ifdef::context[:parent-context: {context}]
 In a {rhos_long} deployment, each service that is deployed on the {rhocp_long} worker nodes requires an IP address from the IP Address Management (IPAM) pool. In a {rhos_prev_long} ({OpenStackShort}) deployment, all services that are hosted on a Controller node shared the same IP address.
 
 Since the new control plane deployment has different requirements as to the
-number of IP addresses available for services, you cannot reuse the existing IP ranges in the {rhos_acro} environment, depending on the size.
+The RHOSO control plane has different requirements for the number of IP addresses that are made available for services. Therefore, you cannot use the IP ranges from the existing {rhos_acro} environment in the RHOSO control plane.
 
 The total number of IP addresses that are required for the new control plane services in each isolated network is calculated as the sum of the following:
 
--- a/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
+++ b/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
@@ -6,65 +6,39 @@ ifdef::context[:parent-context: {context}]
 
 = Planning your IPAM configuration
 
-The new deployment model puts additional burden on the size of IP allocation
-pools available for {rhos_prev_long} ({OpenStackShort}) services. This is because each service deployed
-on {rhocp_long} worker nodes will now require an IP address from the IPAM pool (in
-the previous deployment model, all services hosted on a controller node shared
-the same IP address.)
+In a {rhos_long} deployment, each service that is deployed on the {rhocp_long} worker nodes requires an IP address from the IPAM pool. In a {rhos_prev_long} ({OpenStackShort}) deployment, all services that are hosted on a Controller node shared the same IP address.
 
 Since the new control plane deployment has different requirements as to the
-number of IP addresses available for services, it may even be impossible to
-reuse the existing IP ranges used in adopted environment, depending on its
-size. Prudent planning is required to determine which options are available in
-your particular case.
+number of IP addresses available for services, you cannot reuse the existing IP ranges in the {rhos_acro} environment, depending on the size.
 
-The total number of IP addresses required for the new control plane services,
-in each isolated network, is calculated as a sum of the following:
+The total number of IP addresses that are required for the new control plane services in each isolated network is calculated as the sum of the following:
 
-* The number of {OpenShiftShort} worker nodes. (Each node will require 1 IP address in
-  `NodeNetworkConfigurationPolicy` custom resources (CRs).)
-* The number of IP addresses required for the data plane nodes. (Each node will require
-  an IP address from `NetConfig` CRs.)
-* The number of IP addresses required for control plane services. (Each service
-  will require an IP address from `NetworkAttachmentDefinition` CRs.) This
-  number depends on the number of replicas for each service.
-* The number of IP addresses required for load balancer IP addresses. (Each
-  service will require a VIP address from `IPAddressPool` CRs.)
+* The number of {OpenShiftShort} worker nodes. Each node requires 1 IP address in the `NodeNetworkConfigurationPolicy` custom resources (CRs).
+* The number of IP addresses required for the data plane nodes. Each node requires an IP address from the `NetConfig` CRs.
+* The number of IP addresses required for control plane services. Each service requires an IP address from the `NetworkAttachmentDefinition` CRs. This number depends on the number of replicas for each service.
+* The number of IP addresses required for load balancer IP addresses. Each service requires a Virtual IP address from the `IPAddressPool` CRs.
 
-As of the time of writing, the simplest single worker node {OpenShiftShort} deployment
-(CRC) has the following IP ranges defined (for the `internalapi` network):
+For example, a simple single worker node {OpenShiftShort} deployment
+with Red Hat OpenShift Local has the following IP ranges defined for the `internalapi` network:
 
-* 1 IP address for the single worker node;
-* 1 IP address for the data plane node;
+* 1 IP address for the single worker node
+* 1 IP address for the data plane node
 * `NetworkAttachmentDefinition` CRs for control plane services:
-  `X.X.X.30-X.X.X.70` (41 addresses);
+  `X.X.X.30-X.X.X.70` (41 addresses)
 * `IPAllocationPool` CRs for load balancer IPs: `X.X.X.80-X.X.X.90` (11
-  addresses).
+  addresses)
 
-Which comes to a total of 54 IP addresses allocated to the `internalapi`
+This example shows a total of 54 IP addresses allocated to the `internalapi`
 allocation pools.
 
 // TODO: update the numbers above for a more realistic multinode cluster.
 
-The exact requirements may differ depending on the list of {OpenStackShort} services
-to be deployed, their replica numbers, as well as the number of {OpenShiftShort}
-worker nodes and data plane nodes.
+The requirements might differ depending on the list of {OpenStackShort} services
+to be deployed, their replica numbers, and the number of {OpenShiftShort} worker nodes and data plane nodes.
 
-Additional IP addresses may be required in future {OpenStackShort} releases, so it is
-advised to plan for some extra capacity, for each of the allocation pools used
-in the new environment.
+Additional IP addresses might be required in future {OpenStackShort} releases, so you must plan for some extra capacity for each of the allocation pools that are used in the new environment.
 
-Once you know the required IP pool size for the new deployment, you can choose
-one of the following scenarios to handle IPAM allocation in the new
-environment.
-
-The first listed scenario is more general and implies using new IP ranges,
-while the second scenario implies reusing the existing ranges. The end state of
-the former scenario is using the new subnet ranges for control plane services,
-but keeping the old ranges, with their node IP address allocations intact, for
-data plane nodes.
-
-Regardless of the IPAM scenario, the VLAN tags used in the existing deployment will be reused in the new deployment. Depending on the scenario, the IP address ranges to be used for control plane services will be either reused from the old deployment or defined anew. Adjust the configuration as described in xref:configuring-isolated-networks_configuring-network[Configuring isolated networks].
+After you determine the required IP pool size for the new deployment, you can choose to define new IP address ranges or reuse your existing IP address ranges.Regardless of the scenario, the VLAN tags in the existing deployment are reused in the new deployment. Adjust the configuration as described in xref:configuring-isolated-networks_configuring-network[Configuring isolated networks].
 
 include::../modules/proc_using-new-subnet-ranges.adoc[leveloffset=+1]
 
--- a/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
+++ b/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
@@ -38,7 +38,7 @@ to be deployed, their replica numbers, and the number of {OpenShiftShort} worker
 
 Additional IP addresses might be required in future {OpenStackShort} releases, so you must plan for some extra capacity for each of the allocation pools that are used in the new environment.
 
-After you determine the required IP pool size for the new deployment, you can choose to define new IP address ranges or reuse your existing IP address ranges.Regardless of the scenario, the VLAN tags in the existing deployment are reused in the new deployment. Adjust the configuration as described in xref:configuring-isolated-networks_configuring-network[Configuring isolated networks].
+After you determine the required IP pool size for the new deployment, you can choose to define new IP address ranges or reuse your existing IP address ranges. Regardless of the scenario, the VLAN tags in the existing deployment are reused in the new deployment. Ensure that the VLAN tags are properly retained in the new configuration. For more information, see xref:configuring-isolated-networks_configuring-network[Configuring isolated networks].
 
 include::../modules/proc_using-new-subnet-ranges.adoc[leveloffset=+1]
 
--- a/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
+++ b/docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
@@ -6,7 +6,7 @@ ifdef::context[:parent-context: {context}]
 
 = Planning your IPAM configuration
 
-In a {rhos_long} deployment, each service that is deployed on the {rhocp_long} worker nodes requires an IP address from the IPAM pool. In a {rhos_prev_long} ({OpenStackShort}) deployment, all services that are hosted on a Controller node shared the same IP address.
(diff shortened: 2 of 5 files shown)


docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
```diff
@@ -6,65 +6,39 @@ ifdef::context[:parent-context: {context}]
...
+The total number of IP addresses that are required for the new control plane services in each isolated network is calculated as the sum of the following:
+
+* The number of {OpenShiftShort} worker nodes. Each node requires 1 IP address in the `NodeNetworkConfigurationPolicy` custom resources (CRs).
+* The number of IP addresses required for the data plane nodes. Each node requires an IP address from the `NetConfig` CRs.
+* The number of IP addresses required for control plane services. Each service requires an IP address from the `NetworkAttachmentDefinition` CRs. This number depends on the number of replicas for each service.
+* The number of IP addresses required for load balancer IP addresses. Each service requires a Virtual IP address from the `IPAddressPool` CRs.
+
+For example, a simple single worker node {OpenShiftShort} deployment
+with Red Hat OpenShift Local has the following IP ranges defined for the `internalapi` network:
+
```
klgill, line 23: Does this sentence still make sense? CRC (CodeReady Containers) is not used anymore, so I changed it to "Red Hat OpenShift Local."

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
```diff
@@ -6,65 +6,39 @@ ifdef::context[:parent-context: {context}]
...
+  `X.X.X.30-X.X.X.70` (41 addresses)
 * `IPAllocationPool` CRs for load balancer IPs: `X.X.X.80-X.X.X.90` (11
-  addresses).
+  addresses)
 
-Which comes to a total of 54 IP addresses allocated to the `internalapi`
+This example shows a total of 54 IP addresses allocated to the `internalapi`
 allocation pools.
 
 // TODO: update the numbers above for a more realistic multinode cluster.
```
klgill, line 34: Has the TODO comment been addressed?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -1,43 +1,23 @@
 [id="using-new-subnet-ranges_{context}"]
 
-= Scenario 1: Using new subnet ranges
+= Configuring new subnet ranges
 
-This scenario is compatible with any existing subnet configuration, and can be
-used even when the existing cluster subnet ranges don't have enough free IP
-addresses for the new control plane services.
+You can define new IP ranges for control plane services that belong to a different subnet that is not used in the existing cluster. Then you configure link local IP routing between the old and new subnets to enable old and new service deployments to communicate. This involves using the {OpenStackPreviousInstaller} mechanism on a pre-adopted cluster to configure additional link local routes. This enables the EDP deployment to reach out to adopted nodes by using the old subnet addresses. You can use new subnet ranges with any existing subnet configuration, and when the existing cluster subnet ranges do not have enough free IP addresses for the new control plane services.
```
klgill, line 5: EDP = EDPM? (We use “data plane” instead of EDPM).

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -1,43 +1,23 @@
...
-The general idea here is to define new IP ranges for control plane services
-that belong to a different subnet that was not used in the existing cluster.
-Then, configure link local IP routing between the old and new subnets to allow
-old and new service deployments to communicate. This involves using {OpenStackPreviousInstaller}
-mechanism on pre-adopted cluster to configure additional link local routes
-there. This will allow EDP deployment to reach out to adopted nodes using their
-old subnet addresses.
+You must size the new subnet appropriately to accommodate the new control
+plane services. There are no specific requirements for the
+existing deployment allocation pools already consumed.
```
klgill, line 9: I’m a little bit confused about what “consumed” means in this context. Already consumed by what? The new RHOSO deployment?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -1,43 +1,23 @@
...
-scenario, as the old subnet ranges are kept for the adopted nodes, which means
-they don't consume any IP addresses from the new range.
+In the following procedure, you configure `NetworkAttachmentDefinition` custom resources (CRs) to use a different subnet from what is configured in the `NetConfig` CR for the same networks. The former range is used for control plane services, while the new range is used to manage IP Address Management (IPAM) for data plane nodes.
 
-In this scenario, you will configure `NetworkAttachmentDefinition` custom resources (CRs) to use a
-different subnet from what will be configured in `NetConfig` CR for the same
-networks. The former range will be used for control plane services,
-while the latter will be used to manage IPAM for data plane nodes.
+[IMPORTANT]
+Do not change adopted node IP addresses during the adoption process. List the IP addresses in the `fixedIP` fields in the `OpenStackDataPlaneNodeSet` CR `per-node` section.
```
klgill, line 14: Is “per-node” the actual name of the section? I’m unable to find it in other guides, so I want to confirm.

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -46,59 +26,53 @@ network_config:
...
+----
+$ openstack overcloud node provision
+--stack <stack>
+--network_config
+----
++
+[NOTE]
+Network configuration changes are not applied by default to avoid
+the risk of network disruption. You must enforce the changes by setting the
 `StandaloneNetworkConfigUpdate: true` in the {OpenStackPreviousInstaller} configuration files.
```
klgill, line 46: Please carefully review steps 1 and 2  and ensure that the order is logical.  Also, the `openstack overcloud node provision` command in step 2 is not complete. Can you help provide a complete command that the customers would run?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -46,59 +26,53 @@ network_config:
...
++
+As a result, the following route is added to your {OpenShiftShort} nodes:
++
+----
 # ip route | grep 192
 192.168.122.0/24 dev ospbr proto static scope link
-```
+----
++
+At this point, you can ping the adopted nodes from the {OpenShiftShort} nodes by using the old subnet IP addresses.
```
klgill, line 71: Can we provide a way for customers to verify this? And what should they do if they can’t ping the adopted nodes with the old subnet addresses?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -46,59 +26,53 @@ network_config:
...
-
----
-
-
-Finally, during the data plane adoption, you will have to take care of several aspects:
-
-- in network_config, add link local routes to the new subnets, for example:
-
-```yaml
+. During the data plane adoption, in the `network_config` section of the `NetworkAttachmentDefinition` CR, add link local routes to the new subnets. For example:
```
klgill, line 73: The last 3 steps of this procedure can only be done while the customer is adopting the data plane?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -127,25 +101,20 @@ Finally, during the data plane adoption, you will have to take care of several a
...
-the old one.
-
----
-
-Since you are applying new network configuration to the nodes, consider also
-setting `edpm_network_config_update: true` to enforce the changes.
+----
++
+This provides SSH access from the new subnet to the adopted nodes as well as
+the old nodes.
```
klgill, line 115: Is “old nodes” correct?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -127,25 +101,20 @@ Finally, during the data plane adoption, you will have to take care of several a
...
 
----
+. Optional: Set `edpm_network_config_update: true` to enforce the changes that you are applying to the nodes:
```
klgill, line 117: Where do customers set the edpm_network_config_update: true? Can we please provide a code example in this step?

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -127,25 +101,20 @@ Finally, during the data plane adoption, you will have to take care of several a
...
 
 Note that the examples above are incomplete and should be incorporated into
```
klgill, line 119: I’m not sure what purpose this serves or what we are trying to say. Are we trying to say that the examples don’t necessarily reflect the customers’ configuration and that they need to use values that are specific to their configuration?

docs_user/modules/proc_reusing-existing-subnet-ranges.adoc
```diff
@@ -1,25 +1,14 @@
 [id="reusing-existing-subnet-ranges_{context}"]
 
-= Scenario 2: Reusing existing subnet ranges
+= Reusing existing subnet ranges
 
-This scenario is only applicable when the existing subnet ranges have enough IP
-addresses for the new control plane services. On the other hand, it allows to
-avoid additional routing configuration between the old and new subnets, as in xref:using-new-subnet-ranges_{context}[Scenario 1: Using new subnet ranges].
+You can reuse existing subnet ranges if they have enough IP addresses to allocate to the new control plane services. You configure the new control plane services to use the same subnet as you used in the adopted environment, and configure the allocation pools that are used by the new services to exclude IP addresses that are already allocated to existing cluster nodes. By reusing existing subnets, you avoid additional routing configuration between the old and new subnets. as in xref:using-new-subnet-ranges_{context}[Configuring new subnet ranges].
```
klgill, line 5: Shouldn’t "adopted environment" say the "RHOSP environment" or the " previous environment"?

docs_user/modules/proc_reusing-existing-subnet-ranges.adoc
```diff
@@ -1,25 +1,14 @@
...
 
-This scenario implies that the remaining IP addresses in the existing subnet is
-enough for the new control plane services. If not,
-xref:using-new-subnet-ranges_{context}[Scenario 1: Using new subnet ranges] should be used
-instead. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
-
-No special routing configuration is required in this scenario; the only thing
-to pay attention to is to make sure that already consumed IP addresses don't
-overlap with the new allocation pools configured for {rhos_prev_long} control plane services.
+No special routing configuration is required to reuse subnet ranges. However, you must ensure that already consumed IP addresses do not overlap with the new allocation pools configured for {rhos_long_noacro} control plane services.
```
klgill, line 9: "Already consumed" = consumed by the RHOSO deployment?

docs_user/modules/proc_reusing-existing-subnet-ranges.adoc
```diff
@@ -1,25 +1,14 @@
...
 
 If you are especially constrained by the size of the existing subnet, you may
```
klgill, line 11: The last sentence trails off. What is the link supposed to be?
Also, does “especially constrained by the size of the existing subnet” mean that the subnet does not have enough IP addresses?

docs_user/modules/proc_reusing-existing-subnet-ranges.adoc
```diff
@@ -1,25 +1,14 @@
 [id="reusing-existing-subnet-ranges_{context}"]
 
-= Scenario 2: Reusing existing subnet ranges
+= Reusing existing subnet ranges
```
klgill, line 3: General comment:  I’m confused by the purpose of this module. Is any configuration required at all by the user? If not, is the purpose of this module to just explain what reusing subnet ranges means, and that no additional configuration is required?

docs_user/modules/proc_reusing-existing-subnet-ranges.adoc
booxter: I'm not exactly sure but looks like it was garbled in 5a1d7566c9e401a9c91399cb23fa16365e83dd70? And you can find the original text from the section here: 2c88da094b44d4400026b022918fd58d07a18c07 (It was referring to a "section below" before the split.)

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: Yes, `data plane`.

---

The "old" term is probably not the best because they are still used in the "adopted" / "new" environment. They are subnets that were used before adoption, but they are still current / used in the new.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: Already consumed by the RHOSP (pre-adoption) environment. Context: The reason why the existing pools don't have special requirements is because they will always be "enough", because any new control plane IPs will be borrowed from the new subnet ranges; while the "old"/"previous" ranges will always be large enough to fit data plane needs - because during adoption the number of IP addresses consumed by "data plane" doesn't go up.

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -1,43 +1,23 @@
...
+plane services. There are no specific requirements for the
+existing deployment allocation pools already consumed.
 
-The new subnet should be sized appropriately to accommodate the new control
-plane services, but otherwise doesn't have any specific requirements as to the
-existing deployment allocation pools already consumed. Actually, the
-requirements as to the size of the new subnet are lower than in the second
-scenario, as the old subnet ranges are kept for the adopted nodes, which means
-they don't consume any IP addresses from the new range.
+In the following procedure, you configure `NetworkAttachmentDefinition` custom resources (CRs) to use a different subnet from what is configured in the `NetConfig` CR for the same networks. The former range is used for control plane services, while the new range is used to manage IP Address Management (IPAM) for data plane nodes.
```
booxter, line 11: It's vice versa: the "old" is used for data plane, while the "new" is for the NEW control plane.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: No, there is no `per-node`-named section.

An example of how this is configured can be traced e.g. here:

1. [The section with all nodes.](https://github.com/openstack-k8s-operators/openstack-operator/blob/b61660cac429b4c48a403d4bf9597c218782e722/config/samples/dataplane/networker/values.yaml#L18)
2. [Inside the section, an entry per node.](https://github.com/openstack-k8s-operators/openstack-operator/blob/b61660cac429b4c48a403d4bf9597c218782e722/config/samples/dataplane/networker/values.yaml#L19)
3. [Inside each entry, the field with the fixedIP.](https://github.com/openstack-k8s-operators/openstack-operator/blob/b61660cac429b4c48a403d4bf9597c218782e722/config/samples/dataplane/networker/values.yaml#L27)

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -1,43 +1,23 @@
...
 
----
-
-Before proceeding, configure host routes on the adopted nodes for the
-control plane subnets.
-
-To achieve this, you will need to re-run `openstack overcloud node provision` with additional `routes` entries added to `network_config`. (This change should be applied for every adopted node configuration.) For example, you may add the following to `net_config.yaml`:
-
-```yaml
+. In the `NetworkAttachmentDefinition` CR, configure host routes on the adopted nodes for the control plane subnets. For example:
```
booxter, line 18: This is incorrect. The example below is part of TripleO configuration and it demonstrates how to configure static routes there (before adoption proper is started). It is *not* `NetworkAttachmentDefinition`. (TripleO configuration is a custom .yaml file that is not a CR - CR being `Custom Resource` in OpenShift API parlance.)

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: I am not sure which steps you refer to here, and which command you think is incomplete. Please clarify.

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -46,59 +26,53 @@ network_config:
...
-
-```bash
+----
+<1> The original subnet of the isolated network on the data plane.
+<2> The {rhocp_long} worker network interface that corresponds to the isolated network on the data plane.
++
+As a result, the following route is added to your {OpenShiftShort} nodes:
++
+----
 # ip route | grep 192
```
booxter, line 67: I now think we may want to avoid advising customers to run commands directly on CoreOS shell, as I think it's generally discouraged in OpenShift world. Perhaps the `ip route` and `ping` checks here should be avoided. I wonder what @jistr would say.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: I am not sure we should, see above.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter, line 73: This is not referring to `NetworkAttachmentDefinition`. The configuration will go into `OpenStackDataPlaneNodeSet` CR.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: I think this sentence refers to "old subnets", not "old nodes". (It's communicating that the example above sets up SSH daemon running on data plane nodes to accept connections from two ranges - one that is adopted from the pre-adopted setup (172.x in the example), and one subnet that is newly defined for the new control plane services (192.x).

---

As alluded above in a different comment, we may want to use a different word for "old", though it probably applies for any use of "old" in the documentation, not just in networking / subnet context. These are not really "old" as much as "resources / nodes / subnets that originate from the setup before adoption".

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: I can find an example, but first I'd like to check with e.g. @slagle or @dprince if we really need to add this setting when we do the initial adoption. Guys, is it really needed?

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: Yes, this was the intent. I guess it's redundant, if we assume that readers are careful. One can only hope! (But if they are not, this note won't help much either.) I'm ok if we just remove it.

booxter (changes requested):
There are a bunch of issues that will have to be addressed before merge.

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
klgill: @booxter It's okay to leave the TODO comments. I've just been calling them out to remind engineers to either address them, or if they have been addressed, to delete them.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: @booxter Can you please review lines 18 - 45 in docs_user/modules/proc_using-new-subnet-ranges.adoc? I hope that helps.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: Maybe "RHOSP" or "previous" would be a better term.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: I think we can keep it, but it should be moved to the beginning of the procedure and rewritten, e.g.,:

"The values that are used in the following procedure are examples. Use values that are specific to your configuration."

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: OK, maybe something like the following instead?

"In your network configuration file, configure host routes on the RHOSP nodes for the control plane subnets. For example:"

I'm not sure what the custom yaml file should be called. I guess it varies depending on the customer, but I'm trying to help the user understand which file they are supposed to be updating.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: "Existing" subnets might be better.

docs_user/modules/proc_reusing-existing-subnet-ranges.adoc
klgill: It looks like it should link to: xref:configuring-isolated-networks_configuring-network[Configuring isolated networks].

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: So it should be, "The former range is used to manage IP Address Management (IPAM) for data plane nodes, while the new range is used for the new control plane services."

booxter: Yeah. Though the "former" here seems a bit ambiguous to me. Maybe word out exactly what is meant instead.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: It probably can use a different name, but AFAIU the `net_config.yaml` is a common (default) name, so we may mention it as a likely example.

---

Yes, your new wording is better. Thanks.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: Ack. I think I did already.

---

As to the incomplete command,

```
openstack overcloud node provision
--stack <stack>
--network_config
```

Indeed this looks weird. I am not very knowledgeable about TripleO. I wonder if @jistr can dig for a more complete example on how to update nodes' network configuration.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: @booxter I edited this section further to change "NetConfig CR" to "`network_config` section of the `OpenStackDataPlaneNodeSet` CR. Here is the new text:

"...you configure `NetworkAttachmentDefinition` custom resources (CRs) to use a different subnet from what is configured in the `network_config` section of the `OpenStackDataPlaneNodeSet` CR for the same networks. The range in the `NetworkAttachmentDefinition` CR is used for control plane services, while the new range in the `OpenStackDataPlaneNodeSet` CR is used to manage IP Address Management (IPAM) for data plane nodes."

docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
```diff
@@ -4,43 +4,24 @@ ifdef::context[:parent-context: {context}]
...
-// TODO: should we parametrize the version in the links somehow?
-
-When adopting a new {OpenStackShort} deployment, it is important to align the network
+When you adopt a new {rhos_long} deployment, you must align the network
 configuration with the adopted cluster to maintain connectivity for existing
 workloads.
 
-The following logical configuration steps will incorporate the existing network
-configuration:
+You need to incorporate the existing network configuration in the following steps:
```
igallagh-redhat, line 13:
```suggestion
Perform the following tasks to incorporate the existing network configuration:
```
Suggestion to avoid using "steps" here cos it isn't a procedure, sounds more like a workflow.

docs_user/assemblies/assembly_configuring-network-for-RHOSO-deployment.adoc
```diff
@@ -4,43 +4,24 @@ ifdef::context[:parent-context: {context}]
...
-* configure **{OpenShiftShort} worker nodes** to align VLAN tags and IPAM
-  configuration with the existing deployment.
-* configure **Control Plane services** to use compatible IP ranges for
-  service and load balancing IPs.
-* configure **Data Plane nodes** to use corresponding compatible configuration for
-  VLAN tags and IPAM.
+* You configure {rhocp_long} worker nodes to align VLAN tags and IP Address Management (IPAM) configuration with the existing deployment.
+* You configure control plane services to use compatible IP ranges for service and load-balancing IP addresses.
+* You configure data plane nodes to use corresponding compatible configuration for VLAN tags and IPAM.
+* You can either reuse IPAM configuration from the existing deployment or, depending on IP address availability in the existing allocation pools, define new ranges to be used for the new control plane services. If you define new ranges, you configure IP routing between the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
```
igallagh-redhat, line 18:
```suggestion
* Reuse IPAM configuration from the existing deployment or, depending on IP address availability in the existing allocation pools, define new ranges for the new control plane services. If you define new ranges, you configure IP routing between the old and new ranges. For more information, see xref:planning-your-ipam-configuration_configuring-network[Planning your IPAM configuration].
```

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
```diff
@@ -6,65 +6,39 @@ ifdef::context[:parent-context: {context}]
...
-  number depends on the number of replicas for each service.
-* The number of IP addresses required for load balancer IP addresses. (Each
-  service will require a VIP address from `IPAddressPool` CRs.)
-
-As of the time of writing, the simplest single worker node {OpenShiftShort} deployment
-(CRC) has the following IP ranges defined (for the `internalapi` network):
-
-* 1 IP address for the single worker node;
-* 1 IP address for the data plane node;
+number of IP addresses available for services, you cannot reuse the existing IP ranges in the {rhos_acro} environment, depending on the size.
```
igallagh-redhat, line 12: "Since the new control plane deployment has different requirements as to the ..."
IBM Style guide: 
"since - Use with caution. To facilitate translation, use this word only to refer to time. Do not use it as a synonym for because."

```suggestion
The new control plane deployment has different requirements for the number of IP addresses that are available for services, therefore you cannot reuse the existing IP ranges in the {rhos_acro} environment.
```
OR

```suggestion
The number of IP addresses that are made available for services is different for the new control plane deployment than the {rhos_acro} environment. Therefore, you cannot use the IP ranges from the existing {rhos_acro} environment in the new control plane environment.
```

What is ", depending on the size" relating to/trying to convey?

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
```diff
@@ -6,65 +6,39 @@ ifdef::context[:parent-context: {context}]
...
+
+The total number of IP addresses that are required for the new control plane services in each isolated network is calculated as the sum of the following:
+
+* The number of {OpenShiftShort} worker nodes. Each node requires 1 IP address in the `NodeNetworkConfigurationPolicy` custom resources (CRs).
```
igallagh-redhat, line 16:
```suggestion
* The number of {OpenShiftShort} worker nodes. Each node requires 1 IP address in each `NodeNetworkConfigurationPolicy` custom resource (CR).
```

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
```diff
@@ -6,65 +6,39 @@ ifdef::context[:parent-context: {context}]
...
+* The number of IP addresses required for the data plane nodes. Each node requires an IP address from the `NetConfig` CRs.
```
igallagh-redhat, line 17: I think you just create one NetConfig CR for the data plane network:

```suggestion
* The number of IP addresses required for the data plane nodes. Each node requires an IP address from the `NetConfig` CR.
```

docs_user/modules/proc_retrieving-network-information-from-your-existing-deployment.adoc
```diff
@@ -18,20 +26,13 @@ file. For example,
...
-into separate pools for control plane services and load balancer IP addresses.
-
-You should also determine the list of IP addresses already consumed in the
-adopted environment. Consult `tripleo-ansible-inventory.yaml` file to find this
-information. In the file, for each listed host, note IP and VIP addresses
-consumed by the node.
+----
 
-For example,
+. In the `network_data.yaml` file, make a note of the VLAN tag that is used in the `vlan` key and the IP range in the `ip_subnet` key for each isolated network. The IP range is later split into separate pools for control plane services and load balancer IP addresses.
```
igallagh-redhat, line 31:
```suggestion
. Retrieve the VLAN tag that is used in the `vlan` key and the IP range in the `ip_subnet` key for each isolated network from the `network_data.yaml` file. The IP range is later split into separate pools for control plane services and load balancer IP addresses.
```

igallagh-redhat (commented):
Peer reviewed, added some suggestions and questions.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: @jistr  or @hjensas Can you help provide a more complete example of an `overcloud node provision` command?

hjensas: There is an example in OSP 17 docs: https://docs.redhat.com/en/documentation/red_hat_openstack_platform/17.1/html/installing_and_managing_red_hat_openstack_platform_with_director/assembly_provisioning-and-deploying-your-overcloud#proc_manually-configuring-bare-metal-node-hardware-information_ironic_provisioning

Step 10:
```
(undercloud)$ openstack overcloud node provision \
 [--templates <templates_directory> \]
 --stack <stack> \
 --network-config \
 --output <deployment_file> \
 /home/stack/templates/<node_definition_file>
```

The example here is missing the `--output <deployment_file>` argument and the path to input `<node_definition_file>`.

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
klgill: @booxter Can you confirm that you create one NetConfig CR for the data plane network? See the suggested edit, above^.

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: @jistr ^

klgill:
I am going to merge this PR. The rest of the comments that have been addressed or yet to be addressed will be added in a a new PR.

docs_user/modules/proc_using-new-subnet-ranges.adoc
hjensas: It looks like several steps involved to pre-populates routes on `{OpenStackPreviousInstaller}` is missing from this doc? It depends on the NIC config templates used in the OSP17 deployment, but typically the routes would have to be added to to subnets in `network_data.yaml` -> then the `openstack overcloud network provision` command must be used to update the networks and then `openstack overcloud node provision` / `openstack overcloud deploy` commands must re-run.

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -1,43 +1,25 @@
...
-Before proceeding, configure host routes on the adopted nodes for the
-control plane subnets.
-
-To achieve this, you will need to re-run `openstack overcloud node provision` with additional `routes` entries added to `network_config`. (This change should be applied for every adopted node configuration.) For example, you may add the following to `net_config.yaml`:
-
-```yaml
+. In the `NetworkAttachmentDefinition` CR, configure host routes on the {OpenStackShort} nodes for the control plane subnets. For example:
++
+----
 network_config:
```
hjensas, line 23: This does not look like a `NetworkAttachmentDefinition`[1] at all, this looks like config for os-net-config.

[1] https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#networkattachmentdefinition-with-json-cni-config

hjensas (changes requested):

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: CC: @booxter

docs_user/modules/proc_using-new-subnet-ranges.adoc
klgill: CC: @booxter

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: I'm looking at the latest version of the doc, and I think it was messed by some changes. E.g. this diff is incorrect:

```diff
----
-
-Before proceeding, configure host routes on the adopted nodes for the
-control plane subnets.
-
-To achieve this, you will need to re-run `openstack overcloud node provision` with additional `routes` entries added to `network_config`. (This change should be applied for every adopted node configuration.) For example, you may add the following to `net_config.yaml`:
-
-```yaml
+. In the `NetworkAttachmentDefinition` CR, configure host routes on the adopted nodes for the control plane subnets. For example:
++
+----
```

which came with 8476a2630fb50427e370ca7cbe451d3bf0bfb885.

The change above suggests that `network_config` section belongs to `NetworkAttachmentDefinition` now. This is incorrect. I assume `network_data.yaml` that @hjensas mentioned is the `net_config.yaml` the previous version of the document suggested.

(The change as merged is quite concerning. I think I can't vouch for the document without doing another re-reading from scratch.)

---

I would also rely on @hjensas suggestion that both `openstack overcloud network provision` and `openstack overcloud node provision` are needed. This part of the original instructions were written without validation on a real TripleO cluster due to limited availability of such in adoption test env; where we ended up starting the `standalone` env with the desired static routes preconfigured instead of going through the `network provision` step. So the instruction may be off.

docs_user/modules/proc_using-new-subnet-ranges.adoc
booxter: Yes, as commented in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/641#discussion_r1805124736, there were changes to the document that changed the desired meaning. Now that this is merged, someone from Networking team should sit down and re-read the whole section to catch such issues.

docs_user/assemblies/assembly_planning-your-ipam-configuration.adoc
booxter: Correct. Please apply this change.

PR #643, opened 2024-10-04
Fix inconsistent uses of storageClass

jistr:
- Advise in docs that 'local-storage' should be changed according to the OCP env.

- Use the `storage_class_name` variable consistently in tests.

- Remove explicit `storageClass: local-storage` from places that should inherit the storageClass from the `OpenStackControlPlane` CR.

- Default `storage_class_name` to `local-storage`.

### DIFF
--- a/tests/roles/glance_adoption/files/glance_ceph.yaml
+++ b/tests/roles/glance_adoption/files/glance_ceph.yaml
@@ -15,7 +15,6 @@ spec:
         rbd_store_pool=images
         store_description=Ceph glance store backend.
       storage:
-        storageClass: "local-storage"
         storageRequest: 10G
       glanceAPIs:
         default:
--- a/tests/roles/glance_adoption/files/glance_cinder.yaml
+++ b/tests/roles/glance_adoption/files/glance_cinder.yaml
@@ -6,7 +6,6 @@ spec:
     template:
       databaseInstance: openstack
       storage:
-        storageClass: "local-storage"
         storageRequest: 10G
       customServiceConfig: |
           [DEFAULT]
--- a/tests/roles/swift_adoption/defaults/main.yaml
+++ b/tests/roles/swift_adoption/defaults/main.yaml
@@ -11,7 +11,7 @@ swift_patch: |
           replicas: 0
           networkAttachments:
           - storage
-          storageClass: local-storage
+          storageClass: {{ storage_class_name }}
           storageRequest: 10Gi
         swiftProxy:
           secret: osp-secret
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -198,7 +198,7 @@ metadata:
   name: openstack
 spec:
   secret: osp-secret
-  storageClass: local-storage
+  storageClass: local-storage <1>
 
   barbican:
     enabled: false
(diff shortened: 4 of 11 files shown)


cescgina:
this lgtm @jistr just a note that even with this change, the jobs use `crc-csi-hostpath-provisioner` storage class https://logserver.rdoproject.org/43/643/b4ff38a66b1de722363b4434370500eb66304b15/github-check/adoption-standalone-to-crc-no-ceph/3bda133/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/pvc.log that's because it's set on the rdo-jobs side by default https://github.com/rdo-infra/rdo-jobs/blob/eea4ec0871e8139329e93b269c79768e6f9fabce/playbooks/data_plane_adoption/vars.yaml#L20, so if we decide to go with local-storage we'll need to change that

jistr:
On the default changing i'm actually not sure. I know the RHEV environment overrides it back to local-storage, for example. But perhaps it would make sense to keep the crc-csi-hostpath-provisioner as the default for now, to make it "friendly by default" to the CRC-based dev envs...

tests/roles/glance_adoption/files/glance_cinder.yaml
```diff
@@ -6,7 +6,6 @@ spec:
     template:
       databaseInstance: openstack
       storage:
-        storageClass: "local-storage"
```
jistr, line 9: @fmount Do you think it's ok to remove these?

tests/roles/swift_adoption/defaults/main.yaml
```diff
@@ -11,7 +11,7 @@ swift_patch: |
           replicas: 0
           networkAttachments:
           - storage
-          storageClass: local-storage
+          storageClass: {{ storage_class_name }}
```
jistr, line 14: @cschwede Do you think we could remove the storageClass setting from here altogether? (Would it inherit from the OpenStackControlPlane CR setting?) The reason i didn't remove it yet is that it's actually explicitly mentioned in docs:

https://openstack-k8s-operators.github.io/data-plane-adoption/user/#adopting-the-object-storage-service_adopt-control-plane 

"Must match the RHOSO deployment storage class."

That makes me wonder whether we really need to repeat the setting here, or whether we can inherit from the storageClass setting on the control plane.

tests/roles/glance_adoption/files/glance_cinder.yaml
jistr: I mean, would it then just use what was in OpenStackControlPlane CR as the storageClass?

fmount: Hey @jistr yes, I think the storageClass can be inherited by default from the top-level CR (OpenStackControlPlane), so it can be safely removed here!

tests/roles/glance_adoption/files/glance_ceph.yaml
```diff
@@ -15,7 +15,6 @@ spec:
         rbd_store_pool=images
         store_description=Ceph glance store backend.
       storage:
-        storageClass: "local-storage"
```
fmount, line 18: same comment applies here.

jistr:
Based on the comments and discussion i think this can be merged as-is and if we'd like to tweak the Swift instructions (remove the storageClass altogether), we can do that in a separate PR.

cescgina (approved):
+1, this will be useful for the uni jobs, thanks!

PR #649, opened 2024-10-10
Validate Cinder operations

rajathere:
After the adoption, we check for the correct status and size of the
Cinder resources but don't validate that operations still work on them.
This patch adds the following operations:
1. Detach 'disk' volume from 'test' server
2. Revert to snapshot 'snapshot' to 'disk' volume
3. Attach 'disk' volume to 'test' server

This will help verify that the resources are in consistent state before
and after the adoption.

NOTE: We are skipping restore backup since it's	failing	even with 20
retries	with 10 second delay. Will add that in a follow up PR after
resolving the issues.

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/cinder_verify.yaml
+++ b/tests/roles/dataplane_adoption/tasks/cinder_verify.yaml
@@ -29,3 +29,60 @@
   until: cinder_verify_resources is success
   retries: 10
   delay: 2
+
+- name: Detach disk volume
+  when:
+    - prelaunch_test_instance|bool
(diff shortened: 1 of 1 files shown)


rajathere:
recheck heat isssue shoud be fixed by https://review.rdoproject.org/r/c/rdo-jobs/+/55212

jistr (approved):

fmount (approved):
/lgtm

bogdando (approved):
/lgtm

jistr:
Backport included in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/709

PR #655, opened 2024-10-14
Sync rollback playbooks with their non-rollback counterparts

jistr:


frenzyfriday:
recheck

jistr:
> Would it make more sense to import non-rollback playbook (minimal or with_ceph in case of rollback playbooks you syncing ) and just run rollback role after import?

We'd have to split it too, we're not importing the whole playbooks, just the control plane part of the adoption. But it's a good point, i think this should probably be done so that we don't have to keep doing these syncs. I'd just do it as a refactoring later, because we're now trying to get these jobs green and into criteria asap.

PR #657, opened 2024-10-14
Add tag to allow skipping os diff task

eshulman2:
added a missing tag to allow skipping os-diff in dataplane adoption


marios (approved):
lgtm

jistr (approved):

PR #659, opened 2024-10-15
Add Ceph post configuration section

fmount:
This patch adds the Ceph post migration procedure. At the moment the only thing that should be reconfigured is the Ceph Dashboard failover, but this guide might be extended according to the feedback we receive from the field.
The post-rbd verification is also improved to verify mons are running using the old IP addresses.

### DIFF
--- a/docs_user/modules/proc_migrating-ceph-post.adoc
+++ b/docs_user/modules/proc_migrating-ceph-post.adoc
@@ -1,0 +1,27 @@
+[id="updating-the-cluster-dashboard-configuration_{context}"]
+
+= Updating the {CephCluster} cluster Ceph Dashboard configuration
+
+If the Ceph Dashboard is part of the enabled Ceph Manager modules, you need to
+reconfigure the failover settings.
+
+. Regenerate the following {Ceph} configuration keys to point to the right
+  `mgr` container:
++
+----
+mgr    advanced  mgr/dashboard/controller-0.ycokob/server_addr  172.17.3.33
+mgr    advanced  mgr/dashboard/controller-1.lmzpuc/server_addr  172.17.3.147
+mgr    advanced  mgr/dashboard/controller-2.xpdgfl/server_addr  172.17.3.138
+----
++
+----
+$ sudo cephadm shell
+$ ceph orch ps | awk '/mgr./ {print $1}'
+----
+
+. For each retrieved `mgr` daemon, update the corresponding entry in the {Ceph}
+  configuration:
++
+----
+$ ceph config set mgr mgr/dashboard/<>/server_addr/<ip addr>
+----
--- a/docs_user/modules/proc_migrating-mon-from-controller-nodes-verification.adoc
+++ b/docs_user/modules/proc_migrating-mon-from-controller-nodes-verification.adoc
@@ -6,13 +6,21 @@ After you finish migrating your Ceph Monitor daemons to the target nodes, verify
 
 .Procedure
 
-* Verify that the {CephCluster} cluster is healthy:
+. Verify that the {CephCluster} cluster is healthy:
 +
 ----
-[ceph: root@oc0-controller-0 specs]# ceph -s
+$ ceph -s
   cluster:
     id:     f6ec3ebe-26f7-56c8-985d-eb974e8e08e3
     health: HEALTH_OK
 ...
 ...
+----
+
+. Verify that the {CephCluster} mons are running with the old IP addresses. SSH
+  into the target nodes and verify that the Ceph Monitor daemons are bound to
+  the expected IP and port:
++
+----
+$ netstat -tulpn | grep 3300
 ----
--- a/docs_user/assemblies/assembly_migrating-ceph-cluster.adoc
+++ b/docs_user/assemblies/assembly_migrating-ceph-cluster.adoc
@@ -37,5 +37,7 @@ include::assembly_migrating-ceph-rgw.adoc[leveloffset=+1]
 
 include::assembly_migrating-ceph-rbd.adoc[leveloffset=+1]
 
+include::../modules/proc_migrating-ceph-post.adoc[leveloffset=+1]
+
 ifdef::parent-context[:context: {parent-context}]
 ifndef::parent-context[:!context:]


docs_user/modules/proc_migrating-mon-from-controller-nodes-verification.adoc
```diff
@@ -9,10 +9,19 @@ After you finish migrating your Ceph Monitor daemons to the target nodes, verify
 * Verify that the {CephCluster} cluster is healthy:
 +
 ----
-[ceph: root@oc0-controller-0 specs]# ceph -s
+$ ceph -s
```
fmount, line 12: @klgill I think this section will be extended when we get feedback from the field.

docs_user/modules/proc_migrating-ceph-post.adoc
```diff
@@ -0,0 +1,27 @@
+[id="updating-the-cluster-mgr-configuration_{context}"]
+
+= Updating the {CephCluster} cluster mgr dashboard configuration
```
fmount, line 3: @klgill right now, as per our conversation, the post section is about fixing leftovers from the previous deploy. For now I'm only thinking about the dashboard failover, but TLS might be a good candidate to extend this section. Considering it's day 2 and I don't have a specific procedure, I would just keep this and extend it when we have more content.

klgill: I think we should avoid using "mgr" in the title. Would "Updating the {CephCluster} cluster with Ceph Manager modules that include Ceph Dashboard"  make sense?  I know that's a mouthful...

docs_user/modules/proc_migrating-mon-from-controller-nodes-verification.adoc
```diff
@@ -9,10 +9,19 @@ After you finish migrating your Ceph Monitor daemons to the target nodes, verify
 * Verify that the {CephCluster} cluster is healthy:
```
klgill, line 9:
```suggestion
. Verify that the {CephCluster} cluster is healthy:
```
Now that we have more than 1 step, I'm changing the steps to a numbered list.

docs_user/modules/proc_migrating-mon-from-controller-nodes-verification.adoc
```diff
@@ -9,10 +9,19 @@ After you finish migrating your Ceph Monitor daemons to the target nodes, verify
...
     id:     f6ec3ebe-26f7-56c8-985d-eb974e8e08e3
     health: HEALTH_OK
 ...
 ...
 ----
+
+Verify that the {CephCluster} mons are running with the old IP addresses.
+
+* SSH into the target nodes and verify the mon daemons are bound to the
+  expected IP and port:
```
klgill, line 23:
```suggestion
. Verify that the {CephCluster} mons are running with the old IP addresses. SSH into the target nodes and verify that the Ceph Monitor daemons are bound to the expected IP and port:
```

docs_user/modules/proc_migrating-ceph-post.adoc
klgill: Also, I want to confirm the sequence of events. Would customers perform this procedure immediately after the "Verifying the Red Hat Ceph Storage cluster after Ceph Monitor migration" procedure?

docs_user/modules/proc_migrating-ceph-post.adoc
```diff
@@ -0,0 +1,27 @@
...
+
+If the Ceph Dashboard is part of the enabled mgr modules, you need to
```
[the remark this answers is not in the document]
fmount: Done (thank you!)

docs_user/modules/proc_migrating-mon-from-controller-nodes-verification.adoc
fmount: Ack, makes sense! (Fixed)

docs_user/modules/proc_migrating-ceph-post.adoc
fmount: yes about the sequence, customers are supposed to perform this part right after checking the cluster status.
Regarding the title, I think that suggestion would change the meaning here.
Basically we're going to update the Ceph dashboard configuration, I think we don't have to specify that is Ceph Manager module.
I'll push an update and let me know if it's better than what we currently have.

docs_user/modules/proc_migrating-ceph-post.adoc
```diff
@@ -0,0 +1,27 @@
+[id="updating-the-cluster-dashboard-configuration_{context}"]
+
+= Updating the {CephCluster} cluster Dashboard configuration
```
klgill, line 3:
```suggestion
= Updating the {CephCluster} cluster Ceph Dashboard configuration
```

One nit, because I think "Ceph Dashboard" is the full official name.

docs_user/modules/proc_migrating-ceph-post.adoc
klgill: Regarding the sequence, would it be more logical to put this procedure as the last step in the "Migrating Ceph Monitor daemons to Red Hat Ceph Storage nodes" assembly?

docs_user/modules/proc_migrating-ceph-post.adoc
fmount: ack works for me.

docs_user/modules/proc_migrating-ceph-post.adoc
fmount: When I think about a "post procedure" I imagine something generic that has the Dashboard update right now, but it might be extended to something else. For this reason, given this is a standalone procedure, it's part of the main migration guide, while adding in the migrate monitor daemons assembly let me thing that is somehow related to that section, while it's not.
If this is a concern feel free to reorganize the way you think it's better, I just based the location choice on a logical view of the main procedure.

klgill: OK thank you for the explanation. We can leave it where it is.

igallagh-redhat (approved):
LGTM

PR #664, opened 2024-10-18
[ovn] Add iptables rules on controllers for scenario A

averdagu:
If the adoption is done with different subnets between the wallaby environment and the podified controlplane, it will fail on the ovn dump database since on the wallaby environment there's a firewall rule which only enables traffic to the NB/SB DB if it belongs to the same wallaby internal_api network.

If the adoption has different subnets (wallaby internal_api != podified internal_api) a new firewall rule is needed to accept all traffic comming from podified internal_api.


tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -143,6 +143,24 @@
...
+# filter allowing traffic from podified network is needed. Adding the rules
+# no matter the scenario to simplify logic and on a non-scenario-A environment
+# this rule shouldn't harm.
+- name: Add iptables rule to allow podified internalapi trafic to controllers
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    {{ ovn_copy_shell_vars }}
+
+    $CONTROLLER1_SSH sudo iptables -I INPUT -s 172.17.1.0/24 -p tcp -m tcp --dport 6641 -m conntrack --ctstate NEW -m comment --comment \'121 OVN DB server and cluster ports for 172.17.1.0/24 ipv4\' -j ACCEPT
```
jistr, line 157: This is a good default but going forward we'll probably want to template this out and have a var in `defaults/main.yaml`, at least the subnet range if not the whole command? Not a blocker as i think this won't hurt our test envs, but would be nice to have i think.

docs_user/modules/proc_using-new-subnet-ranges.adoc
```diff
@@ -8,6 +8,9 @@ You must size the new subnet appropriately to accommodate the new control
 plane services. There are no specific requirements for the
 existing deployment allocation pools that are already consumed by the {OpenStackShort} environment.
 
+[IMPORTANT]
+Defining a new subnet for Storage it is not supported.
```
jistr, line 12: What's the rationale behind this?

jistr (approved):

docs_user/modules/proc_using-new-subnet-ranges.adoc
averdagu: You mean why we can't modify storage network during adoption? If that's the case, the reason it's related to nova and cinder IIRC, nova doesn't support changing the network where it stores the VM's volumes. (I'm not an expert neither on nova nor cinder, will try to get a better explanation on nova/cinder team to proper include an explanation)

tests/roles/ovn_adoption/tasks/main.yaml
averdagu: Totally agree, will create an issue to modify this on a near future (not doing now since FR1 is around the corner)

averdagu:
Error: "Task 
Register node with red hat subscription for data plane adoption failed running on host standalone" not caused by patch

jistr:
:eyes: If the CI comes back green then IMO let's merge.

tests/roles/ovn_adoption/tasks/main.yaml
```diff
@@ -143,6 +143,24 @@
     $CONTROLLER2_SSH sudo systemctl stop tripleo_ovn_cluster_northd.service
     $CONTROLLER3_SSH sudo systemctl stop tripleo_ovn_cluster_northd.service
 
+# If ovn_adoption is done using scenario A (different networks between podified
+# and tripleo deployments) in order to be able to dump OVN database an iptable
```
slawqo, line 147: I think it should be "iptables", no?

jistr:
(We may want a backport to the fr1 branch?)

PR #681, opened 2024-10-24
Enable stale PR auto cleaning

ciecierski:
If a PR is open for more than 15 days without any activity, the bot will send a gentle reminder.

If nothing moves during the next 7 days, the bot will close the PR.

The same stale-pr workflow was introduced in ci-framework

### DIFF
--- a/.github/workflows/stale-pr.yml
+++ b/.github/workflows/stale-pr.yml
@@ -1,0 +1,17 @@
+---
+name: Manage stale PRs
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-pr-message: |
+            This PR is stale because it has been for over 15 days with no activity.
+            Remove stale label or comment or this PR will be closed in 7 days.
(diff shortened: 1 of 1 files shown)


ciecierski:
@jistr @holser There are already 30(including this one) open PRs. At least half of them with no activity in comments or new commits for some time already. What do you think about idea of closing stale PR requests? That way we can clearly see by looking at PR list what is actively being worked on, needs review etc.

jistr (approved):

odyssey4me (approved):

jistr:
Perhaps this doesn't trigger the github check, i'm not really sure. I'd suggest waiting for after FR1 is released and then bypassing the branch protections to merge, and we'll see what happens. IMO it should be ok to merge but let's wait not to disrupt FR1, just in case...

fao89 (approved):

ciecierski:
rdoproject.org/github-check is never going to run as .github/workflows is in irrelevant-files

ciecierski:
With noop job addition we don't need to bypass the branch protections to merge it.

jistr:
Cool, and FR1 is cut now so let the PR clean up begin :tornado: . Thanks :+1:

PR #684, opened 2024-10-25
Create additional manila resources

fmount:


fmount:
rebased

cescgina:
@fmount is this change only relevant to the ceph job? Should it run only there or it can run in both ceph and no-ceph jobs?

fmount:
this is required for manila w/ Ceph (because we enabled tempest there), but the task can be used for (and it's required by) other no-ceph jobs, so it's generic.

cescgina (approved):

fmount:
recheck

fmount:
rechecking after https://review.rdoproject.org/r/c/rdo-jobs/+/55212 landed.

gouthampacha (approved):
This looks good to me, thank you @fmount .. 
In the CI framework, we also set a couple of extra-specifications on this default share type for future proofing: https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/manila_create_default_resources.yml 

Something to consider if this randomly starts failing because new tests creep in..

fmount:
@cescgina @jistr feel free to land this patch now that we have green CI, so we can move forward w/ rdo-jobs and the remaining Storage services (cinder)!

tests/roles/manila_adoption/tasks/main.yaml
```diff
@@ -36,3 +36,13 @@
   register: manila_responding_result
   until: manila_responding_result is success
   retries: 15
+
+- name: Create default share type
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    alias openstack="oc exec -t openstackclient -- openstack"
+    ${BASH_ALIASES[openstack]} share type create {{ share_type_name }} {{ driver_handles_share_servers }}
```
marios, line 45: should we provide some default for that thought like there https://github.com/marios/data-plane-adoption/blob/aa7c61287ae041d7a6f79f14ea4ca471ef1e324a/tests/roles/manila_adoption/defaults/main.yaml#L1 for share_type_name

or do we need to make that configurable https://github.com/marios/data-plane-adoption/blob/aa7c61287ae041d7a6f79f14ea4ca471ef1e324a/tests/roles/manila_adoption/templates/manila_cephfs.yaml.j2#L41

fmount: I think having the command: `openstack share type create default false` is sufficient and generic enough for now.
We can think about customizing it in the context of unijobs implementation or multibackend, but right now I simply hardcoded `default` and `false` because these two parameters, in the adoption context, are always valid.
I'm planning to expand this part when the CI coverage will be extended further.

jistr (approved):

PR #686, opened 2024-10-25
hiding content that is TP for FR1

klgill:
PR that hides content that QE does not need to test, and provides QE with a clean preview of the guide.


jistr (approved):

klgill:
@jistr Just FYI, I'm not planning on merging this. It was only intended to hide content that Archana doesn't need to review. However, I don't want to merge this because we won't be hiding TP content from the final guide; the content will just get a Tech Preview disclaimer at the top.

PR #688, opened 2024-10-27
ospdo adoption makefile tests

pinikomarov:
Add Director Operator test execution scripts for adoption

Related CI patch and background:
https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/966

PRs submitted together :
[related PR list](https://issues.redhat.com/browse/OSPRH-2984?focusedId=25950569&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25950569)

Jiras:
https://issues.redhat.com/browse/OSPRH-1891
https://issues.redhat.com/browse/OSPRH-2984

PR #444, opened 2024-05-07
ospdo adoption rebased

pinikomarov:
Add Director Operator scripts for adoption

Related CI patch and background: 
https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/966

Jiras:
https://issues.redhat.com/browse/OSPRH-1891
https://issues.redhat.com/browse/OSPRH-2984

PR #689, opened 2024-10-28
ospdo adoption backend services

pinikomarov:
Add Director Operator backend services scripts for adoption

Related CI patch and background:
https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/966

Jiras:


jistr:
```
TASK [backend_services : execute alternative tasks when source env is ODPdO] ***
fatal: [localhost]: FAILED! => {"msg": "The conditional check 'ospdo_src| bool' failed. The error was: error while evaluating conditional (ospdo_src| bool):
```
We'll need to default that to `false`.

tests/roles/backend_services/defaults/main.yaml
```diff
@@ -18,3 +18,15 @@ heat_stack_domain_admin_password: ''
...
+# Whether source env is OSPD Director Operator
+# i.e. https://github.com/openstack-k8s-operators/osp-director-operator
+ospdo_src: false
+# rhoso namespace
+rhoso_namespace: "openstack"
+# director operator namespace
+org_namespace: "ospdo_openstack"
+# adoption repo default location
+dpa_dir: "/home/zuul/src/github.com/openstack-k8s-operators/data-plane-adoption"
+dpa_tests_dir: "{{ dpa_dir }}/tests"
```
jistr, line 32: The `dpa_dir` and `dpa_tests_dir` variables weren't in vars.sample.yaml or any defaults.yaml in #444 so i missed that we're creating them. Let's get rid of these two and keep using the same relative paths approach that we're using?

At least based on my understanding, there shouldn't be a reason to create these variables, and while they may be passing the checks right now, having them defined this way will break running the test suite locally. Going back to relative paths would fix it.

jistr (changes requested):

tests/roles/backend_services/defaults/main.yaml
pinikomarov: The reason for those was that while the original test suite scripts were running on the controller, the ospdo version is executed from controller node but running on the ospdo dedicated BM machine - the adoption_host.
so that relative scripts paths won't have the same meaning. 
```
tests/playbooks/test_minimal.yaml:
  hosts: "{{ adoption_host | default('localhost') }}"

```
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/688/files#diff-59a73c8a700093e77d026bfd1c67a98f868a6303fbc922472160294b35918fd4R5

pinikomarov: As an alternative I can move the scripts requiring that variable into the specific ospdo files and make them execute on the condition of ospdo_src=true so htat the original code would remain the same and I'll have an ospdo version of those tasks ?

jistr: Ok i see what you mean, we can perhaps add the variables then, but can we try using relative defaults so that the default behavior does not depend on providing absolute paths? I think one of the goals we should have in mind is keeping the configuration of the test suite as simple as possible -- not asking the user for configuring variables that can have sane defaults that "just work" in most test envs.

```
dpa_dir: "../.."
dpa_tests_dir: "{{ dpa_dir }}/tests"
```

pinikomarov: sure , sounds good, I'll make the adjustments

jistr (approved):
LGTM, thanks!

We should do a "squash and merge" on this one, to not have too many "fix" commits in git history.

jistr:
Thanks, together with #444 and predecessors this code has been on review for long enough so i'll merge :)

PR #691, opened 2024-10-28
ospdo adoption ovn adoption

pinikomarov:
Add Director Operator ovn_adoption scripts for adoption

Related CI patch and background:
https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/966

PRs submitted together :
[related PR list](https://issues.redhat.com/browse/OSPRH-2984?focusedId=25950569&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-25950569)

Jiras:

### DIFF
--- a/tests/roles/ovn_adoption/defaults/main.yaml
+++ b/tests/roles/ovn_adoption/defaults/main.yaml
@@ -33,3 +33,16 @@ ovn_nic_mapping_patch: |
         ovnController:
           nicMappings:
             baremetal: baremetal
+
+dpa_dir: "../.."
+dpa_tests_dir: "{{ dpa_dir }}/tests"
+
+# ospdo env:
+# Whether source env is OSPD Director Operator
+# i.e. https://github.com/openstack-k8s-operators/osp-director-operator
+ospdo_src: false
+# rhoso namespace
+rhoso_namespace: "openstack"
+# director operator namespace
+org_namespace: "ospdo_openstack"
+# adoption repo default location
--- a/tests/roles/ovn_adoption/defaults/main.yaml
+++ b/tests/roles/ovn_adoption/defaults/main.yaml
@@ -44,5 +44,5 @@ ospdo_src: false
 # rhoso namespace
 rhoso_namespace: "openstack"
 # director operator namespace
-org_namespace: "ospdo_openstack"
+org_namespace: "openstack"
 # adoption repo default location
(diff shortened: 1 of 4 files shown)


pinikomarov:
recheck

pinikomarov:
recheck

jistr (approved):
LGTM, also tagging @holser and @archana203 because the RHV based job is more sensitive to the cleanup logic being correct, but i think what we have here should work fine there too.

tests/roles/ovn_adoption/defaults/main.yaml
```diff
@@ -33,3 +33,16 @@ ovn_nic_mapping_patch: |
         ovnController:
           nicMappings:
             baremetal: baremetal
+
+dpa_dir: "../.."
```
archana203, line 37: Why these variables here?

pinikomarov: these are the default paths to the adoption repo dir as it is on the controller node in regular jobs
alternatively, in ospdo jobs the adoption repo dir  location would be different, and should be specified with this variable.
This is because of a fundamental difference in how the ospdo job is run , 
the zuul "controller" node is executing scripts on the BM titan node , so that dir locations are relative to the bmaremetal host the titan that's has the test env (odp with ospdo and rhoso) on it

pinikomarov:
@archana203 @holser Hi guys are there any more comments ?

tests/roles/ovn_adoption/defaults/main.yaml
archana203: These variables are not used in adoption role playbooks. I guess dpa_dir and tests_dir is used in zuul job playbooks. So, shouldn't it be defined as zuul variable? 
Any thoughts @holser @jistr

jistr: So these were introduced by Pini but i can share my view on things:

`dpa_tests_dir` is used for example here https://github.com/openstack-k8s-operators/data-plane-adoption/blob/df159026b647c743e5f55ca75879189419cd04d1/tests/roles/backend_services/tasks/ospdo_backend_services.yaml#L15

And `dpa_dir` isn't used directly but it allows you to just point it to the DPA repo root directory and then `dpa_tests_dir` will get automatically set to the correct path inside the repo.

The default values should work for all existing jobs, but Pini's environment has the DPA repo not on the Ansible controller so he needs more control over how the paths are resolved.

jistr:
I confirmed with Archana that we're ok to merge this and i'm merging.

PR #696, opened 2024-10-30
Support adoption for different cinder backends

katarimanojk:


tests/roles/cinder_adoption/defaults/cinder-volume-nfs-secrets.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
fmount, line 1: This should go into `files/` in the first place

PR #697, opened 2024-10-31
Enable dhcp agent notification

fyanac:
For testing purpose in adoption job in order to test dhcp agent

### DIFF
--- a/docs_user/modules/proc_adopting-the-networking-service.adoc
+++ b/docs_user/modules/proc_adopting-the-networking-service.adoc
@@ -50,6 +50,21 @@ spec:
 '
 ----
 
+* Note: If neutron-dhcp-agent was used in the OSP-17.1 deployment and should still be used after adoption, please enable also dhcp_agent_notification for neutron-api service.
+This can be done with patch:
++
+----
+$ oc patch openstackcontrolplane openstack --type=merge --patch '
+spec:
+  neutron:
+    template:
+      customServiceConfig: |
+        [DEFAULT]
+        dhcp_agent_notification = True
+'
+----
+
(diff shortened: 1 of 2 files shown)


slawqo:
This looks good for me and makes sense as by default this is disabled in the neutron-operator but in case of adoption we are testing adoption of the neutron-dhcp-agent too so we should have it enabled.

One additonal thing: maybe you can also add note in the docs: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/proc_adopting-the-networking-service.adoc about this. Something like:

> If neutron-dhcp-agent was used in the OSP-17.1 deployment and should still be used after adoption, please enable also dhcp_agent_notification for neutron-api service. This can be done with command:
...

docs_user/modules/proc_adopting-the-networking-service.adoc
```diff
@@ -50,6 +50,39 @@ spec:
...
+                metallb.universe.tf/address-pool: internalapi
+                metallb.universe.tf/allow-shared-ip: internalapi
+                metallb.universe.tf/loadBalancerIPs: 172.17.0.80
+            spec:
+              type: LoadBalancer
+      databaseInstance: openstack
+      databaseAccount: neutron
+      secret: osp-secret
+      networkAttachments:
+      - internalapi
```
slawqo, line 81: You can do here smaller patch:

```
$ oc patch openstackcontrolplane openstack --type=merge --patch '
spec:
  neutron:
    template:
      customServiceConfig: |
        [DEFAULT]
        dhcp_agent_notification = True
```

PR #698, opened 2024-10-31
doc maintenance updates

klgill:
Removed {rhos_prev_ver} attribute from links because they were not rendering correctly upstream.
Removed references to "Wallaby" upstream.

### DIFF
--- a/docs_user/adoption-attributes.adoc
+++ b/docs_user/adoption-attributes.adoc
@@ -12,7 +12,7 @@ ifeval::["{build}" != "downstream"]
 :OpenShiftShort: OCP
 :ocp_curr_ver: 4.16
 :rhos_curr_ver: Antelope
-:rhos_prev_ver: 
+:rhos_prev_ver:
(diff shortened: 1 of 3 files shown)


docs_user/adoption-attributes.adoc
```diff
@@ -12,7 +12,7 @@ ifeval::["{build}" != "downstream"]
 :OpenShiftShort: OCP
 :ocp_curr_ver: 4.16
 :rhos_curr_ver: Antelope
-:rhos_prev_ver: Wallaby
+:rhos_prev_ver:
```
ciecierski, line 15: Do you need this attribute if it is not used in assemblies?

klgill: @ciecierski Yes, I still need it because it is used downstream.

klgill: Will do!

jistr (approved):

PR #699, opened 2024-10-31
Basic ping test for test_with_ceph.

sathlan:


tests/playbooks/_stop_ping_test.yaml
```diff
@@ -0,0 +1,11 @@
...
+  hosts: localhost
+  gather_facts: false
+  module_defaults:
+    ansible.builtin.shell:
+      executable: /bin/bash
+  tasks:
+    - name: stop the ping test and validate the result
+      when: prelaunch_test_instance|bool
+      ansible.builtin.shell: |
+        ~/stop_ping.sh 0 0
```
jistr, line 11: I think we should parametrize the failure thresholds?

And maybe even parametrize whether we want to verify the ping test results at all? Essentially giving us a way to switch the ping test off if need be.

jistr: The test_with_ceph and test_minimal playbooks are the basis of all our CI jobs. The jobs don't redefine this playbook, they call it. So by adding it to test_with_ceph.yaml we're in fact adding it to several jobs both upstream and downstream. I think that's good but i think we should have the flag :).

Also if we won't be stopping the ping test, we shouldn't be starting it either i think.

-----

Later we should add it to test_minimal.yaml too but for now i'd leave it untouched. In case something goes wrong with the pingtest in periodic/downstream jobs, at least the non-ceph ones will be unaffected for now.

tests/roles/development_environment/defaults/main.yaml
```diff
@@ -1,5 +1,17 @@
 prelaunch_test_instance: true
 prelaunch_test_instance_script: pre_launch.bash
+# Ping test activation flag, note that it needs
+# prelaunch_test_instance as well.
+ping_test: true
```
jistr, line 5: I was wondering whether we want to default it to true or false, i guess either option is fine, just has different benefits and drawbacks.

tests/roles/development_environment/defaults/main.yaml
```diff
@@ -1,5 +1,17 @@
...
+# Relative to
+# ~/src/github.com/openstack-k8s-operators/data-plane-adoption
+# Collected in
+# /controller/data-plane-adoption-tests-repo/data-plane-adoption/
+ping_test_log_file: ./tests/logs/ping_results.log
+# max cut in seconds
+ping_test_loss_threshold: 0
+# max cut in percent
+ping_test_loss_threshold_percent: 0
```
jistr, line 14: The default thresholds will have to be higher :). Until 17.1.4 is released, we'll have significant ping loss due to https://issues.redhat.com/browse/OSPRH-10283

jistr (approved):
LGTM i think this will certainly fail and we'll have to adjust the thresholds but otherwise it looks good. Thanks!

tests/roles/development_environment/defaults/main.yaml
sathlan: With the current thresholds, it makes sense to keep them true by default, so we can collect logs.
In 18 update, we have it to false and only activated for the update job, so it doesn't affect other group jobs and they can activate it as they see fit.  Not sure if we affect other groups here.

jistr: Yea thats' a fair point, we do affect others & dev envs. So it probably does make sense to switch it to false here and propose companion patches to CI to enable it there.

tests/roles/development_environment/defaults/main.yaml
sathlan: Adjusted.  Now the threshold are high enought too pass the current failure.  What will happen though, is that a warning will be registered in the Ansible output and a log file with human readable datetime will be produced.

tests/roles/development_environment/defaults/main.yaml
sathlan: So with the current threshold nobody should have any error but everybody will have logs and if something insane like > 900 m cut and 50% loss, there a error will appear.

So another approach would be to lower the threshold in our job when the fix is there.  And then propose to other group to have lower threshold as well.

It seems to provide best of both world : everybody gets logs, nobody get failure until they opt-in for stringent threshold.

And if cut of 900m and 50% then ... well failure seems like a good idea.  Or we can maxout everything with 9000s and 100%, and re-inforce the "opt-in" idea.

But yeah, another approach to activate things that is less on/off type of things.

sathlan: Oki, that's convincing.  I'll set it to false then.  Should I keep those insane threshold or should we set them in the job as well, until the problem is resolved.  I think we should put the intend here (so 0 ping loss) and override it in the job. WDYT?

jistr: +1 on your suggestion. If we want the defaults here edited, we can do so easily later. But each job should probably be setting its expectation.

sathlan:
Note that before disabling the ping test both jobs were success:
https://softwarefactory-project.io/zuul/t/rdoproject.org/buildset/8957e051cd6847ba82ea1ef2f90a0b51

The job did a warning and created the human readable ping result:

https://logserver.rdoproject.org/99/699/6326b5690996b2b4167540c95e3dcafc1ef4b52e/github-check/adoption-standalone-to-crc-ceph/7b9f258/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/ping_results-hr.log

In https://logserver.rdoproject.org/99/699/6326b5690996b2b4167540c95e3dcafc1ef4b52e/github-check/adoption-standalone-to-crc-ceph/7b9f258/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/test_with_ceph_out_2024-11-07T08:53:28EST.log we can see the 'Stop ping' playbook:

```
LAY [Stop ping test] **********************************************************

TASK [Stop the ping and register the result.] **********************************
fatal: [localhost]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}
...ignoring

TASK [Outside thresholds ping result.] *****************************************
skipping: [localhost] => {"changed": false, "false_condition": "ping_result.rc not in [0, 64]", "skip_reason": "Conditional result was False"}

TASK [Inside thresholds ping result, but still some ping loss.] ****************
ok: [localhost] => {
    "msg": "WARNING: Ping test succeed, but some ping loss occured: 4707 packets transmitted, 4361 received, +285 errors, 7.35075% packet loss, time 4736141ms\nrtt min/avg/max/mdev = 0.580/1.437/101.357/3.147 ms, pipe 4"
}
```

In the last patchset I've changed the ignore error handling and the output, but tested that it behaved like in this test.

The default are now not to run the ping test and threshold are 0.

PR #701, opened 2024-11-04
Block storage: correct steps order

rajathere:
To deploy block storage service i.e. cinder, we should first remove
the service entries from old deployments to avoid any issues in
the new RHOSO deployment.

For that we need to rewrite the steps to:

1. Deploy cinder-api pod
2. List all scheduler and backup services from old deployment
3. Remove the old scheduler and backup services
4. Deploy scheduler, volume and backup services
5. Perform DB migration

This will ensure that our RHOSO deployment is fresh and does not
contain traces of errors from services of old deployment.

### DIFF
--- a/docs_user/modules/proc_adopting-the-block-storage-service.adoc
+++ b/docs_user/modules/proc_adopting-the-block-storage-service.adoc
@@ -21,7 +21,7 @@ $CONTROLLER1_SSH cat /var/lib/config-data/puppet-generated/cinder/etc/cinder/cin
 
 .Procedure
 
-. Create a new file, for example, `cinder.patch`, and apply the configuration:
+. Create a new file, for example, `cinder_api.patch`, and apply the configuration:
 +
 ----
 $ oc patch openstackcontrolplane openstack --type=merge --patch-file=<patch_name>
@@ -29,7 +29,7 @@ $ oc patch openstackcontrolplane openstack --type=merge --patch-file=<patch_name
 +
 * Replace `<patch_name>` with the name of your patch file.
 +
-The following example shows a `cinder.patch` file for an RBD deployment:
+The following example shows a `cinder_api.patch` file:
 +
 [source,yaml]
 ----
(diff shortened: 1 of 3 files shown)


jistr (approved):

rajathere:
recheck unrelated error while deploying standalone

    + ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i /home/zuul/.ssh/id_rsa root@192.168.122.100 'rm -f /tmp/repo-setup.sh'
    kex_exchange_identification: Connection closed by remote host
    Connection closed by 192.168.122.100 port 22

docs_user/modules/proc_adopting-the-block-storage-service.adoc
```diff
@@ -74,6 +74,61 @@ spec:
...
++
+----
+$ openstack volume service list
+
++------------------+------------------------+------+---------+-------+----------------------------+
+| Binary           | Host                   | Zone | Status  | State | Updated At                 |
++------------------+------------------------+------+---------+-------+----------------------------+
+| cinder-scheduler | standalone.localdomain | nova | enabled | down  | 2024-11-04T17:47:14.000000 |
+| cinder-backup    | standalone.localdomain | nova | enabled | down  | 2024-11-04T17:47:14.000000 |
+| cinder-volume    | hostgroup@tripleo_ceph | nova | enabled | down  | 2024-11-04T17:47:14.000000 |
```
katarimanojk, line 100: adoption code cleans up only scheduler and backup services which are down but not volume, shouldn't we cleanup this entry in db ?

rajathere: We should but what happens is, when we create the new cinder-volume service in RHOSO, we use the same hostname which re-uses the service record for cinder-volume.
In the documentation, we mention to clean up all the down services but in the ansible test job, we don't need to do it.
We can see that in effect when the job runs successfully.

tests/roles/cinder_adoption/tasks/main.yaml
```diff
@@ -1,8 +1,54 @@
...
+    {{ shell_header }}
+    {{ oc_header }}
+    oc wait pod --for condition=Ready -l component=cinder-api
+  register: cinder_api_running
+  until: cinder_api_running is success
+  retries: 60
+  delay: 2
+
+- name: Get cinder-backup down services
+  ansible.builtin.shell: |
```
katarimanojk, line 18: {{ shell_header }} is missing here and at L34
i see the ci [1] failed with this "ERROR! failed at splitting arguments, either an unbalanced jinja2 block or quotes: {{ oc_header }}"

[1] https://softwarefactory-project.io/zuul/t/rdoproject.org/build/ef8a4e5832b34336b15d2c047d1f84a9

docs_user/modules/proc_adopting-the-block-storage-service.adoc
katarimanojk: Thanks for the clarification.

PR #707, opened 2024-11-07
Split the networker configuration from the role

eshulman2:
added a var for running networker configuration to streamline with nodeset

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -151,89 +151,12 @@
     EOF
 
 - name: Create OpenStackDataPlaneNodeSet_networker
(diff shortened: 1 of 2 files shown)


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -483,7 +406,7 @@
     spec:
       nodeSets:
       - openstack
-      {%+ if edpm_nodes_networker is defined +%}
+      {%+ if edpm_nodes_networker is defined or edpm_networker_deploy +%}
```
jistr, line 409: I think we should have this in `defaults/main.yaml` defaulted to false?

eshulman2: added

jistr (approved):

PR #711, opened 2024-11-08
Setting skip_patching_ansibleee_csv to true

frenzyfriday:
Setting skip_patching_ansibleee_csv to true as edpm-ansible is stable and does not need to be patched frequently anymore.
This will stop patching the ansibleeImage and we will get the desired downstream version of openstack-ansibleee-runner image
Related to https://issues.redhat.com/browse/OSPRH-10704


karelyatin (commented):
This seems will break(patches will not get tested) adoption jobs running with edpm-ansible patches, can set skip_patching_ansibleee_csv=true only where needed in job itself.

frenzyfriday:
> This seems will break(patches will not get tested) adoption jobs running with edpm-ansible patches, can set skip_patching_ansibleee_csv=true only where needed in job itself.

Ack, makes sense. Closing this one. New patch: https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/1296

PR #714, opened 2024-11-09
Execute the cinder volume type command with sudo

archana203:
RHEV is using stack user to ssh the controller nodes. Adding sudo to execute the command for fetching the default volume type from cinder.conf.
Ref: [OSPRH-11363](https://issues.redhat.com/browse/OSPRH-11363)


archana203:
recheck

rajathere (approved):
LGTM

jistr (approved):
+1 we're already using sudo in other places that need root permissions, rather than assuming we're SSH-ing as root.

PR #735, opened 2024-11-14
Add nftables rules to the TripleO CephStorage nodes

fmount:
During the `Ceph` migration, `iptables` or `nftables` rules should be properly set. In `TripleO` (17.1.4) there's a switch from `iptables` to `nftables`, and this patch updates the role implementation to manage those rules as well.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/955

### DIFF
--- a/tests/roles/ceph_migrate/defaults/main.yml
+++ b/tests/roles/ceph_migrate/defaults/main.yml
@@ -30,10 +30,15 @@ ceph_timeout: 30
 # updated
 ceph_wait_mon_timeout: 10
 ceph_keep_mon_ipaddr: true
-ceph_firewall: false
-firewall_path:
+
+# firewall section
+ceph_firewall_enabled: false
+ceph_iptables_path:
   - "/etc/sysconfig/iptables"
   - "/etc/sysconfig/ip6tables"
+ceph_nftables_path:
+  - "/etc/nftables/tripleo-rules.nft"
+ceph_firewall_type: nftables
 
 # DEFAULT Ceph Reef container images
 ceph_haproxy_container_image: "quay.io/ceph/haproxy:2.3"
--- a/tests/roles/ceph_migrate/tasks/firewall.yaml
+++ b/tests/roles/ceph_migrate/tasks/firewall.yaml
@@ -10,29 +10,75 @@
     - iptables
     - nftables
 
-- name: Ceph Migration - Build the list of src and target nodes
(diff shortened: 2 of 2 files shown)


fmount:
Testing via: https://review.rdoproject.org/r/c/testproject/+/53696

tests/roles/ceph_migrate/defaults/main.yml
```diff
@@ -30,10 +30,15 @@ ceph_timeout: 30
...
-ceph_firewall: false
-firewall_path:
+
+# firewall section
+ceph_firewall_enabled: false
+ceph_iptables_path:
   - "/etc/sysconfig/iptables"
   - "/etc/sysconfig/ip6tables"
+ceph_nftables_path:
+  - "/etc/nftables/tripleo-rules.nft"
```
fmount, line 40: The current jobs might still have 17.1.3 content, with `iptables` instead of `nftables`, so we probably need to hold this for a bit.

fmount:
Actually worked as expected: https://logserver.rdoproject.org/96/53696/14/check/periodic-adoption-multinode-to-crc-ceph/033e618/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/test_ceph_migration_out_2024-11-14T09:23:18EST.log

tests/roles/ceph_migrate/defaults/main.yml
fmount: Adopted computes already have `nftables`.

tests/roles/ceph_migrate/tasks/firewall.yaml
```diff
@@ -10,29 +10,76 @@
...
+      delegate_to: "{{ node }}"
+      become: true
+      ansible.builtin.systemd:
+        name: iptables
+        state: restarted
+        enabled: true
+
+
+- name: Manage Ceph nftables rules
+  when: ceph_firewall_type == "iptables"
```
jistr, line 45: is it meant to say `when: ceph_firewall_type == "nftables"` here?

fmount: oh yes, good catch. I'll run testproject again with the updated version (even though by mistake I successfully executed both parts)

fmount:
(rebased)

fmount:
recheck

jistr (approved):

PR #742, opened 2024-11-22
[ovn_adoption] Add Ignore not found in cleanup handler

karelyatin:
During rollback role these are cleaned so ovn_adoption handler fails as the resources already cleaned up, adding ignore-not-found for these.


karelyatin:
Validated in https://softwarefactory-project.io/zuul/t/rdoproject.org/build/04dbc1e296a1462c830d708026fc7f37

jistr (approved):

PR #748, opened 2024-11-25
Update nodeset env variables 

fao89:


tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -312,11 +316,15 @@ networker_cr: |
...
+        value: "ansible.posix.profile_tasks,yaml"
       - name: ANSIBLE_FORCE_COLOR
         value: "True"
+      - name: ANSIBLE_DISPLAY_ARGS_TO_STDOUT
+        value: "True"
+      - name: ANSIBLE_SSH_ARGS
+        value: "-C -o ControlMaster=auto -o ControlPersist=80s"
       - name: ANSIBLE_VERBOSITY
-        value: "{{ dataplane_verbosity | default ('3') }}"
+        value: "{{ dataplane_verbosity | default ('2') }}"
```
rabi, line 327: Earlier verbosity 3 was too  noisy. Any reason we're changing default verbosity to 2 rather than ansible default 1?

fao89: no specific reason, but it would match what we have on CI for the other repos

jistr (approved):

tests/roles/dataplane_adoption/defaults/main.yaml
rabi: I think we should not change the default from 1 and also change in install_yamls that was changed recently it seems https://github.com/openstack-k8s-operators/install_yamls/pull/936/files#diff-3fcb2d93019c076d26f6d06ccd1b8d5ca9c3506914fb0a8bcd5ec26c54a30ae1R9

fao89: I've updated here and there: https://github.com/openstack-k8s-operators/install_yamls/pull/954

rabi (commented):
/lgtm

fao89:
recheck

PR #752, opened 2024-12-03
Fix ospDo documentation

bogdando:
Add missinig endif into the retrieving topology specific service config chapter. That shows the text remained hidden otherwise (the whole page conents, actually). Also add a note that only a single cell topology can be adopted from the source cloud managed with OSPDo.

### DIFF
--- a/docs_user/modules/proc_retrieving-topology-specific-service-configuration.adoc
+++ b/docs_user/modules/proc_retrieving-topology-specific-service-configuration.adoc
@@ -29,7 +29,10 @@ endif::[]
 ifeval::["{build}" == "downstream"]
 SOURCE_DB_ROOT_PASSWORD=$(cat ~/overcloud-deploy/overcloud/overcloud-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
 endif::[]
+endif::[]
+ifeval::["{OpenStackPreviousInstaller}" == "director_operator"]
 MARIADB_CLIENT_ANNOTATIONS='--annotations=k8s.v1.cni.cncf.io/networks=internalapi'
+endif::[]
 ----
 +
 To get the value to set `SOURCE_MARIADB_IP`, query the puppet-generated configurations in a Controller node:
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_retrieving-topology-specific-service-configuration.adoc
```diff
@@ -52,11 +52,13 @@ endif::[]
     mysql -rsh "$SOURCE_MARIADB_IP" -uroot -p"$SOURCE_DB_ROOT_PASSWORD" -e 'SHOW databases;')
 echo "$PULL_OPENSTACK_CONFIGURATION_DATABASES"
 ----
+endif::[]
```
bogdando, line 55: I am not really sure about the placement, but it surely is missing

bogdando: this probably needs to go some lines above, as we need to have MARIADB_CLIENT_ANNOTATIONS defined for ospdo case

bogdando:
PTAL

docs_user/modules/proc_retrieving-topology-specific-service-configuration.adoc
```diff
@@ -29,7 +29,10 @@ endif::[]
 ifeval::["{build}" == "downstream"]
 SOURCE_DB_ROOT_PASSWORD=$(cat ~/overcloud-deploy/overcloud/overcloud-passwords.yaml | grep ' MysqlRootPassword:' | awk -F ': ' '{ print $2; }')
 endif::[]
+endif::[]
+ifeval::["{OpenStackPreviousInstaller}" == "director_operator"]
 MARIADB_CLIENT_ANNOTATIONS='--annotations=k8s.v1.cni.cncf.io/networks=internalapi'
```
jistr, line 34: I think those annotations are for non-ospdo as well. (?)

bogdando: it doesn't look like so? see the password etc evaluated below

PR #756, opened 2024-12-04
[DNM] Test multiple replicas memcached

karelyatin:
Locally seeing issues when running with multiple
replicas with ipv6 adoption, testing it in CI for ipv4 to see if it reproduces.


karelyatin:
Just for test, and it concluded issue was specific to ipv6 + no tls + dogpile.cache.memcached cache backend

PR #761, opened 2024-12-10
Create separate VM/FIP for ping test

karelyatin:
Ping test is using the same vm used during nova
vm stop/start verification and that makes the ping test to fail.
We should not use the same vm for ping test.
With this patch we create a dedicated vm/fip
when ping test is enabled.

Related-Issue: #OSPRH-12246


jistr (approved):

jistr:
Thank you for diving into this :)

PR #763, opened 2024-12-12
Add Barbican to OSP17.1 scenarios tested for adoption

ciecierski:
Add Barbican to all already existing OSP17.1 adoption scenarios. 
Barbican needs to be enabled by default in any future scenario.

### DIFF
--- a/scenarios/hci.yaml
+++ b/scenarios/hci.yaml
@@ -55,6 +55,8 @@ stacks:
       - "/usr/share/openstack-tripleo-heat-templates/environments/enable-legacy-telemetry.yaml"
       - "/usr/share/openstack-tripleo-heat-templates/environments/cephadm/cephadm.yaml"
       - "/usr/share/openstack-tripleo-heat-templates/environments/cephadm/ceph-mds.yaml"
+      - "/usr/share/openstack-tripleo-heat-templates/environments/services/barbican.yaml"
+      - "/usr/share/openstack-tripleo-heat-templates/environments/barbican-backend-simple-crypto.yaml"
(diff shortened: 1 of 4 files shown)


jistr (approved):

xek:
Are there still adoption jobs with Barbican being disabled on 17.1? We need to test both scenarios, since Barbican service is enabled by default in 18 but disabled by default on 17.1 greenfield deployments.

ciecierski:
> Are there still adoption jobs with Barbican being disabled on 17.1? We need to test both scenarios, since Barbican service is enabled by default in 18 but disabled by default on 17.1 greenfield deployments.

Good point. I thought since Barbican is enabled by default in 18 we need to have it always enabled in source 17.1. The fact it is disabled in 17.1 by default is important. Since we have alpha and gamma for now, maybe we can have enabled in one and disabled in other. For the rest uni jobs we can spread it even. Wdyt? @jistr @xek

jistr:
For the uni jobs specifically, i think we should match the desired uni topology whenever possible. So if greenfield uni-alpha and hci do include Barbican, we should include it in the adoption uni jobs as well i'd say.

@hjensas are you ok with the uni-alpha edit here?

hjensas:
> For the uni jobs specifically, i think we should match the desired uni topology whenever possible. So if greenfield uni-alpha and hci do include Barbican, we should include it in the adoption uni jobs as well i'd say.
> 
> @hjensas are you ok with the uni-alpha edit here?

Agree with @jistr - it would make sense to match the uni topology.

I have no objection to include Barbican in alpha.

PR #768, opened 2024-12-17
support cinder netapp nfs backend adoption

katarimanojk:
Add support cinder netapp nfs backend adoption

### DIFF
--- a/tests/roles/cinder_adoption/files/cinder-volume-ontap-secrets.yaml
+++ b/tests/roles/cinder_adoption/files/cinder-volume-ontap-secrets.yaml
@@ -1,0 +1,19 @@
+# Define the "cinder-volume-ontap-secrets" Secret that contains sensitive
+# information pertaining to the [ontap] backend.
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    service: cinder
+    component: cinder-volume
+  name: cinder-volume-ontap-secrets
+type: Opaque
+stringData:
+  ontap-cinder-secrets: |
+    [ontap-nfs]
+    netapp_login=vsadmin
+    netapp_password=qum5net
+    netapp_vserver=ntap-rhv-dev-rhos
+    nas_host=10.46.29.88
+    nas_share_path=/cinder_nfs
+    netapp_pool_name_search_pattern=(cinder_volumes)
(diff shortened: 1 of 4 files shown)


tests/roles/cinder_adoption/files/cinder-volume-ontap-secrets.yaml
```diff
@@ -0,0 +1,18 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    service: cinder
+    component: cinder-volume
+  name: cinder-volume-ontap-secrets
+type: Opaque
+stringData:
+  ontap-cinder-secrets: |
```
tosky, line 12: instead of defining the secret here, it may make sense to extract the required information from the OSP deployment. Barbican did this for example:
https://github.com/openstack-k8s-operators/data-plane-adoption/blob/ef823f2f2c5772fce9e7f04876e0bdfe049deb3a/tests/roles/barbican_adoption/tasks/main.yaml#L6

tests/roles/cinder_adoption/tasks/volume_backend.yaml
```diff
@@ -1,6 +1,21 @@
-- name: deploy podified Cinder volume
+- name: deploy podified Cinder volume with ceph
   when: cinder_volume_backend == 'ceph'
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
     oc patch openstackcontrolplane openstack --type=merge --patch '{{ cinder_volume_backend_patch }}'
+
+- name: deploy podified Cinder volume with netapp NFS
+  when: cinder_volume_backend == 'ontap'
```
tosky, line 9: ontap or ontap-nfs ? ontap could provide other protocols too

tosky (changes requested):

tests/roles/cinder_adoption/tasks/volume_backend.yaml
katarimanojk: agreed, changed it as ontap-nfs.

tests/roles/cinder_adoption/files/cinder-volume-ontap-secrets.yaml
katarimanojk: @tosky yeah, i think it can be extracted from OSP deployment , first we need to look at how the secrets section is populated in  /var/lib/config-data/puppet-generated/cinder/etc/cinder/cinder.conf.

katarimanojk:
This patch is no longer needed

PR #779, opened 2025-01-08
WIP: Add customization on netconfig_network

averdagu:
dataplane_adoption role doesn't allow customization with the subnet ranges, this conflicts with testing scenario 1 (using different subnet ranges) during adoption.

Related: OSPRH-5881


tests/roles/glance_adoption/files/glance_local.yaml.j2
```diff
@@ -24,7 +24,7 @@ spec:
                   annotations:
                     metallb.universe.tf/address-pool: internalapi
                     metallb.universe.tf/allow-shared-ip: internalapi
```
karelyatin, line 26: would we better to move these files under templates directory instead now.

averdagu: Done

tests/roles/backend_services/templates/openstack_control_plane.j2
```diff
@@ -126,7 +126,7 @@ spec:
             metadata:
               annotations:
                 metallb.universe.tf/address-pool: internalapi
-                metallb.universe.tf/loadBalancerIPs: 172.17.0.85
+                metallb.universe.tf/loadBalancerIPs: {{ internalapi_prefix | default('172.17.0') }}.85
```
ciecierski, line 129: Do we want to enable to customize just prefix or the whole ip address? The second option for me is more flexible, but maybe you had some other assumptions in mind.

averdagu: Well, the scenario that I'm trying to achieve (adoption where tripleo and podified had different subnets) doesn't require to change the host_id, only the network_id. So I didn't want to change things that were not necessary for the scenario

ciecierski: Ok, makes sense. Thank you

PR #782, opened 2025-01-10
Add static routes to netConfig to support scenario A

averdagu:
Adopting a workload using scenario A (different subnet ranges between wallaby CP and next-gen CP) wont succeed if netConfig doesn't have the routes to reach old CP.

Relates: [OSPRH-7544](https://issues.redhat.com//browse/OSPRH-7544)

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -35,6 +35,11 @@
     file: ospdo.yaml
   when: ospdo_src| bool
 
+- name: Override netconfig_networks if netconfig_networks_override is defined
+  ansible.builtin.set_fact:
+    netconfig_networks: "{{ netconfig_networks_override }}"
+  when: netconfig_networks_override is defined
+
 - name: ensure IPAM is configured
   no_log: "{{ use_no_log }}"
   ansible.builtin.shell: |


averdagu:
recheck

tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -24,6 +24,9 @@ netconfig_networks:
             start: 172.17.0.100
         cidr: 172.17.0.0/24
         vlan: 20
+        routes:
```
karelyatin, line 27: why we need to change defaults? should we apply this only for the envs where we need to modify it?

averdagu: I thought that this won't affect on the default scenario, as this route is already added by system (since the interface belongs already to 172.17.0.0/24) and I didn't want to overengineer it). But I could modify it to only modify it in case it's the scenario A (though this may need some extra variable from install_yamls)

karelyatin: The idea is we shouldn't change default, and add the required routes where needed, both in jobs and in documentation

averdagu: After trying to modify the netconfig_network in flight during main execution I thought of a different approach where you can define the desired netconfig_network on your job definition (using netconfig_network_override) and if the override exist will use that, if not it will work as normal

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -35,6 +35,11 @@
     file: ospdo.yaml
   when: ospdo_src| bool
 
+- name: Override netconfig_networks if netconfig_networks_override is defined
+  ansible.builtin.set_fact:
+    netconfig_networks: "{{ netconfig_networks_override }}"
```
ciecierski, line 40: I think the same can be achived by setting netconfig_networks: {{ netconfig_networks_override }} in job definition in vars, without setting additional fact.

PR #793, opened 2025-01-24
added optional step to power off controllers

klgill:
https://issues.redhat.com/browse/RHOSPDOC-1908


klgill:
Hi @jistr. Are we specifically calling out OVN, Swift, and Ceph because those services would not be migrated off of the Controllers post-adoption? So basically, if a customer has any of those 3 services, they cannot migrate them, and as a result, they cannot power off the Controllers.

jistr:
@klgill They can be migrated off but it doesn't happen "by default" just by completing the adoption of the data plane. There are separate procedures for migrating these services off, so i think it's worth calling these out as it might result in an "aha" moment for the admin who might be about to power off their controllers.

jistr (approved):

PR #798, opened 2025-01-31
Change status of pvs from Released to Available

holser:
During clean up we should change status to available and rely on reclaim policies of pvs. If pv policy is Delete so all data will be deleted. This helps not to produces tons of "Released" pvs on recurrent reinstallation processes of RHOSO

Related-Bug: [OSPRH-13619](https://issues.redhat.com//browse/OSPRH-13619)


bogdando (approved):
lgtm

ciecierski:
recheck

PR #804, opened 2025-02-05
Skip IPAM configuration if it is already configured

ciecierski:
Run ensure IPAM is configured task only when ipam is not configured, otherwise skip re-applying netconfig to not overwrite it with default values. 

Default ipam_configured bool value is false.

Resolves: OSPRH-13659


tests/roles/dataplane_adoption/defaults/main.yaml
```diff
@@ -3,6 +3,7 @@
 supported_backup_backends: []
 supported_volume_backends: []
 edpm_networker_deploy: false
+ipam_configured: false
```
jistr, line 6: Nit: maybe `configure_ipam` would be a better name? Generally i think we've been naming the variables in the convention of "do something: true/false" rather than "this thing was already done so skip it: true/false".

ciecierski: done

PR #806, opened 2025-02-06
Run the online data migrations from cinder-api

holser:
We currently have a mix of using cinder-api and cinder-scheduler pods/containers to run cinder-manage commands whereas it should be better to be consistent.
This patch updates the online data migrations task to be run by cinder-api.

(cherry picked from commit 39e5aa9bf472a17f253043a189b55ece6d01eab2)


ciecierski (changes requested):
-1 for change
This change is part of cherry-pick proposed in https://github.com/openstack-k8s-operators/data-plane-adoption/pull/807

PR #808, opened 2025-02-10
Seperate control_plane_rollback role calling

archana203:
Ref: [OSPRH-12711](https://issues.redhat.com//browse/OSPRH-12711)


ciecierski (changes requested):
There is problem with test-rollback-minimal and test-rollback-with-ceph running both adoption test and rollback in the same Make run.

archana203:
> There is still problem with test-rollback-minimal and test-rollback-with-ceph running both adoption test and rollback in the same Make run.
test-rollback-minimal is used in the periodic rollback jobs where the rollback happens after controlplane adoption. We want to separate it just for the development purpose.

archana203:
recheck

ciecierski (approved):

archana203:
recheck

PR #809, opened 2025-02-11
Neutron-sriov-nic-agent agent enablement on edpm is optional and disabled

ciecierski:
edpm_neutron_sriov_agent_enabled should be set to false by default. It is optional step in documentation and should only required when agents are running on OSP deployment.


ciecierski:
We have dummy-dev for sriov in standalone for very basic tests and this is the reason for having edpm_neutron_sriov_agent_enabled set to true.

ciecierski:
edpm_neutron_sriov_agent_enabled can be set to false in jobs that are not testing sriov and no change is required. Closing pull request.

PR #746, opened 2024-11-25
Multi cell database/mq adoption

bogdando:
1. Fix OSPDo specifics for single vs multi cellls

Declare RUN_OVERRIDES before it is used.
Use env vars instead of docs generation conditions to reuse the same
code in tests:
* Add MARIADB_RUN_OVERRIDES to cover all overrides and client annotations
* Add missing definitions for rhoso/ospd namespace specific vars
* Use env TRIPLEO_PASSWORDS for all cases as OSPDo still deploys
	tripleo
* Define and use NAMESPACE (default openstack) instead of
	RHOSO18_NAMESPACE or OSPDO_NAMESPACE. Remove unused rhoso18 ns value
	(only in these guide).

2. Refactor comments in commands into asciidocs native
    
Illustrate how commands in scripts could have comments
that become (almost as is) native ascii docs foot-notes.
   
When copying code into docs, the minimal adjustments will
be needed, like adding '$' prefix (or '>' for multiline commands).

Provide a static multi-cell config for databases and messaging
for adoption guide and tests, which comprises a 3 cells.

3. Keep renaming 'default' cell consistent for single and multi cells:

Default becomes cellX (or it can be imported as is, for a multi-cell
case only)
cell1 becomes mapped to openstack-cell1 osdp node set
cell2 becomes mapped to openstack-cell2 osdp node set, etc.
cellX (X=3 here) becomes mapped to openstack-cell3. Alternatively,
default cell retains its name for the openstack-default osdpns
mapping
Evaluate podified MariaDB passwords for cells from osp-secret
to align the tests with documented commands. Remove no longer
needed podified DB password variable.

4. Make ansible and shell variables compute cells aware.

5. Rework vars and secrets YAML values for the source and edpm
nodes to not confuse its different naming schemes for cells
in OSP/TripleO and RHOSO.

6. Remove cached fact for pulled OSP configuration as it can no longer
be generated in a multi-cell setup, where related shell variables
become bash arrays.

7. Simplify ENV headers management by collecting in a single place.

8. Adjust storage/storageRequests values to make it better fitting
a multi-cell test scenarios. Also provide values in docs and
add a comment to adjust them as needed.

9. Remove source_db_root_password as it is directly evaluated from
tripleo passwords into an env var.
    
10. Run mysql commands in individual pods.
Finished pods take time to terminate, avoid errors where
consequent mysql commands failing because the old and new pod use the
same name.
    
11. Rename nodesets to openstack-cell1, which is needed for adoption of
[...]

PR #812, opened 2025-02-14
Set canonical hostname to example.com instead of localdomain

ciecierski:
canonical hostname of cloud should match hostname defined in architecture repo[1]

[1]https://github.com/openstack-k8s-operators/architecture/blob/743a8bfb71da2aa5245fda5fe73f2463fb1d0133/examples/va/hci/control-plane/nncp/values.yaml#L32


jistr:
I would like to hold off all merges until we either merge #746 (or determine that it needs another iteration). We can remove the hold afterwards.

PR #816, opened 2025-02-19
Create libvirt-secret to when TLS-e is enabled

holser:
When TLS-e is on libvirt requires authentication. This patch describes how to extract LIBVIRT_PASSWORD from file and create a secret with the same password on EDPM node when TLS-e is enabled

Closes-Bug: [OSPRH-13479](https://issues.redhat.com//browse/OSPRH-13479)

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -148,6 +148,30 @@ oc get secret nova-migration-ssh-key || oc create secret generic nova-migration-
 rm -f id*
 cd -
 ----
++
+
+. If TLS Everywhere is enabled, set LIBVIRT_PASSWORD to match the existing {OpenStackShort} deployment password:
++
+----
+declare -A TRIPLEO_PASSWORDS
+TRIPLEO_PASSWORDS[default]="$HOME/overcloud-passwords.yaml"
+LIBVIRT_PASSWORD=$(cat ${TRIPLEO_PASSWORDS[default]} | grep ' LibvirtTLSPassword:' | awk -F ': ' '{ print $2; }')
+LIBVIRT_PASSWORD_BASE64=$(echo -n "$LIBVIRT_PASSWORD" | base64)
+----
+* Create libvirt-secret when TLS-e is enabled.
+[source,yaml]
+----
+$ oc apply -f - <<EOF
+apiVersion: v1
+kind: Secret
+metadata:
+  name: libvirt-secret
+  namespace: openstack
+type: Opaque
+data:
+  LibvirtPassword: ${LIBVIRT_PASSWORD_BASE64}
+EOF
+----
 
 . If you use a local storage back end for libvirt, create a `nova-compute-extra-config` service to remove pre-fast-forward workarounds and configure Compute services to use a local storage back end:
 +
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -92,6 +92,25 @@
     rm -f id*
     cd -
 
+- name: create libvirt-secret
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc apply -f - <<EOF
+    apiVersion: v1
+    kind: Secret
+    metadata:
+        name: libvirt-secret
+        namespace: {{ rhoso_namespace }}
+    type: Opaque
+    data:
+      LibvirtPassword: {{ libvirt_password | b64encode }}
+    EOF
+  when:
+    - not ospdo_src| bool
+    - libvirt_password | length > 0
+
 - name: create a Nova Compute Extra Config service (no ceph backend in use)
   when:
     - compute_adoption|bool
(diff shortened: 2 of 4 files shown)


docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -148,6 +148,29 @@ oc get secret nova-migration-ssh-key || oc create secret generic nova-migration-
...
+----
+$ oc apply -f - <<EOF
+apiVersion: v1
+kind: Secret
+metadata:
+  name: libvirt-secret
+  namespace: openstack
+type: Opaque
+data:
+  LibvirtPassword: ${LIBVIRT_PASSWORD}
```
jistr, line 171: This needs to be base64 encoded. Will link an internal thread on Slack.

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -148,6 +148,29 @@ oc get secret nova-migration-ssh-key || oc create secret generic nova-migration-
 rm -f id*
 cd -
 ----
++
+
+. Set LIBVIRT_PASSWORD to match the existing {OpenStackShort} deployment password:
```
bogdando, line 153: This should be given a condition, like https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc?plain=1#L173

jistr (changes requested):
(review inline)

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
jistr: Did you test the code you are proposing -- does it result in the correct password being written to the compute nodes config?

All available info indicates that if we provide just `LibvirtPassword: ${LIBVIRT_PASSWORD}`, OpenShift will interpret that value as base64 whether we like it or not. So it will try to run base64 decoding operation on that, and then i'd expect it to either fail, or if the password happens to be a valid base64 by chance, it would get mangled into something incorrect.

Please see the conversation i linked on Slack.

jistr: This also help to see how the secrets work: https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html-single/deploying_red_hat_openstack_services_on_openshift/index#proc_creating-the-data-plane-secrets_dataplane

holser: There is a confusion between `oc set data` and plain `oc apply -f yaml`. You are correct if we apply yaml we need to base64 the password. I've updated PR. Thank you

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -148,6 +148,29 @@ oc get secret nova-migration-ssh-key || oc create secret generic nova-migration-
...
 cd -
 ----
++
+
+. If TLS Everywhere is enabled, set LIBVIRT_PASSWORD to match the existing {OpenStackShort} deployment password:
++
+----
+declare -A TRIPLEO_PASSWORDS
+TRIPLEO_PASSWORDS[default]="$HOME/overcloud-passwords.yaml"
+LIBVIRT_PASSWORD=$(cat ${TRIPLEO_PASSWORDS[default]} | grep ' LibvirtTLSPassword:' | awk -F ': ' '{ print $2; }')
```
jistr, line 158: The tests look fine to me but we'll still need to fix the docs. Rather than trying to do the encoding inside `oc apply` and deal with quote escaping, perhaps we could define another variable, something like:

```
LIBVIRT_PASSWORD_BASE64=$(echo -n "$LIBVIRT_PASSWORD" | base64)
```

And then use that variable in the `oc apply`? Just a suggestion.

jistr:
(The CI failure is relevant.)

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
holser: Fixed latest commit

jistr (approved):

holser:
recheck

tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -92,6 +92,25 @@
...
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc apply -f - <<EOF
+    apiVersion: v1
+    kind: Secret
+    metadata:
+        name: libvirt-secret
+        namespace: openstack
```
ciecierski, line 105:
```suggestion
        namespace: [openstack](namespace:  {{ rhoso_namespace }})
```

ciecierski (changes requested):
Please check my inline comments

tests/roles/dataplane_adoption/tasks/main.yaml
holser: thanks. Fixed in next PR update

tests/secrets.sample.yaml
```diff
@@ -29,6 +29,7 @@ nova_password: "{{ lookup('file', tripleo_passwords['default']) | from_yaml | co
 octavia_password: "{{ lookup('file', tripleo_passwords['default']) | from_yaml | community.general.json_query('*.OctaviaPassword') | first }}"
 placement_password: "{{ lookup('file', tripleo_passwords['default']) | from_yaml | community.general.json_query('*.PlacementPassword') | first }}"
 swift_password: "{{ lookup('file', tripleo_passwords['default']) | from_yaml | community.general.json_query('*.SwiftPassword') | first }}"
+libvirt_password: "{{ lookup('file', tripleo_passwords['default']) | from_yaml | community.general.json_query('*.LibvirtTLSPassword') | first }}"
```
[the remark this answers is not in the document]
jistr: +1 this should be enabled in CI but perhaps it doesn't need to be a prerequisite to merging this? We're mainly after getting the doc fix merged soon i'd say.

ciecierski (approved):
inline comment for variable namespace addressed
test-project with activation of secret creation in progress

PR #827, opened 2025-02-26
Remove hardcoded namespaces

holser:
Set rhoso_namespace variable instead of hardcoded namespace


fultonj:
I do not object

fmount (approved):
/lgtm

PR #828, opened 2025-02-27
Don't cleanup interfaces/routes if not required

rabi:
If os-net-config is not configuring all interfaces/default route don't cleanup interfaces/devices not in network config. We're changing the default so that we cleanup non network-scripts created devices which causes issues.

jira: https://issues.redhat.com/browse/OSPRH-13433


jistr:
Related PRs:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/865/files
https://github.com/openstack-k8s-operators/edpm-ansible/pull/882/files

/lgtm

rabi:
recheck

fao89 (approved):

bshephar (approved):

PR #831, opened 2025-02-28
Fix select DB to omit deleted hosts

bogdando:
Jira: [OSPRH-13482](https://issues.redhat.com//browse/OSPRH-13482)

### DIFF
--- a/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
+++ b/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
@@ -349,7 +349,7 @@ $ for CELL in $(echo $RENAMED_CELLS); do <3>
 >   c1dbs=$(oc exec openstack-$CELL-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p${PODIFIED_DB_ROOT_PASSWORD[$CELL]} -e 'SHOW databases;') <4>
 >   echo $c1dbs | grep -Eq "\bnova_${CELL}\b" && echo "OK" || echo "CHECK FAILED"
 >   novadb_svc_records=$(oc exec openstack-$CELL-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p${PODIFIED_DB_ROOT_PASSWORD[$CELL]} \
->     nova_$CELL -e "select host from services where services.binary='nova-compute' order by host asc;")
+>     nova_$CELL -e "select host from services where services.binary='nova-compute' and deleted=0 order by host asc;")
 >   diff -Z <(echo "x$novadb_svc_records") <(echo "x${PULL_OPENSTACK_CONFIGURATION_NOVA_COMPUTE_HOSTNAMES[@]}") && echo "OK" || echo "CHECK FAILED" <5>
 > done
 ----
--- a/tests/roles/dataplane_adoption/tasks/nova_ffu.yaml
+++ b/tests/roles/dataplane_adoption/tasks/nova_ffu.yaml
@@ -3,7 +3,7 @@
     {{ mariadb_copy_shell_vars_dst }}
     for CELL in $(echo $RENAMED_CELLS); do
       oc exec openstack-$CELL-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p"${PODIFIED_DB_ROOT_PASSWORD[$CELL]}" \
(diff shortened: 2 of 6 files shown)


docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
```diff
@@ -327,7 +327,7 @@ $ echo $dbs | grep -Eq '\bkeystone\b' && echo "OK" || echo "CHECK FAILED"
 $ echo $dbs | grep -Eq '\bneutron\b' && echo "OK" || echo "CHECK FAILED"
 $ echo "${PULL_OPENSTACK_CONFIGURATION_DATABASES[@]}" | grep -Eq '\bovs_neutron\b' && echo "OK" || echo "CHECK FAILED" <1>
 $ novadb_mapped_cells=$(oc exec openstack-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p"${PODIFIED_DB_ROOT_PASSWORD['super']}" \
->   nova_api -e 'select uuid,name,transport_url,database_connection,disabled from cell_mappings;') <2>
+>   nova_api -e 'select uuid,name,transport_url,database_connection,disabled from cell_mappings where disabled=0;') <2>
```
gibizer, line 330: I think this is incorrect. We need to adopt disable computes as they can be enabled later. What the original bug report saw is that we are trying to adopt deleted computes. And I agree there, we should not try to adopt deleted computes.

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
```diff
@@ -349,7 +349,7 @@ $ for CELL in $(echo $RENAMED_CELLS); do <3>
 >   c1dbs=$(oc exec openstack-$CELL-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p${PODIFIED_DB_ROOT_PASSWORD[$CELL]} -e 'SHOW databases;') <4>
 >   echo $c1dbs | grep -Eq "\bnova_${CELL}\b" && echo "OK" || echo "CHECK FAILED"
 >   novadb_svc_records=$(oc exec openstack-$CELL-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p${PODIFIED_DB_ROOT_PASSWORD[$CELL]} \
->     nova_$CELL -e "select host from services where services.binary='nova-compute' order by host asc;")
+>     nova_$CELL -e "select host from services where services.binary='nova-compute' and disabled=0 and deleted=0 order by host asc;")
```
gibizer, line 352: ditto here and blow. Disabled but not deleted computes needs to be adopted. But deleted computes (regardless for disabled or not) should not be adopted.

gibizer (changes requested):

gibizer (commented):
How do we plan to test this? Only manually? Or we want to add a compute to the 16.2 side and then delete it before the adoption then after the adoption we assert that the deleted compute is not adopted?

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
bogdando: that makes sense

bogdando:
> How do we plan to test this? Only manually? Or we want to add a compute to the 16.2 side and then delete it before the adoption then after the adoption we assert that the deleted compute is not adopted?

created https://issues.redhat.com/browse/OSPRH-14422 to figure that

gibizer (changes requested):

tests/roles/dataplane_adoption/tasks/nova_ffu.yaml
```diff
@@ -3,7 +3,7 @@
     {{ mariadb_copy_shell_vars_dst }}
     for CELL in $(echo $RENAMED_CELLS); do
       oc exec openstack-$CELL-galera-0 -n $NAMESPACE -c galera -- mysql -rs -uroot -p"${PODIFIED_DB_ROOT_PASSWORD[$CELL]}" \
-        -e "select a.version from nova_${CELL}.services a join nova_${CELL}.services b where a.version!=b.version and a.binary='nova-compute';"
+        -e "select a.version from nova_${CELL}.services a join nova_${CELL}.services b where a.version!=b.version and a.binary='nova-compute' and a.disabled=0;"
```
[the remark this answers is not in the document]
bogdando: well spotted, thanks!

nope, no need for a ``b.deleted`` test

holser (approved):
/LGTM

holser:
approve

PR #832, opened 2025-03-05
Fix Swift ring update race condition

cschwede:
The adoption role checks if the requested storage pods are up and running, but not if the rings have been updated yet and include the newly added pvs.

PVs get added by the swift-ring-rebalance job after the storage pods are up and running. However, at the same time the adoption playbook continues, also modifying the rings - which might include the new pvs or not. Changes might also be overwritten either by the rebalance job or the adoption task, resulting in a ring that uses the wrong device weights.

At the end old devices won't be drained properly.

This patch adds a check to ensure the new pvs have been added before continuing, ensuring that the rings include the expected pvs.

Jira: [OSPCIX-711](https://issues.redhat.com//browse/OSPCIX-711)


karelyatin:
recheck

karelyatin:
/lgtm
Tested in https://review.rdoproject.org/r/c/testproject/+/55541 - 4 green runs

PR #833, opened 2025-03-05
Update ceph dashboard port

katarimanojk:
This patch updates the dashboard port configured for ceph-mgr and adds a note about it.
JIRA: https://issues.redhat.com/browse/OSPRH-14504

### DIFF
--- a/docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
+++ b/docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
@@ -28,12 +28,24 @@ $ sudo iptables-save
 $ sudo systemctl restart iptables
 ----
 +
+[NOTE]
+The default dashboard port for `ceph-mgr` in a greenfield deployment is 8443. With director-deployed {Ceph}, the default port is 8444 because the service ran on the Controller node, and it was necessary to use this port to avoid a conflict. For adoption, update the dashboard port to 8443 in the `ceph-mgr` configuration and firewall rules.
+
+. Log in to `controller-0` and update the dashboard port in the `ceph-mgr` configuration to 8443:
++
+----
+$ sudo cephadm shell
+$ ceph config set mgr mgr/dashboard/server_port 8443
+$ ceph config set mgr mgr/dashboard/ssl_server_port 8443
+$ ceph mgr module disable dashboard
+$ ceph mgr module enable dashboard
+----
 . If `nftables` is used in the existing deployment, edit `/etc/nftables/tripleo-rules.nft`
 and add the following content:
 +
 ----
-# 113 ceph_mgr {'dport': ['6800-7300', 8444]}
-add rule inet filter TRIPLEO_INPUT tcp dport { 6800-7300,8444 } ct state new counter accept comment "113 ceph_mgr"
+# 113 ceph_mgr {'dport': ['6800-7300', 8443]}
+add rule inet filter TRIPLEO_INPUT tcp dport { 6800-7300,8443 } ct state new counter accept comment "113 ceph_mgr"
 ----
 
 . Save the file.


docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
```diff
@@ -32,9 +32,11 @@ $ sudo systemctl restart iptables
 and add the following content:
 +
 ----
-# 113 ceph_mgr {'dport': ['6800-7300', 8444]}
-add rule inet filter TRIPLEO_INPUT tcp dport { 6800-7300,8444 } ct state new counter accept comment "113 ceph_mgr"
+# 113 ceph_mgr {'dport': ['6800-7300', 8443]}
+add rule inet filter TRIPLEO_INPUT tcp dport { 6800-7300,8443 } ct state new counter accept comment "113 ceph_mgr"
```
fmount, line 36: -1: the `Ceph dashboard`, for a `Director` deployed `Ceph`, listens on `8444`. As per our conversation, and because 18 greenfield uses `8443`, we can add `8443` to make sure we're not going to have firewall issues, but if we remove `8444` here, then we need to `reconfigure` the Ceph dashboard to change port, and we shouldn't do that.

My proposal here is to have an additional `8443` to the port list, but not removing `8444`, which has reconfiguration implications.

fmount (changes requested):

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
```diff
@@ -32,9 +32,11 @@ $ sudo systemctl restart iptables
...
 ----
+[NOTE]
+Default dashboard port for ceph-mgr in greenfield deployment is 8443, whereas the adoption from director deployed ceph needs the port to be changed from 8444 to 8443.
```
fmount, line 39: -1: we do not want to change port for an adopted environment. If that is the idea we should:

1. Document the steps required to reconfigure the dashboard on 8443 (remove the existing `ceph mgr config set ...`, disable and enable the ceph dashboard module) 
2. Update the automation (test suite) to test this scenario during the migration to validate this doc update.

The alternative here is to document that for pre-existing environments are keeping the same port, but 8443 is added because of greenfield.

fmount (commented):
After thinking more, here's a general comment about this patch.
If we decide to take this path (replace 8444 with 8443), we need to add steps to reconfigure the Ceph cluster to have the dashboard listening on 8443, and add a validation similar to what we've done in ci-framework [1].
Because we probably land in a topology (unigamma) where tls is enabled by default, if TripleO was deployed with tls-everywhere, we need to evaluate any implication around the configuration [2].
I think [2] can be simply part of a dedicated follow up, but for now, if we decide to change the Ceph Dashboard port,   think we should update the Ceph migration role to handle this part, as well as adding more wording about the reason behind this choice (L39).


[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/dashboard/dashboard.yml
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/dashboard/dashboard.yml#L31

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
fultonj: I think we should change this:

> Default dashboard port for ceph-mgr in greenfield deployment is 8443, whereas the adoption from director deployed ceph needs the port to be changed from 8444 to 8443.

To this:

Default dashboard port for ceph-mgr in greenfield deployment is 8443. With director-deployed Ceph, the default port is 8444 because the service ran on the controller node and it was necessary to use this port in order to avoid a conflict.

We can make this change without deciding which option (1 or 2) to use as described above. My aim here was to capture this history in the doc since someone might ask and we can have the answer ready before they do.

fultonj (commented):
We need to pick one item from the list below.

1. keep 8444 for adoption
2. switch 8444 to 8443
3. support both and allow the customer to decide

I'll describe the pros/cons:

1. Pro
1 is easier to implement now and we could just add a note that green field is using 8443 as it's the ceph default (and I think we should include this history per my other note). Then we'd not change lines 35 and 36.

1. Con
Subsequent tests would need to take that into account. There wont' be standardization between brown and green deployments which makes things more complicated.

2. Pro
Standardized ports per version and simpler in the long run

2. Con
This documentation becomes more complicated as it needs to include migrating the port.

3. Pro
Customer has more options

3. Con
Supportability and Testing are maximally complicated.

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
katarimanojk: @fmount After we decided to go with option2 (switch from 8444 to 8443 during adoption),  i think this comment is no longer applicable

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
katarimanojk: as we decided to go ahead with idea of changing the port  (8444 ->8443) in the adopted environment, 
- i have updated the doc with the steps to update the port in mgr dashboard config along with firewall rules
- i will work on the ceph_migration role and do the needful to match/validate the doc update.

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
```diff
@@ -28,12 +28,24 @@ $ sudo iptables-save
 $ sudo systemctl restart iptables
 ----
 +
+[NOTE]
+Default dashboard port for ceph-mgr in greenfield deployment is 8443. With director-deployed Ceph, the default port is 8444 because the service ran on the controller node and it was necessary to use this port in order to avoid a conflict.For adoption (brownfield), dashboard port should be updated to 8443 in ceph-mgr config and firewall rules.
```
fmount, line 32: nit: `conflitct.For adoption` <- you missed a space

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
```diff
@@ -28,12 +28,24 @@ $ sudo iptables-save
...
+
+. Update dashboard port in ceph mgr config to 8443 using the following steps
```
fmount, line 34: Because this is a procedure, I think this section needs to be extended with a few more wording around the node where these commands are run (e.g. `Login to the first controller node (controller-0) and update the dashboard port ....`).
@klgill to advice more.

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
```diff
@@ -28,12 +28,24 @@ $ sudo iptables-save
...
++
+----
+$ sudo cephadm shell
```
fmount, line 37: should we split this into two steps?

1. Get a cephadm shell:
`command`
2. Run the following commands to update the ceph dashboard port and restart the ceph-mgr module:
`commands`

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
klgill, line 32:
```suggestion
The default dashboard port for `ceph-mgr` in a greenfield deployment is 8443. With director-deployed {Ceph}, the default port is 8444 because the service ran on the Controller node, and it was necessary to use this port to avoid a conflict. For adoption, update the dashboard port to 8443 in the `ceph-mgr` configuration and firewall rules.
```

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
klgill: With the expanded wording above, I think it's okay to keep these commands as one step. However, I wonder what makes sense from a customer's perspective. Is it confusing for the these commands to remain in one step?

klgill (changes requested):

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
fmount: No, we can keep as a single step, it's fine from me. I was wondering if we should split but if it's ok for you to keep them as a single step no changes are required here.

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
fmount: No is not important, technically cephadm works on the three controller nodes, but controller-0 (the first mon) is (historically) the bootstrap  node. I'm ok to simply ask to "Log in to controller-0".

fmount:
@klgill note that when this patch is ready we need a downstream backport (let me know if you need additional stories/task in Jira)

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
katarimanojk: done

docs_user/modules/proc_migrating-mgr-from-controller-nodes.adoc
katarimanojk: done

katarimanojk: done

katarimanojk:
> > @klgill note that when this patch is ready we need a downstream backport (let me know if you need additional stories/task in Jira)
> 
> Yes, please create a task in Jira for me to port this downstream. Thanks!

@klgill @fmount create this task https://issues.redhat.com/browse/RHOSPDOC-2311 for downstream backport.

klgill (approved):

fmount (approved):
/lgtm

PR #841, opened 2025-03-10
specified base64 encoded password

klgill:
https://issues.redhat.com/browse/OSPRH-14567

I also fixed an unrelated formatting issue that I saw in the preview.

This change can be published async, but I think we still need to backport it to the fr2 branch.


docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -248,8 +248,8 @@ $ oc create secret generic subscription-manager \
 --from-literal rhc_auth='{"login": {"username": "<subscription_manager_username>", "password": "<subscription_manager_password>"}}'
 ----
 +
-* Replace `<subscription_manager_username>` with the applicable user name.
-* Replace `<subscription_manager_password>` with the applicable password.
+* Replace `<subscription_manager_username>` with the applicable base64 encoded username.
```
JoanneOFlynn2018, line 251: I would hyphenate base64-encoded.

PR #842, opened 2025-03-10
added limitations

klgill:
https://issues.redhat.com/browse/OSPRH-14487

### DIFF
--- a/docs_user/modules/con_adoption-limitations.adoc
+++ b/docs_user/modules/con_adoption-limitations.adoc
@@ -11,6 +11,9 @@ The following features are considered a Technology Preview and have not been tes
 * {bare_metal_first_ref}
 * NFS Ganesha back end for {rhos_component_storage_file_first_ref}
 * iSCSI, NFS, and FC-based drivers for {block_storage_first_ref}
+* {block_storage} back end for the {image_service_first_ref}
+* NFS back end for the {image_service}
+* Third-party drivers for the {rhos_component_storage_file}
 * The following {compute_service_first_ref} features:
 ** Compute hosts with `/var/lib/nova/instances` on NFS
 ** NUMA aware vswitches


docs_user/modules/con_adoption-limitations.adoc
```diff
@@ -11,6 +11,9 @@ The following features are considered a Technology Preview and have not been tes
 * {bare_metal_first_ref}
 * NFS Ganesha back end for {rhos_component_storage_file_first_ref}
 * iSCSI, NFS, and FC-based drivers for {block_storage_first_ref}
+* {block_storage} back end for the {image_service_first_ref}
+* NFS back end for the {image_service}
+* Third-party drivers for the {rhos_component_storage_file}
 * The following {compute_service_first_ref} features:
 ** Compute hosts with `/var/lib/nova/instances` on NFS
 ** NUMA aware vswitches
```
mickogeary, line 19:
```suggestion
** NUMA-aware vswitches
```

docs_user/modules/con_adoption-limitations.adoc
```diff
@@ -11,6 +11,9 @@ The following features are considered a Technology Preview and have not been tes
 * {bare_metal_first_ref}
 * NFS Ganesha back end for {rhos_component_storage_file_first_ref}
 * iSCSI, NFS, and FC-based drivers for {block_storage_first_ref}
+* {block_storage} back end for the {image_service_first_ref}
+* NFS back end for the {image_service}
+* Third-party drivers for the {rhos_component_storage_file}
 * The following {compute_service_first_ref} features:
```
mickogeary, line 17: Do we need to make these sub- bullet points / nested unordered list?

Could drop the line "The following {compute_service_first_ref} features:"

Or start another list with that line as a lead-in, something like:
"The following {compute_service_first_ref} features are unsupported:"

docs_user/modules/con_adoption-limitations.adoc
```diff
@@ -11,6 +11,9 @@ The following features are considered a Technology Preview and have not been tes
 * {bare_metal_first_ref}
```
mickogeary, line 11: Adding this comment here because I couldn't add it further up. It's about the previous 2 lines:

"Before you proceed with the adoption, check which features are considered a Technology Preview or are unsupported.

The following features are considered a Technology Preview and have not been tested within the context of the {rhos_long_noacro} adoption:"

I'd drop the "are considered" - it's passive and not very technical IMO.
Also maybe it should be "and are unsupported" instead of "or are unsupported" because usually tech preview = unsupported. 

Maybe I wouldn't mention that the features haven't been tested, because if I was a customer I might be thinking "Why are they not tested? Hurry up and test them!"

Maybe something like:

"The following features have a status of Technology Preview and are unsupported in {rhos_long_noacro} production environments:"

Also see about dropping one of those 2 lead in sentences because they are sort of saying the same thing.

docs_user/modules/con_adoption-limitations.adoc
klgill: It means that the following features that only apply to Compute are Tech Preview.

mickogeary: Yes I can see that, but could start a new list instead. Also if it's only for Compute, does that mean a NUMA-aware switch in an NFV use case might be ok, but not in a Compute use case?

klgill: Ok I'll start a new list. This list means that a numa-aware switch would not work in a Compute use case. I'm not sure about an NFV use case.

klgill:
Note to self: Backport to 18.0/fr2 branch after merge.

gfidente:
thanks @klgill , looks good to me

PR #848, opened 2025-03-12
remove base64-encoded

klgill:
https://issues.redhat.com/browse/OSPRH-14567


klgill:
git am --show-current-patch=diff'

PR #856, opened 2025-03-16
Make ospdo adoption namespace changes simpler

pinikomarov:
Jira: https://issues.redhat.com/browse/OSPRH-14644

### DIFF
--- a/docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
+++ b/docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
@@ -17,7 +17,7 @@ You must upgrade the Compute services from {rhos_prev_long} {rhos_prev_ver} to {
 DEFAULT_CELL_NAME="cell1"
 RENAMED_CELLS="$DEFAULT_CELL_NAME"
 
-NAMESPACE="openstack"
+oc project "rhoso"
 
 declare -A PODIFIED_DB_ROOT_PASSWORD
 for CELL in $(echo "super $RENAMED_CELLS"); do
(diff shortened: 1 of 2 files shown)


docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
```diff
@@ -17,7 +17,7 @@ You must upgrade the Compute services from {rhos_prev_long} {rhos_prev_ver} to {
 DEFAULT_CELL_NAME="cell1"
 RENAMED_CELLS="$DEFAULT_CELL_NAME"
 
-NAMESPACE="openstack"
+oc project "rhoso"
```
bogdando, line 20: You should reflect that on the test code as well, there is some logic build around NAMESPACE env var, so those must match the docs

bogdando (changes requested):
see inline

docs_user/modules/proc_performing-a-fast-forward-upgrade-on-compute-services.adoc
jistr: Yes but is the `oc project "rhoso"` scoped for OSPDO docs only? To me it looks like we're changing the default on non-OSPDO too here.

docs_user/modules/proc_preparing-controller-nodes-for-director-operator-adoption.adoc
```diff
@@ -12,6 +12,9 @@ The following procedure uses a single {rhocp_long} cluster that is shared by the
 .Procedure
 
 . Log in to the {OpenShiftShort} environment where {OpenStackShort} OSPdO is deployed and change to the project that hosts your {OpenStackShort} deployment:
+[NOTE]
+In Director Operator adoption you must define which namespace is being used for resources manipulation.
+Since with Director Operator the adopted namespace would be "openstack", the adopting namespace ,for rhoso , would be "rhoso".
```
klgill, line 17:
```suggestion
In director Operator adoption, you must define which namespace is being used for resources manipulation.
For director Operator adoption, the adopted namespace is `openstack`. If you are adopting a RHOSO environment, the adopted namespace is `rhoso`.
```
I'm not sure if we need the last sentence about RHOSO, since this procedure only applies to director Operator.

klgill (changes requested):

PR #517, opened 2024-07-03
Multi-cell adoption

bogdando:
Split edpm nodes into compute cells by 1:1 mapping it as
dataplane nodesets.

Use edpm_nodes var to describe compuptes for each cell,
instead of static host and ip vars that only used to work for
a single-cell standalone, or multi-node single cell cases.
Also explain EDPM net config requirements in vars.sample, when
it is used outside of ci-framework (local deployments).

Remove edpm_computes vars no longer used after moving stopping
control-plane tripleo services into edpm-ansible

Simplify ENV headers management by collecting in a single place.

Provide a variable to define the source cloud Ironic topology,
for any cells with Ironic services.

Align nova/libvirt and related services ordering in the
lists of services defined in multiple places, with those
specified in VA.

Align the names in the tests to follow the documented steps
to make the corresponding code easy discoverable.

Adjust storage/storageRequests values to make it better fitting
a multi-cell test scenarios. Also provide values in docs and
add a comment to adjust them as needed.

Stop ovn services only if active, or not missing (like on
the cell controllers)

Retain EDPM host IPs on internalapi network. Without that, edpm-ansible's os-net-config
changes IPs on internalapi, and also breaks connectivity to EDPM hosts for ansible
(which restores after a node reboot).

Add edpmRoleServiceName value for tlsCerts.

Jira: #[OSPRH-6548](https://issues.redhat.com/browse/OSPRH-6548)

PR #858, opened 2025-03-17
Add missing ceph-hci-pre alongside ceph-client

bogdando:
In the architecture repo, before the ceph-client service, there is always a block of EDPM services:

- install-os
- ceph-hci-pre
- configure-os

Add missing ceph-hci-pre to the list.

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -478,6 +478,7 @@ endif::[]
     - configure-network
     - validate-network
     - install-os
+    - ceph-hci-pre
     - configure-os
     - ssh-known-hosts
     - run-os


bogdando:
Don't merge before https://github.com/openstack-k8s-operators/data-plane-adoption/pull/517

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -478,6 +478,7 @@ endif::[]
     - configure-network
     - validate-network
     - install-os
+    - ceph-hci-pre
```
fmount, line 481: quick question: this service is supposed to trigger [1], which currently only adds the ceph firewall rules.
In theory this is valid for the use case of HCI adoption (especially if the idea is to decommission controller nodes and migrate the ceph daemons on the HCI nodes).
For non HCI use cases this might not be required. I didn't look at the whole procedure, and I'm wondering if you always want to have ceph-hci pre or not.
Other than this question I have no objections to the patch itself, just wanted to clarify the use case because we have multiple scenarios from a documentation perspective:

1. tripleo with cephHCI
2. tripleo with cephStorage nodes (they're currently not enrolled as EDPM nodes)

I assume the use case is for HCI and you need, in any case, to pass a variation of [1] to only open Ceph OSDs ports.
In that case I think this patch should provide those variable overrides.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_ceph_hci_pre/defaults/main.yml#L130

[1] https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_ceph_hci_pre

bogdando: > I'm wondering if you always want to have ceph-hci pre or not.

That is what I got from the architecture repository, I cannot really answer to that Q

fmount: > That is what I got from the architecture repository, I cannot really answer to that Q

ack, it doesn't hurt but I'd like to have more details around this change. Adding @fultonj that might have an opinion other than a good understanding of architecture repo.

bogdando: Thanks for explaining, I guess we don't need this change then
@jistr fyi

PR #859, opened 2025-03-17
Optional Octavia SSH adoption test

weinimo:
Octavia SSH key adoption is an optional step and should not block verification. Currently this test does not work in downstream CI.


klgill (approved):
@weinimo FYI I only need to review docs that are in the docs_user folder, but this LGTM.

jistr (commented):
/lgtm
/approve

PR #722, opened 2024-11-12
implementing QE comments

klgill:
I implemented the QE comments on the adoption guide from https://docs.google.com/document/d/1f6t_Jkq9jl4xfWytWIkX3FnYPRFir-oPufy3AMESAGg/edit?tab=t.0#heading=h.kg1e9845y8p6

PR #866, opened 2025-03-21
Control plane verification fixes

jistr:
* [x] check/amend tests

Resolves: https://issues.redhat.com/browse/OSPRH-14744

### DIFF
--- a/docs_user/modules/proc_adopting-the-identity-service.adoc
+++ b/docs_user/modules/proc_adopting-the-identity-service.adoc
@@ -72,18 +72,16 @@ done
 
 .Verification
 
-. Verify that the setup of the `OpenStackControlPlane` CR is now complete:
-+
-----
-$ oc get OpenStackControlPlane openstack
-NAME        STATUS   MESSAGE
-openstack   True     Setup complete
-----
-
 . Verify that you can access the `OpenStackClient` pod. For more information, see link:{defaultURL}/maintaining_the_red_hat_openstack_services_on_openshift_deployment/assembly_accessing-the-rhoso-cloud#proc_accessing-the-OpenStackClient-pod_cloud-access-admin[Accessing the OpenStackClient pod] in _Maintaining the {rhos_long_noacro} deployment_.
 
 . Confirm that the {identity_service} endpoints are defined and are pointing to the control plane FQDNs:
 +
 ----
 $ openstack endpoint list | grep keystone
+----
+
+. Wait for the `OpenStackControlPlane` resource to become `Ready`:
++
+----
+$ oc wait --for=condition=Ready --timeout=1m OpenStackControlPlane openstack
 ----
--- a/docs_user/modules/proc_adopting-the-identity-service.adoc
+++ b/docs_user/modules/proc_adopting-the-identity-service.adoc
@@ -72,6 +72,14 @@ done
 
 .Verification
 
+. Verify that the setup of the `OpenStackControlPlane` CR is now complete:
++
+----
+$ oc get OpenStackControlPlane openstack
+NAME        STATUS   MESSAGE
+openstack   True     Setup complete
+----
+
 . Verify that you can access the `OpenStackClient` pod. For more information, see link:{defaultURL}/maintaining_the_red_hat_openstack_services_on_openshift_deployment/assembly_accessing-the-rhoso-cloud#proc_accessing-the-OpenStackClient-pod_cloud-access-admin[Accessing the OpenStackClient pod] in _Maintaining the {rhos_long_noacro} deployment_.
 
 . Confirm that the {identity_service} endpoints are defined and are pointing to the control plane FQDNs:
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -26,7 +26,7 @@ endif::[]
 
 * If you enabled TLS everywhere (TLS-e) on the {OpenStackShort} environment, you must copy the `tls` root CA from the {OpenStackShort} environment to the `rootca-internal` issuer.
 
-* There are free PVs available for MariaDB and RabbitMQ.
+* There are free PVs available for Galera and RabbitMQ.
 ifeval::["{build}" != "downstream"]
 +
 For developer/CI environments driven by install_yamls, make sure
@@ -415,7 +415,7 @@ spec:
 
 .Verification
 
-* Verify that the MariaDB and RabbitMQ status is `Running` for all defined cells:
+* Verify that the Galera and RabbitMQ status is `Running` for all defined cells:
 +
 [source,yaml]
 ----
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -429,4 +429,11 @@ done
 +
 The given cells names are later referred to by using the environment variable `RENAMED_CELLS`.
 
-* Verify that you can access the `OpenStackClient` pod. For more information, see link:{defaultURL}/maintaining_the_red_hat_openstack_services_on_openshift_deployment/assembly_accessing-the-rhoso-cloud#proc_accessing-the-OpenStackClient-pod_cloud-access-admin[Accessing the OpenStackClient pod] in _Maintaining the {rhos_long_noacro} deployment_.
+* Verify that the `OpenStackControlPlane` CR is waiting for deployment
+  of the `openstackclient` pod:
++
+----
+$ oc get OpenStackControlPlane openstack
+NAME        STATUS    MESSAGE
+openstack   Unknown   OpenStackControlPlane Client not started
+----
--- a/docs_user/modules/proc_deploying-backend-services.adoc
+++ b/docs_user/modules/proc_deploying-backend-services.adoc
@@ -415,16 +415,17 @@ spec:
 
 .Verification
 
-* Verify that MariaDB and RabbitMQ are running for all defined cells:
+* Verify that the MariaDB and RabbitMQ status is `Running` for all defined cells:
 +
+[source,yaml]
 ----
 $ RENAMED_CELLS="cell1 cell2 cell3"
-$ oc get pod openstack-galera-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-$ oc get pod rabbitmq-server-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
+$ oc get pod openstack-galera-0 -o jsonpath='{.status.phase}{"\n"}'
+$ oc get pod rabbitmq-server-0 -o jsonpath='{.status.phase}{"\n"}'
 $ for CELL in $(echo $RENAMED_CELLS); do
-    oc get pod openstack-$CELL-galera-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-    oc get pod rabbitmq-$CELL-server-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-done
+>     oc get pod openstack-$CELL-galera-0 -o jsonpath='{.status.phase}{"\n"}'
+>     oc get pod rabbitmq-$CELL-server-0 -o jsonpath='{.status.phase}{"\n"}'
+> done
 ----
 +
 The given cells names are later referred to by using the environment variable `RENAMED_CELLS`.
--- a/tests/roles/keystone_adoption/tasks/main.yaml
+++ b/tests/roles/keystone_adoption/tasks/main.yaml
@@ -56,6 +56,17 @@
   retries: 60
   delay: 2
 
+- name: verify that OpenStackControlPlane setup is complete
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+
+    oc get openstackcontrolplane openstack -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}{"\n"}' | grep True
+  register: openstack_control_plane_cr_result
+  until: openstack_control_plane_cr_result is success
+  retries: 60
+  delay: 2
+
 - name: clean up services and endpoints
   ansible.builtin.shell: |
     {{ shell_header }}
--- a/tests/roles/keystone_adoption/tasks/main.yaml
+++ b/tests/roles/keystone_adoption/tasks/main.yaml
@@ -60,12 +60,7 @@
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-
-    oc get openstackcontrolplane openstack -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}{"\n"}' | grep True
-  register: openstack_control_plane_cr_result
-  until: openstack_control_plane_cr_result is success
-  retries: 60
-  delay: 2
+    oc wait --for=condition=Ready --timeout=1m OpenStackControlPlane openstack
 
 - name: clean up services and endpoints
   ansible.builtin.shell: |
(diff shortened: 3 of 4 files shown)


docs_user/modules/proc_adopting-the-identity-service.adoc
```diff
@@ -72,9 +72,18 @@ done
 
 .Verification
 
+. Verify that the setup of `OpenStackControlPlane` CR is now complete:
```
klgill, line 75:
```suggestion
. Verify that the setup of the `OpenStackControlPlane` CR is now complete:
```

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -401,18 +401,25 @@ named `cell1`. Adjust the names, counts, IP addresses, and numbers, such as for
 
 .Verification
 
-* Verify that MariaDB and RabbitMQ are running for all defined cells:
+* Verify that MariaDB and RabbitMQ status is "Running" for all defined cells:
```
klgill, line 404:
```suggestion
* Verify that the MariaDB and RabbitMQ status is `Running` for all defined cells:
```
Do customers see an output called `Running`?
If not, I'd probably change this sentence to "...is running for all defined cells:"

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -401,18 +401,25 @@ named `cell1`. Adjust the names, counts, IP addresses, and numbers, such as for
...
 $ RENAMED_CELLS="cell1"
-$ oc get pod openstack-galera-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-$ oc get pod rabbitmq-server-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
+$ oc get pod openstack-galera-0 -o jsonpath='{.status.phase}{"\n"}'
+$ oc get pod rabbitmq-server-0 -o jsonpath='{.status.phase}{"\n"}'
 $ for CELL in $(echo $RENAMED_CELLS); do
-    oc get pod openstack-$CELL-galera-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-    oc get pod rabbitmq-$CELL-server-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
+    oc get pod openstack-$CELL-galera-0 -o jsonpath='{.status.phase}{"\n"}'
+    oc get pod rabbitmq-$CELL-server-0 -o jsonpath='{.status.phase}{"\n"}'
```
klgill, line 412: Is the dollar sign omitted on lines 411 and 412 because these lines are indented?

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -401,18 +401,25 @@ named `cell1`. Adjust the names, counts, IP addresses, and numbers, such as for
...
 done
 ----
 +
 The given cells names are later referred to by using the environment variable `RENAMED_CELLS`.
 
-* Verify that you can access the `OpenStackClient` pod. For more information, see link:{defaultURL}/maintaining_the_red_hat_openstack_services_on_openshift_deployment/assembly_accessing-the-rhoso-cloud#proc_accessing-the-OpenStackClient-pod_cloud-access-admin[Accessing the OpenStackClient pod] in _Maintaining the {rhos_long_noacro} deployment_.
+* Verify that the `OpenStackControlPlane` CR is waiting for deployment
+  of the `openstackclient` pod:
```
klgill, line 419: FYI: We're inconsistent about the use of `openstackclientpod` vs. OpenStackClientPod in the adoption guide and other 18.0 guides. Another writer on my team is investigating which term is correct. We'll leave this as is until we get an answer.

klgill (changes requested):

docs_user/modules/proc_deploying-backend-services.adoc
jistr: Yes lines 410-413 are to be copy pasted as a single command essentially. I saw in some other places we use the `>` character at the beginning of the line to signify this is a continuation line, but it kinda prevents copy-pasting the multi-line command.

jistr: In general i think achieving consistency on this would be nice :)

klgill: We usually use forward slashes in the docs to signify a continuation line, like this example in the 17.1 FFU guide:
export STACK=<stack> \
$ sudo awk '/tripleo_role_name/ {print "--role " $2}' \
sudo awk '/tripleo_role_name/ {print "--role " $2}' \
/var/lib/mistral/${STACK}/tripleo-ansible-inventory.yaml \
| grep -vi compute

We should be using them upstream as well, instead of the > character.

jistr: The example you pasted didn't render well in the comment, but i guess you meant backward slashes? I can see those in FFU guide.

Here it's different though, i misspoke when i mentioned it's a continuation line, it's actually an open `for` command but not really a continuation line. There are multiple lines and it's in a single control loop, but it's not a single command. If we put there the backslashes, it wouldn't work correctly, unless we also use semicolons after the commands. So it's unclear to me what the best convention for these multi-line control structures should be...

klgill: I'll ask the docs team for advice.

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -401,18 +401,25 @@ named `cell1`. Adjust the names, counts, IP addresses, and numbers, such as for
 
 .Verification
 
-* Verify that MariaDB and RabbitMQ are running for all defined cells:
+* Verify that the MariaDB and RabbitMQ status is `Running` for all defined cells:
 +
```
klgill, line 405:
```suggestion
+
[source,yaml]
```
The source tag needs to be added because the copy icon on the published guide is only copying the first 4 lines.

docs_user/modules/proc_deploying-backend-services.adoc
jistr: I tried the copy button and i saw the copied text does not include the `$` and `>` characters that are used to signal start of the command and multi-line commands. I don't know what's the mechanism behind that but i was pleasantly surprised :). So if we'd like to use that solution globally in the adoption docs, i'd be +1.

jistr: For now i updated the PR to use the `>` prefix for the multi-line command.

ciecierski:
Can you link jira bug in description on this pull request? There is automation which adds linking in jira to pull request. Thank you

jistr:
Linked the bug, i think perhaps we should also put this on hold until #517 lands.

/hold

docs_user/modules/proc_deploying-backend-services.adoc
klgill: Let's wait to add the solution globally. In the past week, I've noticed that the copy icon only copies the first line in many of the code blocks in the adoption guide. I solved some of them by adding the [source,yaml] tag. I need to test whether adding the [source,yaml] tag works on code blocks without the `>` . If it does work, then we don't need the `>` and the docs would look cleaner.

jistr:
/hold
Holding this until it's clear whether we will revert #517 or not.

docs_user/modules/proc_adopting-the-identity-service.adoc
```diff
@@ -72,9 +72,16 @@ done
...
+----
+
 . Verify that you can access the `OpenStackClient` pod. For more information, see link:{defaultURL}/maintaining_the_red_hat_openstack_services_on_openshift_deployment/assembly_accessing-the-rhoso-cloud#proc_accessing-the-OpenStackClient-pod_cloud-access-admin[Accessing the OpenStackClient pod] in _Maintaining the {rhos_long_noacro} deployment_.
 
-. Confirm that the {identity_service} endpoints are defined and are pointing to the control plane FQDNs:
+. Using the `OpenStackClient` pod remote shell, verify that the {identity_service}
+endpoints are defined and are pointing to the control plane FQDNs:
 +
 ----
 $ openstack endpoint list | grep keystone
```
ciecierski, line 87: This command uses alias created before. Maybe ` Using the `OpenStackClient` pod remote shell` should be explained above in alias creation step?

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -429,18 +429,26 @@ such as for `replicas`, `storage`, or `storageRequest`, as needed.
 
 .Verification
 
-* Verify that MariaDB and RabbitMQ are running for all defined cells:
+* Verify that the MariaDB and RabbitMQ status is `Running` for all defined cells:
```
ciecierski, line 432: s/MariaDB/Galera

docs_user/modules/proc_adopting-the-identity-service.adoc
jistr: Ah i didn't scan for the alias, i thought this was missing a "run this command in the client pod" explanation. But since we rely on the alias instead, i'll undo this edit and leave the original formulation as-is.

jistr:
recheck image pull error

docs_user/modules/proc_adopting-the-identity-service.adoc
```diff
@@ -72,6 +72,12 @@ done
 
 .Verification
 
+. Wait for the `OpenStackControlPlane` resource to become `Ready`:
```
ciecierski, line 75: This should be after `openstack endpoint list | grep keystone` as in test, or test should be fixed to be consistent

jistr: Yea fair enough it probably makes sense to check the keystone-specific functionality first and only then see if the control plane as a whole converged.

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -415,18 +415,26 @@ spec:
 
 .Verification
 
-* Verify that MariaDB and RabbitMQ are running for all defined cells:
+* Verify that the Galera and RabbitMQ status is `Running` for all defined cells:
 +
+[source,yaml]
 ----
 $ RENAMED_CELLS="cell1 cell2 cell3"
```
odyssey4me, line 422: This may be slightly tangential, but given that the most common use case is not to have multiple cells don't you think we should have a note here to clarify that this should be modified to suit the environment?

docs_user/modules/proc_deploying-backend-services.adoc
```diff
@@ -415,18 +415,26 @@ spec:
...
-$ oc get pod rabbitmq-server-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
+$ oc get pod openstack-galera-0 -o jsonpath='{.status.phase}{"\n"}'
+$ oc get pod rabbitmq-server-0 -o jsonpath='{.status.phase}{"\n"}'
 $ for CELL in $(echo $RENAMED_CELLS); do
-    oc get pod openstack-$CELL-galera-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-    oc get pod rabbitmq-$CELL-server-0 -o jsonpath='{.status.phase}{"\n"}' | grep Running
-done
+>     oc get pod openstack-$CELL-galera-0 -o jsonpath='{.status.phase}{"\n"}'
+>     oc get pod rabbitmq-$CELL-server-0 -o jsonpath='{.status.phase}{"\n"}'
+> done
```
odyssey4me, line 428: I think it may make sense to include a sample of expected output here.

jistr:
@odyssey4me Thanks for the review, but given that this patch has been on review for 3 weeks and the sprint is coming to a close, i think i'd focus any edits here on things that are clear breakages.

jistr:
I'd like to avoid respinning the CI on tangential issues.

docs_user/modules/proc_deploying-backend-services.adoc
jistr: The expected output is the word `Running` per cell, i tried to hint at that in the introductory sentence for the code block. It could be made more clear a the cost of being more verbose, but i'm not sure if that's worth respinning the CI.

docs_user/modules/proc_deploying-backend-services.adoc
jistr: Yes could be done but it's not related to the bug i'm trying to fix. I could piggy-back another commit to the PR if it was brought up earlier in the sprint but now i'd just focus on the must haves.

The `RENAMED_CELLS` variable is used across the docs in many places, it is described in https://openstack-k8s-operators.github.io/data-plane-adoption/user/downstream.html#adopting-the-compute-service_adopt-control-plane

Now that is searched for it, i see that description is not the first place the variable occurs in the docs, which could be improved upon, but again it's irrelevant to this bug.

odyssey4me:
@jistr Yes, fair enough - my suggestions are docs-only and could follow up in a PR that's docs-only to avoid respinning the integration testing in CI.

holser (commented):
/lgtm

holser (approved):

docs_user/modules/proc_deploying-backend-services.adoc
bogdando: In the guide, it is stated as an example we stick with to deploy 3 cells further on.
We could surely improve that in some other commits

docs_user/modules/proc_deploying-backend-services.adoc
bogdando: could you preserve the docs link?

bogdando (approved):
lgtm

docs_user/modules/proc_deploying-backend-services.adoc
jistr: The content doesn't belong here at all. It already got deleted from here once (moved to adopting identity service where we can actually connect to the pod, here we can't). It got reintroduced as a mistake in a rebase of one of the larger patches, so now i'm just re-deleting it :).

jistr: It's here: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/866/files#diff-3a3c5df1a0bf117a4167ea85e9e87727e6ceec6c3a466bc5f60b566cf55f53bfR75

jistr: Relevant PR that originally moved the content: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/722

docs_user/modules/proc_deploying-backend-services.adoc
ciecierski: We can be more verbose in documentation and have grep in CI at the same time.
Here is example output:
`$ oc get pods -n openstack rabbitmq-server-0  
NAME                READY   STATUS    RESTARTS   AGE
rabbitmq-server-0   1/1     Running   0          51m`

and for galera:
` oc get pods -n openstack openstack-galera-0
NAME                 READY   STATUS    RESTARTS   AGE
openstack-galera-0   1/1     Running   0          53m`

tests/roles/keystone_adoption/tasks/main.yaml
```diff
@@ -56,6 +56,12 @@
   retries: 60
   delay: 2
 
+- name: verify that OpenStackControlPlane setup is complete
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc wait --for=condition=Ready --timeout=1m OpenStackControlPlane openstack
```
ciecierski, line 63: It won't hurt to add this, but if previous task pass this will always pass. Maybe the order of checks should be reverted. Check on OpenShift level first and then on Openstack level.

Just a side note: I was thinking at some point how it is possible that it is working as OpenStackControlPlane is not fully deployed yet returns condition=Ready.
I came to realization that is, because we are enable services one-by-one and by doing so this will return Ready state after all currently adopted services are deployed.

jistr: Based on your earlier comment

> This should be after openstack endpoint list | grep keystone as in test, or test should be fixed to be consistent

I synced the docs to use the same order as tests.

jistr:
2x lgtm, comments addressed.
/approve

tests/roles/keystone_adoption/tasks/main.yaml
ciecierski: ack

jistr:
@odyssey4me Follow ups:
https://github.com/openstack-k8s-operators/data-plane-adoption/pull/932 
https://issues.redhat.com/browse/OSPRH-15896

PR #871, opened 2025-03-25
Enable Ceph dashboard in unigamma

fmount:
Unigamma lands in a Ceph configuration that includes ceph dashboard and the associated monitoring stack. This patch enables the Ceph dashboard in the TripleO scenario so we can cover the monitoring stack migration.


fmount:
In general we document how to move the monitoring stack to the target nodes, but we currently do not test it.

fmount:
@katarimanojk currently testing this as part of unigamma adoption.

katarimanojk:
With this patch , i see the ceph monitoring stack is enabled (in osp17.1 deployment ) during unigamma adoption(hci) scenario
so we could move ahead and test the migration of ceph monitoring stack to target nodes successfully.

fmount:
recheck

katarimanojk:
recheck

katarimanojk (commented):
Good to test ceph dashboard migration in unigamma jobs

ciecierski:
You will need cherry-pick to FR2 if target is before FR3

fmount:
@ciecierski makes sense, cherry-picking!

PR #874, opened 2025-03-26
Fixed incorrect grep in octavia adoption

gthiemonge:
Use "\\|" instead of "|" in grep, previous regexp didn't work as expected because grep considers "octbr|enp6s0.24" as a full-string.

It fixes a bug if the octavia bridge and interface are already present in the NNCP, this NNCP might have been patched again with a different configuration.

JIRA: [OSPCIX-747](https://issues.redhat.com//browse/OSPCIX-747)


eduolivares (approved):

karelyatin (commented):
/lgtm
Other option would have been to use either -E or egrep

eduolivares (approved):

PR #876, opened 2025-03-27
fixed ovn/networking adoption bugs

klgill:
This PR resolves the following doc bugs:

1.  https://issues.redhat.com/browse/OSPRH-14900
2. https://issues.redhat.com/browse/OSPRH-14902


docs_user/modules/proc_migrating-ovn-data.adoc
```diff
@@ -288,9 +285,18 @@ spec:
...
+  ovn:
+    enabled: true
+    template:
+      ovnController:
+        nicMappings:
+          physNet: NIC <1>
 ----
 +
+<1> `physNet` is the name of your physical network. `NIC` is the name of the physical interface that is connected to your physical network.
++
```
klgill, line 299: @slawqo Are users expected to *replace* `physNet` and `NIC` with the names of their own physical network and physical interface?

slawqo: yes, by default physnet which was set in OSP was named `datacentre` and it should be configured to be the same as it was in the adopted OSP environment. NIC needs to be set to the one which actually provides connectivity to that physical network.

slawqo (approved):
LGTM for me

PR #460, opened 2024-05-14
Fixing the adopting the data plane doc

afaranha:
Fixing the command to deploy a standalone environment with TLS, create rootca-internal secret on openstack namespace.

PR #893, opened 2025-04-03
[tls_adoption] Set namespace explicitly for secret create

karelyatin:
Docs were fixed long back with [1], also fixing the role.

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/pull/460


jistr:
Based on recent failures of the standalone-non-ceph job this could fix it. Approving to get it auto-merged if CI passes.
/approve

PR #900, opened 2025-04-04
Fix ping test failures in pre_launch.bash

jistr:
The `ping -c4` is sometimes not enough time for the instance to start responding, and the ping test fails. We shouldn't just increase the number for `-c` because that would prolong the test even if the ping is already responding. We now use a small bash loop to quit at the first ping success or error out due to (now considerably longer) timeout.

### DIFF
--- a/tests/roles/development_environment/files/pre_launch.bash
+++ b/tests/roles/development_environment/files/pre_launch.bash
@@ -118,7 +118,14 @@ ${BASH_ALIASES[openstack]} security group rule list --protocol tcp --ingress -f 
 
 export FIP=192.168.122.20
 # check connectivity via FIP
-ping -c4 ${FIP}
+TRIES=0
+until ping -D -c1 -W2 "$FIP"; do
+    ((TRIES++)) || true
+    if [ "$TRIES" -gt 20 ]; then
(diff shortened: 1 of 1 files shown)


tests/roles/development_environment/files/pre_launch.bash
```diff
@@ -118,7 +118,14 @@ ${BASH_ALIASES[openstack]} security group rule list --protocol tcp --ingress -f
 
 export FIP=192.168.122.20
 # check connectivity via FIP
-ping -c4 ${FIP}
```
karelyatin, line 121: this will not work as set -e is set on top so script will exit on first failure. we can add set +e before ping block, and set -e post it.

tests/roles/development_environment/files/pre_launch.bash
```diff
@@ -118,7 +118,14 @@ ${BASH_ALIASES[openstack]} security group rule list --protocol tcp --ingress -f
...
+TRIES=0
+until ping -D -c1 -W2 "$FIP"; do
```
holser, line 122: Do we really need output here?

> until ping -D -c1 -W2 "$FIP" &> /dev/null; do

PR #904, opened 2025-04-07
copy keys from undercloud for tobiko

fyanac:


jistr:
The GH diff view doesn't make it apparent that it's conditioned inside `when: tobiko_qe_test | default('false') | bool`, but it is :).
/lgtm
/ok-to-test

tests/roles/development_environment/tasks/main.yaml
```diff
@@ -150,3 +152,7 @@
       ansible.builtin.shell:
         chdir: ~/src/x/tobiko/
         cmd: ansible-playbook tobiko-playbook.yaml -e test_workflow=create-resources -e pytest_addopts_global="--skipregex='TestFloatingIPLogging|ExtraDhcpOptsPortLoggingTest|NoFipPortTest|StatelessSecurityGroupInstanceTest|StatelessSecurityGroupTest|migrate_server|qos|create_share|extend_share'"
+
+    - name: copy keys from undercloud for tobiko
+      ansible.builtin.shell: |
+        mkdir -p ~/ci-framework-data/tests/test_operator; scp -r ${OS_CLOUD_IP}:~/.ssh/id_ecdsa* ~/ci-framework-data/tests/test_operator/
```
ciecierski, line 158: Do you need all keys or key? 
There is `edpm_privatekey_path` variable pointing to key path and you can use it instead of scp -r if you need just a key.

fyanac: The Tobiko workload is created before adoption and the _id_ecdsa_ key is also created by Tobiko and is required to access the workload(vm's) after adoption.

PR #905, opened 2025-04-08
Add edpm_container_registry_insecure_registries for networker

fyanac:


fyanac:
I ran the patch with the d/s job and it works

ciecierski:
This change is just aligning of compute and networker 
edpm_container_registry_insecure_registries to be present in both.
/lgtm
/approve

PR #913, opened 2025-04-09
removed source yaml tag

klgill:
https://issues.redhat.com/browse/OSPRH-14901


klgill:
@jistr I removed the [source, yaml] tag because this tag is causing the dollar sign to be copied when clicking the copy button in the downstream docs. Generally, only the first line (the command) should be copied. However, there are a lot of instances in the adoption guide where the command (oc patch or oc apply) is followed by "spec:" or "apiVersion:" . Should the content following the command be copied as well, or is that content just an example?

jistr:
The content in the blocks is an example that the user will likely want to edit. But even then i would expect that the copy button copies the example content too.

PR #927, opened 2025-04-16
Fix rgw keystone endpoint command

fmount:


katarimanojk:
@fmount thanks for identifying the bug, with this fix,  keystone_url will be set correctly there by we won't face any problems in using swift client to reach the rgw backend.

PR #930, opened 2025-04-16
removed fixed known issue and hid known issues section

klgill:
The bug was fixed and this known issue is no longer needed: https://issues.redhat.com/browse/OSPRH-11301


klgill:
@karelyatin FYI, I'm removing a known issue for the following bug because it was fixed in FR2: https://issues.redhat.com/browse/OSPRH-11301

PR #873, opened 2025-03-25
ospdo adoption ns overhaul

pinikomarov:
- Set namespace only at the beginning of adoption guide for ospdo differentiation
- OSPdO namespaces should be explicitly specified

PR #938, opened 2025-04-28
[DNM] Test if override-checkout is required

ciecierski:
Test if 'override-checkout: main' introduced in #873  is needed


ciecierski:
Based on testing in the change, override-checkout is not needed.

pinikomarov (commented):
/lgtm

PR #939, opened 2025-04-28
Fix order of OpenStackDataplaneServices in adopting compute docs

ciecierski:
Fix order of OpenStackDataplaneServices in documentation to be correct and the same as in test suite.[1]

[1]https://github.com/openstack-k8s-operators/data-plane-adoption/blob/18.0-fr2/tests/roles/dataplane_adoption/defaults/main.yaml#L184

Resolves: [OSPRH-15291](https://issues.redhat.com//browse/OSPRH-15291)


klgill (approved):
LGTM

holser (approved):

PR #949, opened 2025-05-14
Disable NetworkManager DNS config update

Jaganathancse:
This chenge is to prevent the DNS config update via NetworkManager for ifcfg provider and only set this edpm ansible variable 'edpm_bootstrap_network_resolvconf_update' true for supporting nmsate provider.


ekuris-redhat (approved):
Legit

Jaganathancse:
recheck

ekuris-redhat:
recheck

ekuris-redhat:
recheck

PR #952, opened 2025-05-15
Configure swift to use rgw backend during ceph migration

katarimanojk:
In the brownfield context,  currently  swift is not configured for rgw post ceph migration.

This patch configures Swift service and endpoints so that swift client can be used to reach ceph rgw backend post migration.

This is now done as post activity during ceph migration.

JIRA: https://issues.redhat.com/browse/OSPRH-16076


tests/roles/ceph_migrate/defaults/main.yml
```diff
@@ -48,3 +48,6 @@ ceph_node_exporter_container_image: "quay.io/prometheus/node-exporter:v1.5.0"
 ceph_prometheus_container_image: "quay.io/prometheus/prometheus:v2.43.0"
 ceph_storagenfs_nic: "nic2"
 ceph_storagenfs_vlan_id: "70"
+
+cifmw_cephadm_certificate: ""
```
katarimanojk, line 52: This may not be needed if we don't want to think/act about tls configuration during ceph_migration.

In  the current context, we use this to set `cifmw_cephadm_urischeme` only and nothing more for tls.

Also  there shoudn't be 'cifmw' references in data_plane_adoption repo

tests/roles/ceph_migrate/tasks/configure_object.yaml
```diff
@@ -0,0 +1,76 @@
...
+- name: Create swift service
+  delegate_to: localhost
+  ansible.builtin.shell: |
+    oc -n openstack rsh openstackclient openstack service create --name swift --description 'OpenStack Object Storage' object-store
+  when:
+    - not swift_in_ctlplane.stdout | bool
+    - swift_endpoints_count.stdout == "0"
+    - swift_services_count.stdout == "0"
+
+- name: Update urischeme based on cert/key
```
katarimanojk, line 40: This task can be removed and just use http (without tls) endponts during ceph_migration.

In case if the tls is needed, later rgw tls configuration is done by cifmw and the endpoints are updated anyway

tests/roles/ceph_migrate/tasks/configure_object.yaml
```diff
@@ -0,0 +1,68 @@
+- name: Check for existing object store config
+  delegate_to: localhost
+  block:
+    - name: Check if swift is enabled in deployed controlplane
+      ansible.builtin.shell: |
+        {{ shell_header }}
+        oc -n openstack get $(oc get oscp -n openstack -o name) -o json| jq .spec.swift.enabled
```
fmount, line 7: `-n $NAMESPACE`

can we have a variable to make sure it will work with a NS != openstack?

tests/roles/ceph_migrate/tasks/configure_object.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - swift_services_count.stdout == "0"
+
+- name: Configure object store to use rgw
+  delegate_to: localhost
+  when:
+    - not swift_in_ctlplane.stdout | bool
+    - swift_endpoints_count.stdout == "0"
+  block:
+    - name: Capture required UUIDs
+      ansible.builtin.command: "oc -n openstack rsh openstackclient openstack {{ item.os_command }} show {{ item.os_command_object }} -c id -f value"
```
fmount, line 47: same comment here about the namespace (it can be unpacked from an ansible variable probably)

tests/roles/ceph_migrate/tasks/configure_object.yaml
```diff
@@ -0,0 +1,68 @@
...
+- name: Display a note about swift deployment
+  ansible.builtin.debug:
+    msg: "WARNING: Swift is deployed and the endpoint exists already, ceph RGW cannot be configured as object store service"
+  when:
+    - swift_in_ctlplane.stdout | bool
+
+- name: Create swift service
+  delegate_to: localhost
+  ansible.builtin.shell: |
+    oc -n openstack rsh openstackclient openstack service create --name swift --description 'OpenStack Object Storage' object-store
```
fmount, line 34: same comment about the namespace, that might probably be a default global var

fmount (commented):
Just a minor comment around the namespace otherwise lgtm.

Last question: I assume this is not called in period multijobs because rgw is not deployed there, correct?

tests/roles/ceph_migrate/tasks/configure_object.yaml
katarimanojk: dpme

tests/roles/ceph_migrate/tasks/configure_object.yaml
katarimanojk: done

tests/roles/ceph_migrate/tasks/configure_object.yaml
katarimanojk: done

fmount (approved):
/lgtm

tests/roles/ceph_migrate/defaults/main.yml
```diff
@@ -48,3 +48,4 @@ ceph_node_exporter_container_image: "quay.io/prometheus/node-exporter:v1.5.0"
 ceph_prometheus_container_image: "quay.io/prometheus/prometheus:v2.43.0"
 ceph_storagenfs_nic: "nic2"
 ceph_storagenfs_vlan_id: "70"
+openstack_namespace: "openstack"
```
jistr, line 51: In the rest of the adoption docs & tests, we got rid of explicit `-n <namespace>`, in favor of selecting the namespace via `oc project <namespace>` at the beginning of Adoption and then relying on it being already set correctly. This was to allow documenting both Director and Director Operator adoption without too much confusion. The only guide which does `-n` are some bits of Director Operator guide which are meant to talk to the source Director Operator namespace rather than the RHOSO namespace.

I wouldn't say it's a blocker for this PR but i think we should probably converge onto that approach.

jistr (commented):
/lgtm

tests/roles/ceph_migrate/defaults/main.yml
katarimanojk: @jistr thanks for the details , i see namespace is set in [1] when the adoption playbook (test_with_ceph.yaml) is run. 
The current role 'ceph_migrate' is run from another playbook (test_externalize_ceph.yaml) which is run immediately after adoption playbook.

i don't think the namespace set in first playbook is applicable for second playbook? so may be we should explicitly do 'oc project <namespace>' in the second playbook and get rid of -n <namespace> in every command.

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/tests/roles/prelude_local/tasks/main.yaml#L32

katarimanojk: @fmount yeah i will incorporate the @jistr suggestion in this patch only.

fmount (approved):
/lgtm

jistr:
recheck

PR #953, opened 2025-05-16
Fix hardcoded tripleo passwords file

bogdando:
Note: rdo-jobs, and ci-framework-jobs still do hardcode it to overcloud-passwords.yaml in multiple places and jobs templates.

Jira: [OSPRH-16719](https://issues.redhat.com/browse/OSPRH-16719)


bogdando:
Please note that we can not simply change this, because multiple CI jobs could be not defining tripleo_passwords, nor its ['default'] key, then all those will become borked

ciecierski:
recheck

ciecierski:
recheck

PR #961, opened 2025-06-04
Make tobiko options configurable

eduolivares:
With this change, the following tobiko parameters can be configured per job:
- tobiko_version
- tobiko_conf_file
- tobiko_test_workflow
- tobiko_pytest_addopts

[TOBIKO-154](https://issues.redhat.com//browse/TOBIKO-154)


jistr:
@fyanac can you review please? Adoption core folks aren't as familiar with Tobiko i think.

PR #962, opened 2025-06-04
Use uniquePodNames for cinder and glance pods

holser:
There is a descrepency between adoption and clean deployemnt as clean deployment of OSP18 uses uniquePodNames. This patch changes the behavior to use uniquePodNames: true for glance and cinder. The documentation is changes accordingly

Jira: [OSPRH-15353](https://issues.redhat.com//browse/OSPRH-15353)

### DIFF
--- a/tests/roles/backend_services/defaults/main.yaml
+++ b/tests/roles/backend_services/defaults/main.yaml
@@ -18,6 +18,7 @@ heat_stack_domain_admin_password: ''
 swift_password: ''
 dns_lb_ip: 192.168.122.80
 dns_server_ip: 192.168.122.1
+enable_unique_pod_names: false
 
 dpa_dir: "../.."
 dpa_tests_dir: "{{ dpa_dir }}/tests"
--- a/tests/roles/backend_services/templates/openstack_control_plane.j2
+++ b/tests/roles/backend_services/templates/openstack_control_plane.j2
@@ -14,6 +14,9 @@ spec:
       barbicanKeystoneListener: {}
 
   cinder:
+{% if enable_unique_pod_names %}
+    uniquePodNames: true
+{% endif %}
     enabled: false
     template:
       cinderAPI: {}
@@ -39,6 +42,9 @@ spec:
       replicas: 1
 
   glance:
+{% if enable_unique_pod_names %}
+    uniquePodNames: true
+{% endif %}
     enabled: false
     template:
       glanceAPIs: {}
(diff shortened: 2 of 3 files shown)


jistr:
@fmount @katarimanojk Can you please review if this looks like the way to go?

fmount:
Technically this is set as `False` by default. We only enable it in some circumstances like `unibeta` where both glance and cinder are backed by the same `NFS`, and there might be issues if the same `hostname` is used in several CI deployments.
Being that said having it set to True doesn't hurt, but I'm wondering if you see a particular issue that led to turn this parameter on by default or not.
We can decide to have it on by default, but it won't be the default behavior in production or customers' environments.

fmount (approved):
Just added a comment about the reason behind this parameter.
In theory I would keep it false by default, but it doesn't hurt, so if you think it will make your life easier to match what we do in greenfield unijobs, we can land this change.

fmount:
recheck

fmount:
> If it's not the default for customers and our "default" greenfield testing, then i think we shouldn't enable it in adoption docs and in tests by default. E.g. unibeta adoption topology should enable this similarly like the greenfield unibeta, but i don't think we should make it the documented default.

I'm not entirely sure why in architectures this is enabled by default [1] (for both glance and cinder), maybe this helped for CI purposes, but definitely we do not ask customers to enable this parameter (and is set to `false` by default).

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/lib/control-plane/openstackcontrolplane.yaml#L26

pinikomarov:
I'm not sure I see the advantage of introducing randomness to pod names and given the fact that adoption CI uses pod names in adoption repo :  [pastebin_link - dataplane adoption grep "oc get"](https://privatebin.corp.redhat.com/?ed7693e479a7a0ff#AUHPfXmpJnhNtXfy3pxQH8VJeB9tDqXhJGxCZSP54apS)
I see more work needed to implement side issues from this change.

holser:
@pinikomarov @fmount If you look at [OSPRH-15353](https://issues.redhat.com//browse/OSPRH-15353) you will see that I created openstack environment on unigamma topology. We should either disable it in architecture repository or enable in adoption. Otherwise we'll see discrepancies between greenfield and adopted environments.

jistr (approved):
Thanks, that's also a way to go i think.
/lgtm

ciecierski:
recheck

holser:
recheck

pinikomarov:
looks good to me

holser:
recheck

PR #965, opened 2025-06-09
Manila Netapp adoption

fmount:
This patch introduces the basic flow to run `Manila` adoption when `netapp` is selected as a backend.
The set of tasks added in this change follow this flow:
1. Render/Apply `netapp` `Secret` with the sensitive information required to run `Manila` with this backend
2. Render/Apply `Manila` spec section and patch the existing OpenStack control plane

Jira: https://issues.redhat.com/browse/OSPRH-16787


tests/roles/manila_adoption/tasks/main.yaml
```diff
@@ -36,6 +36,47 @@
...
+- name: Deploy Podified Manila - Netapp
+  when: manila_backend == "netapp"
+  block:
+    # Note: I might need to replace this task with something more complex where
+    # 1. ssh into controller node
+    # 2. extract the netapp config from manila.conf
+    # 3. populate the ansible variables required to build netapp-secret.conf
+    - name: Fail if netapp input are not defined
+      when: |
+        netapp_host is not defined or
```
fmount, line 48: @jistr adding you here because I have a question. 
The idea here is to provide `netapp_*` parameters as input to the playbook and then perform the regular adoption without doing so much complex operations to retrieve that data.
Am I breaking a pattern? 
I see services like `Barbican` do something like [1], but I thought it was better at "job" level to pass this information as input or retrieve it at a different layer.
Is there a pattern that has been established and we want to follow? How the existing unijob deals with secret creation and these kind of information?

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/ef823f2f2c5772fce9e7f04876e0bdfe049deb3a/tests/roles/barbican_adoption/tasks/main.yaml#L6

tests/roles/manila_adoption/tasks/netapp.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+    - name: Apply the rendered Netapp secret in the openstack namespace
+      ansible.builtin.shell: |
+        {{ shell_header }}
+        {{ oc_header }}
+        oc apply -f /tmp/netapp_secret.yaml
+
+    - name: Generate CR config based on the selected backend
+      ansible.builtin.template:
+        src: manila_netapp.yaml.j2
```
fmount, line 40: I might need to fix the template path resolution

tests/roles/manila_adoption/tasks/netapp.yaml
```diff
@@ -0,0 +1,48 @@
...
+        netapp_config.netapp_password is not defined or
+        netapp_config.netapp_vserver is not defined or
+        netapp_config.netapp_server_hostname is not defined
+      ansible.builtin.fail:
+        msg:
+          - 'Missing required Netapp input'
+
+    - name: Render Netapp secret
+      ansible.builtin.template:
+        src: netapp_secret.yaml.j2
```
fmount, line 28: I might need to fix the template path resolution

fmount:
recheck

fmount:
recheck

tosky (approved):

fmount:
> Do you have documentation for Netapp adoption?

https://openstack-k8s-operators.github.io/data-plane-adoption/user/index.html#adopting-the-shared-file-systems-service_adopt-control-plane

PR #967, opened 2025-06-12
Add adoption support for cinder netapp nfs

katarimanojk:
This patch introduces the basic flow to run Cinder adoption when netapp is selected as a backend.
The set of tasks added in this change follow this flow:
1. Get `cinder.conf` from the `Director` deployed controller node
2. Extract all the relevant `netapp` info from the `DEFAULT` section
3. Render/Apply `netapp` secret with the sensitive information required to run cinder volume, and cinder backup with this backend
4. Render/Apply the full Cinder spec section and patch the existing control plane

Jira: https://issues.redhat.com/browse/OSPRH-13615


tests/roles/cinder_adoption/files/cinder_netapp_nfs.yaml
```diff
@@ -0,0 +1,34 @@
+spec:
+  cinder:
+    template:
+      extraMounts:
```
fmount, line 4: -1: no `extraMounts` needed for NFS. Remove this section.

fmount (changes requested):

tests/roles/cinder_adoption/files/cinder_netapp_nfs.yaml
fmount: maybe I'm missing something, but unibeta has those values within the secret [1]

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni02beta/control-plane/service-values.yaml#L21

fmount:
Because we have at least Manila and cinder with the same tasks, I'm wondering if it makes sense to have a generic role to extract the netapp info and create a secret. We could make it generic enough and reuse the role both here and in manila.

tests/roles/cinder_adoption/files/cinder_netapp_nfs.yaml
tosky: Correct, the idea is to pass them as secret. And it's also what the cinder-operator example does https://github.com/openstack-k8s-operators/cinder-operator/blob/main/config/samples/backends/netapp/ontap/nfs/backend.yaml

katarimanojk: @fmount @tosky thanks and i have simplified the code to just use the same secret to have 'nas_host' and 'nas_share_path' . I have removed the extramounts section and the new secret.

I have also tested the latest patch (using a greenfield deployment) and ran only task volume_backend.yaml (from cinder_adoption role) using a test playbook.
I could create a volume on the netapp nfs backend.

tests/roles/cinder_adoption/tasks/volume_backend.yaml
```diff
@@ -4,3 +4,7 @@
     {{ shell_header }}
     {{ oc_header }}
     oc patch openstackcontrolplane openstack --type=merge --patch '{{ cinder_volume_backend_patch }}'
+
+- name: deploy podified Cinder volume with netapp NFS
+  when: cinder_volume_backend == 'ontap-nfs'
```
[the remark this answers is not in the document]
katarimanojk: well..yeah 'cinder_volume_backend' should be set by the users.
considering the netapp deployment guide [1] , i thought ontap-nfs is the right one.

[1] https://netapp-openstack-dev.github.io/openstack-docs/antelope/cinder/configuration/cinder_config_files/section_rhoso18_configuration.html#single-back-end-configuration.

tests/roles/cinder_adoption/templates/cinder-backup-netapp.yaml.j2
```diff
@@ -0,0 +1,13 @@
...
+    enabled: true
+    template:
+      cinderBackup:
+        networkAttachments:
+        - storage
+        replicas: 1
+        customServiceConfig: |
+          [DEFAULT]
+          backup_driver=cinder.backup.drivers.nfs.NFSBackupDriver
+          backup_share = {{ nfs_server }}:{{ nfs_cinder_backup_path }}
```
katarimanojk, line 12: @tosky 
`nfs_cinder_backup_path` currently is set to "" in defaults/main.yaml, here we expect it to be passed as input to adoption playbook or should set it as `/cinder_backup` always ?

whereas `nfs_server` will anyways be extracted from osp17.1 shares.conf here:  tests/roles/cinder_adoption/tasks/cinder_volume_netapp_nfs.yaml : L19

tests/roles/cinder_adoption/templates/cinder-backup-netapp.yaml.j2
```diff
@@ -0,0 +1,13 @@
...
+          backup_mount_point_base = {{ nfs_backup_mount_point_base }}
```
katarimanojk, line 13: @tosky this will be set to "/var/lib/cinder/backup" always in defaults/main.yaml, is it ok?

tests/roles/cinder_adoption/tasks/cinder_volume_netapp_nfs.yaml
```diff
@@ -0,0 +1,50 @@
...
+        ${CONTROLLER1_SCP}:{{ tripleo_nfs_shares_config_path }} {{ nfs_shares_conf_path }}
+
+    - name: Extract Netapp data from cinder.conf
+      ansible.builtin.set_fact:
+        netapp_config: "{{ netapp_config | default({}) | combine({item: lookup('ansible.builtin.ini', item, file=cinder_conf_path, section=cinder_netapp_backend, allow_no_value=True)}) }}"
+      loop: "{{ netapp_vars }}"
+
+    - name: Extract nfs share vars from nfs_shares_conf_path
+      ansible.builtin.set_fact:
+        nfs_server: "{{ (lookup('file', nfs_shares_conf_path)).split(':')[0] | trim }}"
```
tosky, line 19: s/nfs_server/cinder_nfs_server/

tests/roles/cinder_adoption/tasks/cinder_volume_netapp_nfs.yaml
```diff
@@ -0,0 +1,50 @@
...
+        nfs_path: "{{ (lookup('file', nfs_shares_conf_path)).split(':')[1] | trim }}"
```
tosky, line 20: s/nfs_path/cinder_nfs_path/

tests/roles/cinder_adoption/tasks/cinder_volume_netapp_nfs.yaml
```diff
@@ -0,0 +1,50 @@
...
+      ansible.builtin.shell: |
+        {{ shell_header }}
+        {{ controller_ssh }}
+        CONTROLLER1_SCP="{{ controller1_ssh  | regex_replace('^ssh', 'scp')}}"
+        ${CONTROLLER1_SCP}:{{ cinder_tripleo_path }} {{ cinder_conf_path }}
+        ${CONTROLLER1_SCP}:{{ tripleo_nfs_shares_config_path }} {{ nfs_shares_conf_path }}
+
+    - name: Extract Netapp data from cinder.conf
+      ansible.builtin.set_fact:
+        netapp_config: "{{ netapp_config | default({}) | combine({item: lookup('ansible.builtin.ini', item, file=cinder_conf_path, section=cinder_netapp_backend, allow_no_value=True)}) }}"
```
tosky, line 14: These variables are being set as facts. We should not set them as fact if possible, because the ansible namespace is global.
A first step is to use a less generic name. cinder_netapp_config.

tosky (changes requested):

tests/roles/cinder_adoption/tasks/cinder_volume_netapp_nfs.yaml
katarimanojk: done

tests/roles/cinder_adoption/templates/cinder-backup-netapp.yaml.j2
katarimanojk: This is an optional parameter for swift/ceph backup driver whereas it mandatory for nfs based backup driver. 

It indicates the base directory where NFS shares for backups (backup_share) are mounted. i.e backup driver needs to mount the nfs filesystem to read/write backup data.

Value assigned to it can be any custom path (but the directory should exist), i think it should default to `/var/lib/cinder/backup` and user can always override it if needed.

tests/roles/cinder_adoption/defaults/main.yaml
```diff
@@ -98,3 +98,28 @@ cinder_backup_swift_patch: |
...
+cinder_tripleo_path: "/var/lib/config-data/puppet-generated/cinder/etc/cinder/cinder.conf"
+
+# nfs shares path used in the tripleo for cinder netapp nfs backend
+tripleo_nfs_shares_config_path: "/var/lib/config-data/cinder/etc/cinder/shares.conf"
+
+# The path where we temporarily save cinder.conf fetched from TripleO controller node
+cinder_conf_path: "/tmp/cinder.conf"
+
+# The path where we temporarily save shares.conf fetched from TripleO controller node
+nfs_shares_conf_path: "/tmp/shares.conf"
```
tosky, line 112: shouldn't this variable be prefixed by cinder_ again?

tests/roles/cinder_adoption/defaults/main.yaml
```diff
@@ -98,3 +98,28 @@ cinder_backup_swift_patch: |
...
+
+# The backend name generated by puppet-tripleo
+cinder_netapp_backend: "tripleo_netapp"
+
+# List of netapp credentials to be captured from tripleo cinder.conf
+cvol_netapp_vars:
```
tosky, line 118: any reason for not using a longer cinder_vol_... ?

tests/roles/cinder_adoption/defaults/main.yaml
```diff
@@ -98,3 +98,28 @@ cinder_backup_swift_patch: |
             backup_swift_auth = per_user
             keystone_catalog_info = identity:Identity Service:publicURL
             swift_catalog_info = object-store:swift:publicURL
+
+# The path where we find cinder.conf generated by puppet-tripleo
+cinder_tripleo_path: "/var/lib/config-data/puppet-generated/cinder/etc/cinder/cinder.conf"
+
+# nfs shares path used in the tripleo for cinder netapp nfs backend
+tripleo_nfs_shares_config_path: "/var/lib/config-data/cinder/etc/cinder/shares.conf"
```
tosky, line 106: see below, I would add a cinder in the variable name too

tests/roles/cinder_adoption/tasks/cinder_backup_netapp_nfs.yaml
```diff
@@ -0,0 +1,24 @@
+- name: Deploy Podified Cinder backup - Netapp
+  block:
+    - name: Extract Netapp c-bak data from cinder.conf
+      ansible.builtin.set_fact:
+        cbak_nfs_backup_share: "{{ lookup('ansible.builtin.ini', 'backup_share' , file=cinder_conf_path, section='DEFAULT', allow_no_value=True) }}"
```
tosky, line 5: cinder_backup instead of cbak? Do we have limits for variable names in ansible?

tests/roles/cinder_adoption/defaults/main.yaml
katarimanojk: replaced cvol with cinder_volume at all places in the latest patch.

tosky (approved):

fmount (approved):
/lgtm

PR #972, opened 2025-06-16
Get cinder pod name

holser:
As [1] was merged, testproject failed as cinder-api pod was hardcoded to cinder-api-0 which doesn't work where uniquepodnames are turned on. This patch gets pod name first to access cinder services later on

Jira: [OSPRH-15353](https://issues.redhat.com//browse/OSPRH-15353)

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/pull/962

### DIFF
--- a/tests/roles/cinder_adoption/tasks/main.yaml
+++ b/tests/roles/cinder_adoption/tasks/main.yaml
@@ -14,6 +14,13 @@
   retries: 60
   delay: 2
 
+- name: Get cinder-api pod name
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc get pods -l component=cinder-api -o custom-columns="NAME:.metadata.name" --no-headers | head -1
+  register: cinder_api_pod
+
 - name: check that Cinder is reachable and its endpoints are defined
   ansible.builtin.shell: |
     {{ shell_header }}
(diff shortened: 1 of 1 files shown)


tests/roles/cinder_adoption/tasks/main.yaml
```diff
@@ -14,6 +14,13 @@
   retries: 60
   delay: 2
 
+- name: Get cinder-api pod name
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc get pod -l component=cinder-api -o jsonpath="{.items[*].metadata.name}"
```
fmount, line 21: -1: This is not going to work outside of `standalone/crc` scenario where `replicas: 3`.
In that case, `cinder_api_pod.stdoutoutlines` is made by > 1 element (e.g. `cinder-5aa52-api-0 cinder-5aa52-api-1 cinder-5aa52-api-2`, so I assume it is sufficient to take the `first` of the list.

fmount (changes requested):

tests/roles/cinder_adoption/tasks/main.yaml
```diff
@@ -14,6 +14,13 @@
...
   delay: 2
 
+- name: Get cinder-api pod name
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+    oc get pods -l component=cinder-api -o custom-columns="NAME:.metadata.name" --no-headers | head -1
+  register: cinder_api_pod
+
 - name: check that Cinder is reachable and its endpoints are defined
```
fmount, line 24: we should have a `when` condition to make sure `cinder_api_pod.stdout | length`  is > 0, but I assume if we reach this point we already successfully passed L7 and `condition=Ready` ensures cinder-api(s) are up and running.

works for me.

PR #975, opened 2025-06-19
Reduce ansible verbosity on dataplane

holser:
verbosity 3 is very high value, whic creates a pressure on disk system. As a result we have sporadic failures or timeouts. This patch reduces verbosity to normal which can be overriden on job level when it's needed.


jistr:
To unblock CI
/approve

PR #981, opened 2025-06-26
cinder, manila: various fixes (undef vars, file perms, extra net)

tosky:
cinder: fix the reference to access a controller
controller_ssh by itself is undefined. But it is not used either, so just remove it.

manila: fix the reference to access a controller
controller_ssh by itself is undefined for the netapp codepath (it is defined when using the cephfs backend).
But it is not used either in the netapp codepath, so just remove it.

manila, cinder: fix permissions when fetching conf files
When fetching the configuration files of cinder and manila through scp, make sure the files are given wider permissions so that the subsequent tasks don't need to be run with become: yes.

cinder: remove an unneeded storageMgmt network attachment
Only storage is needed.


holser (approved):
/lgtm

PR #983, opened 2025-06-26
test_minimal: call manila_adoption if a backend is set

tosky:
It is already included by test_with_ceph and test_rollback_with_ceph (where the backend is explicitly set), so ensure it is called only when manila is requested, following the pattern established for other services.

### DIFF
--- a/tests/playbooks/test_minimal.yaml
+++ b/tests/playbooks/test_minimal.yaml
@@ -71,6 +71,10 @@
       tags:
         - autoscaling_adoption
       when: telemetry_adoption|default(true)
+    - role: manila_adoption
+      tags:
+        - manila_adoption
+      when: manila_backend|default('')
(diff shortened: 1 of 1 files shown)


fmount (approved):
/lgtm

tosky:
recheck

tests/playbooks/test_minimal.yaml
```diff
@@ -71,6 +71,10 @@
       tags:
         - autoscaling_adoption
       when: telemetry_adoption|default(true)
+    - role: manila_adoption
+      tags:
+        - manila_adoption
+      when: manila_backend|default('')
```
fmount, line 77: @tosky an alternative would be to have `manila_backend | default('') | length > 0`, so we can actually run the code if there's a backend set as extra_vars.

PR #993, opened 2025-07-04
Refactor pcp_cleanup role for better readility

holser:
* Restructured bash for readability
* fix grammar in documentation
* added a check on RENAMED_CELLS just not to fail when variable is not defined


bogdando (changes requested):
I recommend to ensure RENAMED_CELLS is always set (by default to cell1), instead of adding a non empty (unexpected case) check

PR #995, opened 2025-07-04
uni07eta: base 17.1 tripleo deployment files

tosky:
The patch can deploy an almost working OSP 17.1. It just requires an additional configuration/environment file with the private cinder/iSCSI configuration (when using ci-framework, that would be cifmw_adoption_osp_deploy_overcloud_extra_args).


tosky:
recheck

PR #997, opened 2025-07-05
Fix Glance NFS adoption doc

fmount:
This patch introduces a small fix in the `Image Service` adoption documentation that is reflected in [1]. Even when `Glance` is disabled, because the `ctlplane` creates a `Glance` object that by default is `type: split`, we can't patch a default `API` anymore. For this reason, the documentation (and the automation) is updated to create a new `nfs` `glanceAPI` and register it and keystone as the default endpoint. An additional step is required to decommission the old one, which would be useless in this case.

Jira: https://issues.redhat.com/browse/OSPRH-15612

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/pull/996


tosky (approved):
It matches https://github.com/openstack-k8s-operators/data-plane-adoption/pull/996/, and it reflects the reality.

PR #1005, opened 2025-07-09
used cursor to fix links to comply with DITA

klgill:


tosky:
as this change apparently used an LLM, shouldn't this be documented in the commit message?

tosky:
> @tosky I'm not sure what the best practice is. I reused a script that another writer on my team previously generated from Cursor. The docs team is migrating to DITA by the end of the year, and updating the link format is part of the preparation for that.

If this used a script (leaving aside how it was generated), I would argue that the "cursor" part in the commit message is not relevant for the change itself.

klgill:
I got news that we do not have to convert xrefs "if it causes too much trouble" because there will be a way to do this with the migration tool. Since the links did not work after I converted them to the new format upstream, I'm closing this PR.

PR #1012, opened 2025-07-16
Per backend glance adoption refactor

fmount:
This patch represents a refactor of the glance adoption code that is now run depending on the backend we set in advance.


tosky:
Interesting, so basically glance_local.yaml was never used. Do we need to properly built that codepath, or do we always require a proper backend (which includes file-with-NFS aka glance_nfs.yaml?)

fmount:
> Interesting, so basically glance_local.yaml was never used. Do we need to properly built that codepath, or do we always require a proper backend (which includes file-with-NFS aka glance_nfs.yaml?)

It was there for historical reasons, but it was never used because `File` is not supported (and we moved to swift as default backend for a != Ceph scenario).
We always require a proper backend, otherwise we need to keep `replicas: 0` as per our greenfield documentation. In an adoption scenario we always have a backend in the source cloud.

ciecierski (approved):

katarimanojk (approved):
Code looks clean now, thanks Francesco for refactoring

PR #1015, opened 2025-07-17
Idempotent pre_launch_ironic.bash - cleanup resources to allow re-runs

dsariel:
Check whether the flavor already exists before trying to create it to allow running over and over again.


tests/roles/development_environment/files/pre_launch_ironic.bash
```diff
@@ -1,7 +1,14 @@
-set -e
+set -o pipefail
```
ciecierski, line 1: set -e
set -o pipefail

PR #1018, opened 2025-07-24
cinder netapp+doc: use the source configuration group name

tosky:
Make sure to use the configuration group name of the source cluster for the netapp/NFS backend.
This means generalizing a bit one of the templates, and converting a static configuration file to a template.

documentation: Point out that the driver configuration group name of the new cluster should match the corresponding value of the source cluster.


tosky:
Tested internally

fmount (approved):
/lgtm

PR #1019, opened 2025-07-25
ovn_adoption: don't hardcode quay.io as ovn_base image source

tosky:
Other test roles use/depend on variables like container_registry, container_namespace and container_tag to allow users to override the source of the containers.
So make use of them here too for consistency.
The default behavior is unchanged.


tosky:
Internally tested (ask more for details in case), and the variables were successfully overridden. And the CI jobs executed here on RDO shows that the default value is still used as before.

PR #1020, opened 2025-07-26
cinder adoption: generalize the netapp code (also iSCSI)

tosky:
- reduce the usage of set_facts by defining local variables (inspired by a recent manila change);
- generalize the netapp/NFS playbook to support also iSCSI in addition to NFS. The code is mostly the same, NFS just just needs to check an additional file and a protocol-specific template needs to be used.
- track/source also netapp_pool_name_search_pattern (only if defined)


fmount:
recheck

PR #1034, opened 2025-08-19
doc: add section numbers

tosky:
So that some people can use them as easier to spot mental anchors.

### DIFF
--- a/docs_user/main.adoc
+++ b/docs_user/main.adoc
@@ -5,6 +5,7 @@
 :compat-mode:
 :doctype: book
 :context: adoption-rhoso
+:sectnums:
 
 include::adoption-attributes.adoc[]
 


klgill:
Is this patch just so that the numbers render in the preview, e.g., 1, 1.1, 1.2, 1.3, etc...?
If so, this is fine, but it's not necessary downstream. The section numbers already render when we publish to the customer portal.

tosky:
> Is this patch just so that the numbers render in the preview, e.g., 1, 1.1, 1.2, 1.3, etc...? If so, this is fine, but it's not necessary downstream. The section numbers already render when we publish to the customer portal.

That's fine but as long as users can access the rendered content from https://openstack-k8s-operators.github.io/data-plane-adoption/, I think they (we?) deserve to have section numbers. They make it easier (to me, at least) to browse the documentation.

klgill (approved):

PR #1036, opened 2025-08-19
support ext ceph in ceph_backend_configuration

katarimanojk:
with external ceph, ceph_backend_configuration should fetch ceph details from ceph nodes instead of controller nodes, this patch implements it.

Jira: https://issues.redhat.com/browse/OSPRH-5712

### DIFF
--- a/tests/roles/ceph_backend_configuration/tasks/main.yaml
+++ b/tests/roles/ceph_backend_configuration/tasks/main.yaml
@@ -2,7 +2,9 @@
   no_log: "{{ use_no_log }}"
   ansible.builtin.set_fact:
     ceph_backend_configuration_shell_vars: |
-      CEPH_SSH="{{ controller1_ssh }}"
+      # if tripleo uses external ceph, ssh to ceph nodes otherwise controller nodes
+      # for external ceph, external_ceph and ceph1_ssh vars should be defined
+      CEPH_SSH="{{ external_ceph | default(false) | ternary(ceph1_ssh, controller1_ssh) }}"
       CEPH_KEY=$($CEPH_SSH "cat /etc/ceph/ceph.client.openstack.keyring | base64 -w 0")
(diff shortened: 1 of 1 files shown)


tests/roles/ceph_backend_configuration/tasks/main.yaml
```diff
@@ -2,7 +2,9 @@
   no_log: "{{ use_no_log }}"
   ansible.builtin.set_fact:
     ceph_backend_configuration_shell_vars: |
-      CEPH_SSH="{{ controller1_ssh }}"
+      # if tripleo uses external ceph, ssh to ceph nodes otherwise controller nodes
+      # for external ceph, external_ceph and ceph1_ssh vars should be defined
+      CEPH_SSH="{{ external_ceph | default(false) | ternary(ceph1_ssh, controller1_ssh) }}"
```
fmount, line 7: This is useful for unidelta. In case of unidelta ceph is externally deployed, and we can reach it using `ceph1_ssh` instead of looking at the controller node that has no mons.
Backward compatibility is preserved by using ternary filter, and `external_ceph` is passed as an override only for unidelta based jobs.
I think I'm ok with this approach. I'll let @jistr @ciecierski comment on it and land it.

ciecierski: Change looks fine to me. Do we have user-facing doc for externally deployed ceph explaining this case?

fmount: Technically the same documentation applies, the only difference is the location where the ceph config files are found. For director deployed ceph we have them in the openstack controller nodes, but if ceph is externally deployed, we need to connect to those nodes. The rest of the assumptions are the same.

ciecierski: @katarimanojk Thanks for pointer to documentation.

katarimanojk:
@jistr can you pls review and merge this PR if you think we are good.

PR #1040, opened 2025-08-26
Include ipv6 in dataplane repo

fyanac:
Ticket: [OSPRH-17670](https://issues.redhat.com//browse/OSPRH-17670)

Assisted-By: Cursor-claude-4-sonnet

### DIFF
--- a/tests/roles/common_defaults/defaults/main.yaml
+++ b/tests/roles/common_defaults/defaults/main.yaml
@@ -2,6 +2,11 @@
 # sensitive data.
 use_no_log: false
 
+# Network configuration for IPv4/IPv6 support
+ipv6_enabled: false
+internalapi_prefix: "172.17.0"
+internalapi_prefix_ipv6: "2620:cf:cf:bbbb"
+
 # Whether the adopted node will host compute services
 compute_adoption: true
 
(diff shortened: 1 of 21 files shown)


tests/roles/ovn_adoption/defaults/main.yaml
```diff
@@ -6,6 +6,13 @@ storage_class_name: local-storage
 storage_reclaim_policy: delete
 ironic_adoption: false
 
+# Network configuration for IPv4/IPv6 support
+ipv6_enabled: false
+internalapi_cidr_ipv4: "172.17.1.0/24"
+internalapi_cidr_ipv6: "2620:cf:cf:bbbb::/64"
+ctlplane_prefix: "192.168.122"
+ctlplane_ipv6_prefix: "2620:cf:cf:aaaa"
```
jistr, line 14: Since we're introducing new vars, we should probably make the naming conventions consistent with each other (the vars above have _ipv6 both at the end and in the middle). And also perhaps make them consistent with architecture repo conventions (where possible/reasonable). It seems the architecture repo uses `_v4` and `_v6` suffixes at the end of the variable name, e.g.:

https://github.com/openstack-k8s-operators/architecture/blob/main/automation/net-env/uni01alpha.yaml 
https://github.com/openstack-k8s-operators/architecture/blob/main/automation/net-env/uni04delta-ipv6.yaml 

Also it would be good to base the "cidr" vars on the "prefix" vars, so that when prefix is set to something custom, the cidr var is adjusted automatically.

I'm thinking of something like:

```
ipv6_enabled: false
internalapi_prefix_v4: "172.17.0"
internalapi_prefix_v6: "2620:cf:cf:bbbb"
ctlplane_prefix_v4: "192.168.122"
ctlplane_prefix_v6: "2620:cf:cf:aaaa"
internalapi_cidr_v4: "{{ internalapi_prefix_v4 }}.0/24"
internalapi_cidr_v6: "{{ internalapi_prefix_v6 }}::/64"
```

However i see the `internalapi_prefix` is already in use so i wonder if we want to just stick to this instead:

```
ipv6_enabled: false
internalapi_prefix: "172.17.0"
internalapi_prefix_v6: "2620:cf:cf:bbbb"
ctlplane_prefix: "192.168.122"
ctlplane_prefix_v6: "2620:cf:cf:aaaa"
internalapi_cidr: "{{ internalapi_prefix_v4 }}.0/24"
internalapi_cidr_v6: "{{ internalapi_prefix_v6 }}::/64"
```

Or try to rename the variables to match the architecture convention. Perhaps if we want to do the _v4 renaming (which i think would be nice), we should merge this without the _v4 renaming first, and make the renaming a separate epic? It may need edits in other repos as well, if the CI already sets the existing prefix/cidr vars.

tests/roles/common_defaults/defaults/main.yaml
```diff
@@ -2,6 +2,11 @@
 # sensitive data.
 use_no_log: false
 
+# Network configuration for IPv4/IPv6 support
+ipv6_enabled: false
+internalapi_prefix: "172.17.0"
```
jistr, line 7: This uses 172.17.0 while below ovn_adoption role we use 172.17.1.

tests/roles/ovn_adoption/defaults/main.yaml
fyanac: Hi @jistr 
I think the _**internalapi_cidr_v4: "{{ internalapi_prefix_v4 }}.0/24"**_ it wouldn't apply here, because this part is the case when we have 2 different internal networks and allow the traffic to the additional network, it's related to this patch: https://github.com/openstack-k8s-operators/data-plane-adoption/commit/a396847fdbd4d5845129516156924ab0534eb2f5

PR #1042, opened 2025-08-29
Fix manila adoption for cephnfs

katarimanojk:
- Add missing CONTROLLER1_SSH variable definition to prevent unbound variable errors
    - Fix awk command quoting for proper SSH execution
    - Escape inner quotes and dollar signs in the awk command
    
    This resolves the syntax error when executing the awk command over SSH
    to extract the Ganesha Bind_Addr from the configuration file.


fmount:
makes sense, thank you @katarimanojk

PR #1045, opened 2025-09-02
TLS-e migration fixes

xek:
Move IPA_SSH to prerequisites, fix ordering,
use sudo admin user instead of root.

Jira: https://issues.redhat.com/browse/OSPRH-19131


klgill (approved):
LGTM from a docs perspective.

PR #1047, opened 2025-09-04
Do not always create a manila share type

fmount:
In the adoption context based on unijobs, we assume that a workload exists in the source cloud. Therefore, we do not need to create a new manila share type in the adopted environment, because it wouldn't be realistic.
This patch allows to optionally perform this task, so we can still keep it in standalone, but we don't need to execute it for unijobs.

Jira: https://issues.redhat.com/browse/OSPRH-5712


katarimanojk (approved):
/lgtm

fmount:
recheck

fmount:
Ok we saw this patch working and manila is adopted and able to re-export the share using the new ceph-nfs cluster:

```bash
sh-5.1$ openstack share export location list 41ff8590-a7cb-45a9-9544-a6017488b2e3
+--------------------------------------+--------------------------------------------------------------------------------------------------------------------+-----------+
| ID                                   | Path                                                                                                               | Preferred |
+--------------------------------------+--------------------------------------------------------------------------------------------------------------------+-----------+
| 2d54c5d2-fc29-463f-9518-4136a5ac18b3 | [2620:cf:cf:cf02::227]:/volumes/_nogroup/fec6e542-54ac-4719-9351-f317cb79a550/50624a2a-b34b-469e-bf8e-63fb82f5b244 | False     |
| 281a85a3-45a4-4072-a2d9-94843dd5f71d | [2620:cf:cf:aaaa::8]:/volumes/_nogroup/fec6e542-54ac-4719-9351-f317cb79a550/50624a2a-b34b-469e-bf8e-63fb82f5b244   | True      |
+--------------------------------------+--------------------------------------------------------------------------------------------------------------------+-----------+
```

Based on these results, can we land this patch?

PR #1051, opened 2025-09-08
Add default manila_backend value when evaluating pcs constraints

fmount:
Jobs **without ceph** might not have knowledge about `manila_backend` variable if not explicitly passed. Because the same task is executed in several scenarios, we need to make sure that `manila_backend` does not point to an undefined variable.

Jira: https://issues.redhat.com/browse/OSPCIX-1034


fmount:
Test in progress via: https://review.rdoproject.org/r/c/testproject/+/58182

katarimanojk (approved):
/lgtm

PR #1055, opened 2025-09-14
Remove cell name loop since ospdo doesn't support multi cell currently

pinikomarov:
Also the loop is redundant in that case and might confuse customers

Jira: https://issues.redhat.com/browse/OSPRH-19289


klgill (approved):
Both the director operator and regular adoption previews look good.

holser (approved):

PR #1072, opened 2025-09-30
Increase delay time to 5 seconds

danpawlik:
There is no need to make so often queries if the service is ready. It might raise complication on some CI host, that does not have enough power.

### DIFF
--- a/tests/roles/autoscaling_adoption/defaults/main.yaml
+++ b/tests/roles/autoscaling_adoption/defaults/main.yaml
@@ -1,0 +1,2 @@
+---
+autoscaling_retry_delay: 5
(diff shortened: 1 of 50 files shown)


tests/roles/pcp_cleanup/defaults/main.yaml
```diff
@@ -8,3 +8,4 @@ reset_crc_storage: false
 # Whether 'make standalone_revert' should be run before
 # running the test suite.
 standalone_revert_enabled: false
+delay: 5
```
jistr, line 11: In addition to breaking the lint, this is unused, i think it can be dropped. Other than that LGTM

holser (changes requested):

jistr:
I take Sergii's edit as 2nd lgtm, merging.
/lgtm
/approve

PR #1077, opened 2025-10-02
Increase osp deployment timeout for uni02beta

jamepark4:
Downstream uni02beta jobs are intermittently hitting the timeout threshold during osp deployment.  Increasing the timeout to allow for completion.


sdatko (approved):
/lgtm
/approve

PR #1186, opened 2025-12-19
Configure OVN SB direct IPs for DCN nodesets

fultonj:
Problem:
DCN compute nodes' OVN Controller agents cannot connect to the OVN SB database
because the default ovncontroller-config ConfigMap uses Kubernetes ClusterIP
(tcp:ovsdbserver-sb.openstack.svc:6642) which is not routable from external
EDPM nodes on different network segments.

While DNS resolution works (via dnsmasq at 192.168.122.80), the resolved
ClusterIP cannot be reached from DCN sites which are on different internalapi
subnets (172.17.10.x for dcn1, 172.17.20.x for dcn2 vs central's 172.17.0.x).

This causes port binding failures when launching VMs in DCN availability zones:
  "Binding failed for port, please check neutron logs for more information"

Evidence:
- Central compute OVN Controller agents: Connected and working (`:-)` status)
- DCN compute OVN Controller agents: NOT registered in OVN SB database
- `ovn-sbctl show` shows only central computes and gateway, no DCN chassis

Root Cause:
Setting edpm_ovn_dbs variable is insufficient because the edpm_ovn role loads
ovncontroller-config ConfigMap data which overrides the ovn-remote setting.
The default ConfigMap (created by OVNDBCluster operator) uses ClusterIP.

Solution:
1. Retrieve OVN SB internalapi IPs from pod annotations
2. Create DCN-specific ConfigMap (ovncontroller-config-dcn) with direct IPs
3. Create DCN-specific DataPlaneService (ovn-dcn) referencing this ConfigMap
4. Patch dcn1/dcn2 nodesets to use ovn-dcn service instead of ovn

This ensures DCN nodes connect to OVN SB via routable internalapi IPs:
  tcp:172.17.0.34:6642,tcp:172.17.0.36:6642,tcp:172.17.0.35:6642

Co-Authored-By: Claude <noreply@anthropic.com>
Signed-off-by: John Fulton <fulton@redhat.com>

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/main.yaml
+++ b/tests/roles/dataplane_adoption/tasks/main.yaml
@@ -43,6 +43,69 @@
     EOF
   when: configure_ipam | bool
 
+# DCN requires routes in the internalapi NAD so OVN SB pods can reach DCN compute nodes.
+# Macvlan pods don't inherit routes from the node interface, so we add them to the NAD.
+- name: Patch internalapi NAD with routes to DCN subnets
+  when: edpm_nodes_dcn1 is defined or edpm_nodes_dcn2 is defined
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ shell_header }}
+    {{ oc_header }}
+
+    # Get current NAD config
+    CURRENT_CONFIG=$(oc get net-attach-def internalapi -n openstack -o jsonpath='{.spec.config}')
+
+    # Check if routes already exist
+    if echo "$CURRENT_CONFIG" | grep -q '"routes"'; then
+      echo "Routes already exist in internalapi NAD, skipping"
+      exit 0
+    fi
+
+    # Build routes array for DCN subnets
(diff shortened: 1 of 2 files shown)


tests/roles/dataplane_adoption/tasks/main.yaml
```diff
@@ -43,6 +43,69 @@
...
+
+    # Check if routes already exist
+    if echo "$CURRENT_CONFIG" | grep -q '"routes"'; then
+      echo "Routes already exist in internalapi NAD, skipping"
+      exit 0
+    fi
+
+    # Build routes array for DCN subnets
+    # DCN1: 172.17.10.0/24, DCN2: 172.17.20.0/24, gateway: 172.17.0.1
+    ROUTES='[{"dst":"172.17.10.0/24","gw":"172.17.0.1"},{"dst":"172.17.20.0/24","gw":"172.17.0.1"}]'
```
olliewalsh: could just add a route for 172.17.0.0/16?

olliewalsh (commented):
LGTM but cannot verify as routing just worked via the default gateway on my env (centos-stream-9)

PR #1288, opened 2026-03-06
[dataplane_adoption] Do not override images in the nodeset

karelyatin:
OpenstackVersion CR have info for all the images, we should utilize that instead of overriding in the nodeset. Any image overrides in nodeset has highest priority so any handling/overrides with OpenstackVersion CR is not used.

Also added edpmNodeExporterImage which was missing in
OpenstackVersion template.


karelyatin:
recheck barbican fixed

jistr (approved):
I'm wondering whether this one should have a TP but on high level it seems correct to me to get rid of this logic.

/lgtm

ciecierski (changes requested):
None of edpm images removed in here are present as override in openstack_version.j2 template. Let's take edpm_frr_image as a first example, which is required for BGP, we not only ovveriding it for test run, but also pointing customer in docs to this override.

ciecierski (approved):

PR #1361, opened 2026-04-10
mariadb_copy: replace sleep workarounds with persistent pod retry pattern

ciecierski:
## Summary

- Replace the temporary `mariadb_client_timeout` sleep and `wait bgp` pause workarounds in `mariadb_copy` with a retry loop that polls the podified MariaDB via the already-running `mariadb-copy-data` pod, eliminating the race condition where commands were executed before the network was fully programmed (same approach used by `get_services_configuration`).
- Align `pre_checks.bash` with the docs by replacing the ephemeral `oc run mariadb-client --rm ... sleep && mysql` pattern with `oc rsh mariadb-copy-data`, restoring docs-tests parity.
- Add a corresponding "wait for podified MariaDB reachable" step to the docs procedure with a BGP/IPv6 NOTE.
- Remove the `mariadb_client_timeout` variable from `common_defaults` as it has no remaining references.

Resolves: [OSPRH-27386](https://redhat.atlassian.net/browse/OSPRH-27386)

### DIFF
--- a/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
+++ b/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
@@ -223,6 +223,22 @@ for CELL in $(echo $CELLS); do
 done
 ----
 
+. Wait for the `mariadb-copy-data` pod to reach the podified MariaDB:
++
+----
+$ oc rsh mariadb-copy-data mysql -rsh "${PODIFIED_MARIADB_IP['super']}" \
+  -uroot -p"${PODIFIED_DB_ROOT_PASSWORD['super']}" -e 'select 1;'
+----
++
+[NOTE]
+====
+For BGP-enabled environments, this command might take a few moments to succeed while BGP routes are advertised and propagated through the network. The `mariadb-copy-data` pod needs to receive the route to the podified MariaDB IP address through BGP before it can establish a connection. If the command fails, wait a few seconds and retry. The connection should succeed once the BGP route advertisement is complete.
+
+For IPv6 environments, this command might take a few moments to succeed while the network IPv6 stack completes its setup. If the command fails, wait a few seconds and retry.
+
+For standard deployments, such as non-BGP deployments or IPv4 deployments, this command should succeed immediately.
+====
+
 . Test the connection to the control plane upcall and cells databases:
 +
 ----
--- a/tests/roles/mariadb_copy/tasks/main.yaml
+++ b/tests/roles/mariadb_copy/tasks/main.yaml
@@ -73,16 +73,18 @@
   retries: 25
   delay: "{{ mariadb_retry_delay }}"
 
-# MariaDB client connection timeout in seconds
-# Related to OSPRH-18618
-- name: Add sleep time
-  ansible.builtin.pause:
-    seconds: "{{ mariadb_client_timeout }}"
-
-- name: wait bgp
-  when: bgp
-  ansible.builtin.pause:
-    seconds: 60
+- name: wait until podified MariaDB is reachable
+  no_log: "{{ use_no_log }}"
+  ansible.builtin.shell: |
+    {{ oc_header }}
+    {{ mariadb_copy_shell_vars_dst }}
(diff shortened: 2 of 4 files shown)


tests/roles/mariadb_copy/tasks/main.yaml
```diff
@@ -73,16 +73,18 @@
...
-# MariaDB client connection timeout in seconds
-# Related to OSPRH-18618
-- name: Add sleep time
-  ansible.builtin.pause:
-    seconds: "{{ mariadb_client_timeout }}"
-
-- name: wait bgp
-  when: bgp
-  ansible.builtin.pause:
-    seconds: 60
```
jistr, line 85: I'm wondering about the BGP scenario, if the wait for **podified** MariaDB is able to replace this waiting. My understanding was that when connecting to podified, the traffic goes internally in OpenShift and doesn't use BGP. Wouldn't BGP be used only when connecting to **source** MariaDB, and so we should be connecting either to that one, or to both? cc @eduolivares

eduolivares: I'm not sure if I'm misunderstanding.

BGP is not involved when we access the mariadb-copy-data pod from the controller-0 (i.e. when we run `oc rsh mariadb-copy-data`).
However, BGP routes need to be advertised (and that may take a few seconds) for the connectivity between the mariadb-copy-data pod and `${PODIFIED_MARIADB_IP['super']}`.

jistr: Ack then it was me misunderstanding where BGP is used and the code is likely correct :)

eduolivares (approved):

ciecierski:
Still busy testing this change in downstream

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
```diff
@@ -223,6 +223,22 @@ for CELL in $(echo $CELLS); do
 done
 ----
 
+. Wait for the `mariadb-copy-data` pod to be able to reach the podified MariaDB:
```
klgill:
```suggestion
. Wait for the `mariadb-copy-data` pod to reach the podified MariaDB:
```

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
```diff
@@ -223,6 +223,22 @@ for CELL in $(echo $CELLS); do
...
++
+----
+$ oc rsh mariadb-copy-data mysql -rsh "${PODIFIED_MARIADB_IP['super']}" \
+  -uroot -p"${PODIFIED_DB_ROOT_PASSWORD['super']}" -e 'select 1;'
+----
++
+[NOTE]
+====
+For BGP-enabled environments, this command might take a few moments to succeed while BGP routes are advertised and propagated through the network. The `mariadb-copy-data` pod needs to receive the route to the podified MariaDB IP address through BGP before it can establish a connection. If the command fails, wait a few seconds and retry. The connection should succeed once the BGP route advertisement is complete.
+
```
klgill, line 236: What should the customer do if the command fails on the next try?

klgill (changes requested):
I have one minor edit and a question.

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
ciecierski: There is sentence: `If the command fails, wait a few seconds and retry`. If there is still problem I don't think there is single command to solve it. It depends on situation and should be troubleshooted by end-user.

docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
ciecierski: done

klgill (approved):
/lgtm

fyanac (approved):

jistr (approved):
/approve

PR #1362, opened 2026-04-13
Fix possible race condition in swift_migration check

cschwede:
Updated rings were only pushed to the first controller node, resulting in a data movement loop: controller 1 pushed data to OpenShift and removed it locally, while controller 2 and 3 were detecting missing data
on controller 1 (because they were still using old rings) and replicating back to controller 1.

Update both the docs and the test role to distribute rings to all three controller nodes and restart Swift services on each.

Also add the -ignore_readdir_race option to the find command that checks for remaining .data/.db files on the standalone node. The replicator may  remove empty directories while find is iterating, causing spurious errors.

JIRA: [OSPCIX-1318](https://redhat.atlassian.net/browse/OSPCIX-1318)


cschwede:
recheck

cschwede:
recheck

ciecierski:
Multinode job test-project started here:
https://review.rdoproject.org/r/c/testproject/+/58791/3/.zuul.yaml

cschwede:
> Multinode job test-project started here: https://review.rdoproject.org/r/c/testproject/+/58791/3/.zuul.yaml

Output looks much better, the drain check succeeded after 50 seconds.

ciecierski (approved):

jistr (approved):
/lgtm
/approve

PR #1367, opened 2026-04-17
cinder:  backend_host and volume_backend_name from 17.1

tosky:
This change makes sure that
- makes sure that volume_backend_name has the same value as the overall section, matching the value from 17.1;
- adds backend_host=hostgroup to the templates for the cinder volume configuration which were missing it.

If these keys are not assigned the same value as in 17.1, cinder won't be able to manage the volumes created before the adoption, because they will be assigned a new value for os-vol-host-attr:host, and the value for volume_backend_name set so far did not match the old value.


fmount (approved):
/lgtm

jistr (approved):
/lgtm
/approve

PR #1371, opened 2026-04-28
Fix neutron-adoption with ironic

gthiemonge:
Fix a conversion issue in neutron-adoption, the task triggers a bug:

[..]
oc patch openstackcontrolplane openstack --type=merge --patch '{'spec': {'neutron': {'template': {'ml2MechanismDrivers': ['ovn', 'baremetal']}}}}' Error from server (BadRequest): error decoding patch: invalid character 's' looking for beginning of object key string [..]

ironic_ml2_baremetal_patch is an ansible object, it needs to be converted to a json string before it's used in an oc patch command

### DIFF
--- a/tests/roles/neutron_adoption/tasks/main.yaml
+++ b/tests/roles/neutron_adoption/tasks/main.yaml
@@ -4,12 +4,12 @@
     {{ oc_header }}
     oc patch openstackcontrolplane openstack --type=merge --patch '{{ neutron_config_patch }}'
 
-- name: Patch neutron ml2MechanismDrivers to inlcude ml2 baremetal
+- name: Patch neutron ml2MechanismDrivers to include ml2 baremetal
   when: ironic_adoption|bool
   ansible.builtin.shell: |
     {{ shell_header }}
     {{ oc_header }}
-    oc patch openstackcontrolplane openstack --type=merge --patch '{{ ironic_ml2_baremetal_patch }}'
+    oc patch openstackcontrolplane openstack --type=merge --patch '{{ ironic_ml2_baremetal_patch | to_json }}'
 
 - name: wait for Neutron to start up
   ansible.builtin.shell: |


gthiemonge:
@hjensas ^ FYI I'm testing it in a D/S CI job

jistr:
I guess it's a problem of single quotes within single quotes, and `to_json` makes double quotes instead. (`json.dumps` does that and according to docs, the filter uses that function under the hood)

So this is not resilient to single quotes appearing inside the patch in a string, but since this is testing only, i think it's ok and should fix the mentioned problem.

/lgtm

gthiemonge:
> I guess it's a problem of single quotes within single quotes, and `to_json` makes double quotes instead. (`json.dumps` does that and according to docs, the filter uses that function under the hood)
> 
> So this is not resilient to single quotes appearing inside the patch in a string, but since this is testing only, i think it's ok and should fix the mentioned problem.
> 
> /lgtm

another method would be the use of raw text that represents a yaml doc, for instance for neutron_config_patch: https://github.com/openstack-k8s-operators/data-plane-adoption/blob/c76f0805a22f9fe9cb518ee62f65e795a14ed6dc/tests/roles/neutron_adoption/defaults/main.yaml#L2

ciecierski:
> > I guess it's a problem of single quotes within single quotes, and `to_json` makes double quotes instead. (`json.dumps` does that and according to docs, the filter uses that function under the hood)
> > So this is not resilient to single quotes appearing inside the patch in a string, but since this is testing only, i think it's ok and should fix the mentioned problem.
> > /lgtm
> 
> another method would be the use of raw text that represents a yaml doc, for instance for neutron_config_patch:
> 
> https://github.com/openstack-k8s-operators/data-plane-adoption/blob/c76f0805a22f9fe9cb518ee62f65e795a14ed6dc/tests/roles/neutron_adoption/defaults/main.yaml#L2

I would prefer to keep the same code convention, rather to use to_json.
I opened https://github.com/openstack-k8s-operators/data-plane-adoption/pull/1381 with `|` instead

PR #1374, opened 2026-05-06
Automatically close PRs against the downstream branch

jistr:
Assisted-by: Goose+Gemini

### DIFF
--- a/.github/workflows/auto-close-downstream-prs.yml
+++ b/.github/workflows/auto-close-downstream-prs.yml
@@ -1,0 +1,23 @@
+name: Close pull requests to downstream branch
+
+on:  # yamllint disable-line rule:truthy
+  pull_request_target:
+    branches:
+      - downstream
+    types: [opened, reopened]
+
+jobs:
+  auto-close-downstream-pr:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Comment and close PR
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-close-downstream-prs.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Comment and close PR
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr comment ${{ github.event.pull_request.number }} \
+            --repo ${{ github.repository }} \
```
ciecierski: Maybe it is better to stick to example from https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-github-cli for repo variable. However, if the github.repository works for you this is not blocking change. 

suggested change to follow example` REPO: ${{ github.event.repository.name }}`

jistr: Thanks, i think you're right so i'm updating it. I'm not sure how to best test it, maybe in a fork but i think we may just as well merge and fix if anything is broken :)

Btw in the example the $GITHUB_REPOSITORY is actually not set, i guess they meant to write $REPO :)

ciecierski (approved):

PR #1377, opened 2026-05-06
Fix the auto-close PR workflow

jistr:
Now i managed to find that the github.repository was correct after all:

https://docs.github.com/en/actions/reference/workflows-and-actions/variables


ciecierski (approved):
Good to know. Let's merge it.

PR #1382, opened 2026-05-07
Use server-side apply for swift-ring-files ConfigMap

fmount:
Client-side `oc apply` writes the entire manifest into the last-applied-configuration annotation, which can exceed the limit when Swift ring files are large. Using `oc apply --server-side` avoids this annotation entirely while remaining idempotent, unlike `oc create` which would fail on retries if the resource already exists.

Closes: [OSPRH-25258](https://redhat.atlassian.net/browse/OSPRH-25258)

### DIFF
--- a/docs_user/modules/proc_adopting-the-object-storage-service.adoc
+++ b/docs_user/modules/proc_adopting-the-object-storage-service.adoc
@@ -30,7 +30,7 @@ EOF
 . Create the `swift-ring-files` `ConfigMap` that includes the {object_storage} ring files:
 +
 ----
-$ oc apply -f - <<EOF
+$ oc apply --server-side -f - <<EOF
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


docs_user/modules/proc_adopting-the-object-storage-service.adoc
```diff
@@ -42,6 +42,13 @@ binaryData:
   object.ring.gz: $($CONTROLLER1_SSH "base64 -w0 /var/lib/config-data/puppet-generated/swift/etc/swift/object.ring.gz")
 EOF
 ----
++
+[NOTE]
+====
+The `--server-side` flag is required because the {object_storage} ring files
+can be large enough to exceed the annotation size limit imposed by
+client-side `oc apply`.
```
klgill, line 50:
```suggestion
====
The `--server-side` flag is required if the {object_storage} ring files
are large enough to exceed the annotation size limit imposed by the
client-side `oc apply`.
```
Let me know if my edit changed the meaning. Could there be cases where the files do not exceed the size limit?

klgill (changes requested):

docs_user/modules/proc_adopting-the-object-storage-service.adoc
fmount: Hi @klgill thanks for the review, I will apply the requested changes.
Yes, there might be cases, for small clusters, where this won't be a concern, but we don't know much about the environment we're going to adopt in advance, and this option can be applied both in the case of a small cluster and in case the cluster is big enough to exceed that limit.
Waiting for the command to fail and then suggest the option in case the command doesn't succeed is probably not the best user experience (I guess it's what we're trying to fix here), so I wanted to propose this flag and have a `Note` explaining the reason behind that.

fmount: (Done)

klgill (approved):
/lgtm

ciecierski (approved):

jistr (commented):
/lgtm

PR #1387, opened 2026-05-13
neutron_verify, nova_verify: skip pre_launch.bash checks when it did not run

rebtoor:
## Problem

Several post-adoption verification tasks in neutron_verify.yaml and
nova_verify.yaml are guarded only by `prelaunch_test_instance|bool`.
These tasks assume pre_launch.bash ran and created the cirros image,
test VM, and floating IP. When only pre_launch_ironic.bash runs (e.g.
test_with_ceph_and_ironic), none of those resources exist and the
verifications fail.

This is particularly visible in IPv6 scenarios like uni04delta-ipv6
where the public_subnet_fip_1 default (192.168.122.20) is an IPv4
address unreachable on a pure-IPv6 network.

## Fix

Add the `pre_launch.bash` guard to all affected tasks, matching the
pattern already used in `octavia_post.yaml` and `nova_verify.yaml`
(ironic task). Also replace the hardcoded `ping` in neutron_verify
with the `ping_command` variable for consistency with
`pre_launch.bash.j2` and BGP scenario support.

Affected tasks:
- **neutron_verify**: FIP connectivity ping
- **nova_verify**: cirros image save, test VM diagnostics, stop, start

### DIFF
--- a/tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
+++ b/tests/roles/dataplane_adoption/tasks/neutron_verify.yaml
@@ -49,9 +49,11 @@
     success_msg: "neutron-dhcp-agent is ALIVE after adoption"
 
 - name: verify connectivity to the existing test VM instance using Floating IP
-  when: prelaunch_test_instance|bool
+  when:
+    - prelaunch_test_instance|bool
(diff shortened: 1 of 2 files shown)


jistr (approved):
/lgtm

PR #1395, opened 2026-05-14
Remove custom OVN ConfigMap/DataPlaneService workaround for DCN

fultonj:
Remove the custom ovncontroller-config-dcn ConfigMap, ovn-dcn DataPlaneService, and DCN nodeset patching that was working around OVN SB connectivity for DCN compute nodes. The standard OVN service uses DNS (ovsdbserver-sb.openstack.svc) which resolves to internalapi IPs via DNSData CRs. With proper routes in the internalapi NAD, the standard mechanism should work without hardcoding tcp:<IP> entries, which also broke TLS scenarios.

The NAD route patching is retained as it is still required for macvlan pods to reach DCN subnets.

Tracker: [OSPRH-30146](https://redhat.atlassian.net/browse/OSPRH-30146)


jistr (approved):
/lgtm

PR #1401, opened 2026-05-21
changed root to tripleo-admin

klgill:
https://redhat.atlassian.net/browse/OSPRH-29670


docs_user/modules/proc_retrieving-topology-specific-service-configuration.adoc
```diff
@@ -115,7 +115,7 @@ ifeval::["{build}" != "downstream"]
 $ CONTROLLER1_SSH="ssh -i ~/install_yamls/out/edpm/ansibleee-ssh-key-id_rsa root@192.168.122.100"
```
klgill, line 115: @jistr Some commands are conditionalized for upstream only, like the one on line 115. Should root also be changed to tripleo-admin for upstream?

jistr (approved):
/lgtm

PR #1412, opened 2026-06-05
nova_verify, pre_launch_ironic: fix ironic post-adoption provisioning

rebtoor:
## Problem

The post-adoption ironic instance provisioning task in nova_verify.yaml
can fail with quota exceeded (HTTP 403) when the adopted environment
already uses all available instances (e.g. uni04delta-ipv6 with 10 VMs
from the source cloud at the default quota of 10).

## Fix

- Bump the instance quota by one before creating the post-adoption test
  instance so it works regardless of source cloud instance count.
- Add ironic_post_adoption_create_instance variable (default true) to
  allow scenarios to skip the post-adoption provisioning step entirely.


jistr (approved):
/lgtm

jistr (approved):
/lgtm

holser (approved):
/lgtm

PR #1414, opened 2026-06-10
Update ip vip in controllers to avoid conflict ip with vip undercloud

fyanac:
Following the same config as hci scenario


jistr (approved):
/lgtm

jistr (approved):
Not sure why it's not merging, it looks like it already has everything it needs.
/lgtm
/approve

PR #1415, opened 2026-06-11
removed extra spaces and added note about DNS

klgill:
https://redhat.atlassian.net/browse/OSPRH-30593

### DIFF
--- a/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
+++ b/docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
@@ -560,8 +560,8 @@ endif::[]
         edpm_ovs_packages:
         - openvswitch3.3
         edpm_default_mounts:
-          -  path: /dev/hugepages<size>
-             opts: pagesize=<size>
+          - path: /dev/hugepages<size>
+            opts: pagesize=<size>
             fstype: hugetlbfs
             group: hugetlbfs
   nodes:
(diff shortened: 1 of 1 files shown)


docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -572,6 +572,7 @@ done
 +
 * `${compute}.hostName` specifies the FQDN for the node if your deployment has a custom DNS Domain.
 * `${compute}.networks` specifies the network composition. The network composition must match the source cloud configuration to avoid data plane connectivity downtime. The `ctlplane` network must come first. The commands only retain IP addresses for the hosts on the `ctlplane` and `internalapi` networks. Repeat this step for other isolated networks, or update the resulting files manually.
+* `${compute}.ansible.ansibleHost` specifies the ansible host for the Compute node. If you are using DNS, use the value `${!ip}`.
```
ciecierski: What does it mean ' the value `${!ip}`.' ?

ciecierski: Also use ${!ip}” seems inverted. It should be something along this lines according to jira bug `If that ansibleHost is not resolvable, set ansibleHost to the ctlplane IP (${!ip}), matching the value used for fixedIP on the ctlplane network.`

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
```diff
@@ -560,8 +560,8 @@ endif::[]
         edpm_ovs_packages:
         - openvswitch3.3
         edpm_default_mounts:
-          -  path: /dev/hugepages<size>
-             opts: pagesize=<size>
+          - path: /dev/hugepages<size>
```
ciecierski, line 563: this is fine

docs_user/modules/proc_adopting-compute-services-to-the-data-plane.adoc
klgill: I think this only applies to customers using DNS? Can your suggestion be changed to the following?

"If you are using DNS, set ansibleHost to the ctlplane IP (${!ip}), and match the value that is used for fixedIP on the ctlplane network."

klgill: @ciecierski Please review the above suggested rewrite^.

ciecierski (approved):

PR #1425, opened 2026-06-29
Update the IP brigde for ceph to avoid conflict with test IP range

fyanac:
For test IP range uses:
start=192.168.122.171,end=192.168.122.250
so to avoid IP conflict, it needs to change the current IP bridge of ceph.
In ci-framework-job files, IPs from .106 are assigned for ceph but it seems that it is not used.


jistr (approved):
/lgtm

PR #1438, opened 2026-07-10
Fix quota set to target admin project instead of default

abhibongale:
PR #1412 added a quota increase for pre-adoption Ironic test
  instances, but targeted project 'default' instead of 'admin'.
  The OS_CLOUD=overcloud authenticates as project admin
  (clouds.yaml: project_name: admin), so quota set should
  target admin to match the project instances are created under.

  Since adoption migrates the TripleO DB, this quota carries
  over to the adopted cloud, providing headroom for
  post-adoption instance creation as well.


rebtoor (approved):
/approve
/lgtm

jistr (approved):
/lgtm
/approve

PR #1439, opened 2026-07-12
Disable DhcpAgentNotification for uni01alpha and uni04delta-ipv6

renjingxiao:
Both scenarios include ironic-overcloud.yaml which sets DhcpAgentNotification: true. Override it to false to match the 17.1 gate job behavior.

Related-to: [OSPRH-27715](https://redhat.atlassian.net/browse/OSPRH-27715)

### DIFF
--- a/scenarios/uni01alpha/config_download.yaml
+++ b/scenarios/uni01alpha/config_download.yaml
@@ -47,6 +47,7 @@ parameter_defaults:
   OVNDBsVirtualFixedIPs:
     - ip_address: 192.168.122.131
       use_neutron: false
+  DhcpAgentNotification: false
   NeutronDnsDomain: 'example.com'
   BarbicanSimpleCryptoGlobalDefault: true
   SwiftEncryptionEnabled: true
--- a/scenarios/uni04delta-ipv6/config_download.yaml
+++ b/scenarios/uni04delta-ipv6/config_download.yaml
@@ -44,6 +44,7 @@ parameter_defaults:
     neutron::config::plugin_ml2_config:
       ovn/disable_ovn_dhcp_for_baremetal_ports:
         value: true
+  DhcpAgentNotification: false
   BarbicanSimpleCryptoGlobalDefault: true
   Debug: true
   DockerPuppetDebug: true


scenarios/uni01alpha/config_download.yaml
```diff
@@ -47,6 +47,7 @@ parameter_defaults:
   OVNDBsVirtualFixedIPs:
     - ip_address: 192.168.122.131
       use_neutron: false
+  DhcpAgentNotification: false
```
hjensas, line 50: I'm reluctant, the option is enabled for a reason in environments/services/ironic-overcloud.yaml. `DhcpAgentNotification` is `false` by default in TripleO, but `true` by default in neutron. When is is not enabled there is no notification sent to the DHCP Agent when a port is updated, early return[1], if notification are not sent the DHCP service is not updated when Ironic changes port configuration - for example setting `mac_address` and adding `extra_dhcp_opts`[2].

@renjingxiao 
1. Is the 17.1 gate job you refer to a job that uses Ironic?
2. Can you re-run test projects with the options to enable "pre-launch test Ironic instance" in role: `development_environment`?

[1] https://opendev.org/openstack/neutron/src/branch/master/neutron/api/rpc/agentnotifiers/dhcp_rpc_agent_api.py#L91
[2] https://opendev.org/openstack/neutron/src/branch/master/neutron/api/rpc/agentnotifiers/dhcp_rpc_agent_api.py#L308-L318

hjensas (changes requested):
In logs from eodweb I see that the `adoption-run-adoption` run's the pre-launch test ironic instance.

```
TASK [development_environment : pre-launch test Ironic instance] ***************
changed: [localhost] => {"changed": true, "cmd": ...
```

In @renjingxiao's log links in Jira I see:
```
TASK [development_environment : pre-launch test VM instance] *******************
skipping: [localhost] => {"changed": false, "false_condition": "prelaunch_test_instance|bool", "skip_reason": "Conditional result was False"}

TASK [development_environment : pre-launch test Ironic instance] ***************
skipping: [localhost] => {"changed": false, "false_condition": "prelaunch_test_instance|bool", "skip_reason": "Conditional result was False"}
```

Please see inline comments, I am reluctant to approve this change.
Can the TP be re-run with appropriate var's that do not disable the test seeds an ironic "workload" prior to adoption?

scenarios/uni01alpha/config_download.yaml
renjingxiao: Thanks for the review.

The 17.1 gate-ovn job is a non-Ironic deployment.

Rerunning with `prelaunch_test_instance` enabled to verify this.

PR #1446, opened 2026-07-16
mariadb_copy: wait for connectivity to TripleO Galera before sync check

eduolivares:
Commit 824d9edf replaced sleep-based BGP workarounds with a retry loop that checks connectivity to the podified MariaDB, but did not add an equivalent check for the source TripleO Galera nodes. With BGP, the route to the mariadb-copy-data pod may take a few seconds to be advertised, causing the subsequent WSREP sync check to fail with "ERROR 2002 (HY000): Can't connect to MySQL server" (EINPROGRESS).

Add a retry-loop task that verifies TCP connectivity to every TripleO Galera member via 'select 1;' before running the non-retrying WSREP sync check. Also update the documentation procedure to include this new connectivity check step and fix the BGP note wording to correctly describe route advertisement direction.

Assisted-By: Claude (claude-opus-4.6)

### DIFF
--- a/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
+++ b/docs_user/modules/proc_migrating-databases-to-mariadb-instances.adoc
@@ -186,6 +186,26 @@ $ oc wait --for condition=Ready pod/mariadb-copy-data --timeout=30s
 
 .Procedure
 
+. Wait for the `mariadb-copy-data` pod to reach the source TripleO Galera databases:
++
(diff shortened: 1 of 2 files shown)


jistr (approved):
/lgtm

This issue is what i was wondering about in my comment here but i didn't express myself clearly 😀 : https://github.com/openstack-k8s-operators/data-plane-adoption/pull/1361#discussion_r3091545532

PR #1448, opened 2026-07-17
Pin Github actions to SHAs

jistr:
Additionally, update the basic checkout and setup-python actions to non-EOL releases.

Assisted-by: goose+gemini


cjeanner (approved):

holser (approved):
/lgtm

PR #1450, opened 2026-07-23
Restrict test log outputs

jistr:
This to avoid printing secrets into logs, as another layer of security, even though the CI environments are isolated and short-lived.

The change does the following:

* Default `use_no_log: true`.

* Add `use_no_log` onto more tasks.

* Disable xtrace (`set -x`) on such tasks as another precaution. This is done via a new `shell_header_no_xtrace` variable so that the behavior is not hardcoded.

If debugging of the relevant tasks is necessary, it can be done by tweaking these variables for a testproject.

Assisted by: goose+gemini

### DIFF
--- a/tests/roles/backend_services/tasks/main.yaml
+++ b/tests/roles/backend_services/tasks/main.yaml
@@ -1,6 +1,8 @@
 - name: create osp-secret
+  no_log: "{{ use_no_log }}"
   ansible.builtin.shell: |
     {{ shell_header }}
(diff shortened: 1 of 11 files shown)


cjeanner (approved):
Sounds a reasonable way to avoid leaking data. I love the way you allow to override the "set -x" set in the default `shell_header`. Maybe this needs to be properly documented though in the adoption documentation?

ciecierski (approved):

PR #1458, opened 2026-07-27
Refactor docs workflow to use GitHub Pages Actions

ciecierski:
Replace force-pushing the gh-pages branch with a build job that uploads a Pages artifact and a deploy job that publishes via actions/deploy-pages.


jistr (approved):
/lgtm
Looks good, i wonder if we can run it somehow before merging, or we have to merge and see. Maybe in a fork it could be ran but i guess it involves setting up github pages and whatnot, maybe we can just merge and if anything breaks we can fix or as a last resort do a revert?

